PR #5, opened 2019-08-12
Generate sample icon

Jawnnypoo:
This creates app icons for the sample project using the Coil logo. 😎 

Screen Shot 2019-08-12 at 10 34 09 AM


colinrtwhite:
Awesome! Thanks. Can you run the PNGs through a compressor (I'm a fan of `zopfli`) and the vector drawables through [Avocado](https://github.com/alexjlockwood/avocado) if you haven't already?

Jawnnypoo:
All done!

colinrtwhite:
Two more comments:

- Can you rename `ic_launcher_background.xml` -> `colors.xml`.
- We should move `ic_launcher_foreground.xml` to `drawable-v24` so it's not packaged with the app on lower API levels.

Thanks!

Jawnnypoo:
Makes sense, should be good now!

PR #8, opened 2019-08-13
Proposal: Switch to just speak Call.Factory

ZacSweers:
This is a proposal PR that switches the `ImageLoaderBuilder` + internals to just speak `Call.Factory`. This is similar to [how Retrofit handles it](https://github.com/square/retrofit/blob/master/retrofit/src/main/java/retrofit2/Retrofit.java#L456-L473), and allows for consumers to provide their own implementations of `Call.Factory`.

Two primary examples:
- A clever technique to improve startup perf is to lazily instantiate the OkHttpClient stack (instance + cache) by deferring its construction until the first network call. This way, construction is done _off_ the main thread. I demoed this in c03e7c2, and can revert 69ca796 if you want to keep that in the sample.
- Some apps provide entirely different network implementations behind `Call.Factory`, using the interface to allow them to interop with popular libraries like Retrofit still. This would enable those as well.

In order to sanely support this, I think the builder helper method didn't have a place, and instead made the logic for applying those configurations available via public helper extension function. This way consumers can use it in their own construction, and also better supports sharing connection pools between instances. If clients want to use the optimized configuration, they can just call `newBuilder()` on their existing instance to share the connection pool while gettin the optimized configuration for a new instance ultimately provided into `callFactory`.

### DIFF
--- a/coil-base/src/main/java/coil/ImageLoaderBuilder.kt
+++ b/coil-base/src/main/java/coil/ImageLoaderBuilder.kt
@@ -23,7 +23,6 @@ import okhttp3.OkHttpClient
 class ImageLoaderBuilder(private val context: Context) {
 
     private var callFactory: Call.Factory? = null
-    private var canRebuildWithOptimizations: Boolean = false
 
     private var registry: ComponentRegistry? = null
 
--- a/coil-base/src/main/java/coil/ImageLoaderBuilder.kt
+++ b/coil-base/src/main/java/coil/ImageLoaderBuilder.kt
@@ -15,6 +15,7 @@ import coil.util.getDrawableCompat
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.Dispatchers
 import okhttp3.Cache
+import okhttp3.Call
 import okhttp3.Dispatcher
 import okhttp3.OkHttpClient
 
@@ -22,8 +23,8 @@ import okhttp3.OkHttpClient
 @BuilderMarker
 class ImageLoaderBuilder(private val context: Context) {
 
-    private var okHttpClient: OkHttpClient? = null
-    private var okHttpClientBuilder: (OkHttpClient.Builder.() -> Unit)? = null
+    private var callFactory: Call.Factory? = null
+    private var canRebuildWithOptimizations: Boolean = false
 
     private var registry: ComponentRegistry? = null
 
@@ -33,24 +34,12 @@ class ImageLoaderBuilder(private val context: Context) {
     private var defaults = DefaultRequestOptions()
 
     /**
-     * Set the [OkHttpClient] to be used for network requests.
+     * Set the [Call.Factory] to be used for network requests.
      *
-     * Prefer `okHttpClient(OkHttpClient.Builder.() -> Unit)` if possible,
-     * as the default [OkHttpClient] instance is optimized for Coil.
+     * If using an [OkHttpClient] instance under the hood, use [applyCoilOptimizations] if possible to optimize it for Coil.
      */
-    fun okHttpClient(client: OkHttpClient) = apply {
-        this.okHttpClient = client
-        this.okHttpClientBuilder = null
-    }
-
-    /**
-     * Set the callback that is invoked when building the default [OkHttpClient].
-     *
-     * @see buildDefaultOkHttpClient
-     */
-    fun okHttpClient(builder: OkHttpClient.Builder.() -> Unit) = apply {
-        this.okHttpClientBuilder = builder
-        this.okHttpClient = null
+    fun callFactory(callFactory: Call.Factory) = apply {
+        this.callFactory = callFactory
     }
 
     /**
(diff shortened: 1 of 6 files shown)


ZacSweers:
Quick note: for those who don't know - `OkHttpClient` is an implementation of `Call.Factory`, so clients can be passed directly as well via `callFactory(client)`

colinrtwhite:
I really like the idea of using `Call.Factory`. A few thoughts:

- We should keep the `okHttpClient(OkHttpClient)` method, but also provide a `callFactory(CallFactory)` method similar to how Retrofit handles it. This helps with discovery.
- I'm not a huge fan of `applyCoilOptimizations`. I think I would prefer to keep `okHttpClient(builder: OkHttpClient.Builder.() -> Unit)` and only apply the optimizations in the case where a user doesn't set a client or call factory. If the user sets an OkHttpClient builder, I'd apply the optimizations there as well.

^ This doesn't cover the case where a user lazy inits the `OkHttpClient`, but still wants Coil's default OkHttpClient configuration. Maybe it would make sense to make a public `Utils` method for that? What do you think?

ZacSweers:
> We should keep the okHttpClient(OkHttpClient) method, but also provide a callFactory(CallFactory) method similar to how Retrofit handles it. This helps with discovery.

Agreed, confirmed this is why Retrofit does it as well 👍 

> Maybe it would make sense to make a public Utils method for that? What do you think?

Yep I'd agree with that. That was the idea behind extracting it to the function. I do think it should be one or the other though. Trying to balance all of them together would create confusing API imo, as it'd have `okHttpClient()`, `callFactory()`, and `okHttpClient(Builder.())` where they all have different implications/behavior. Applying the builder onto the directly-handed OkHttpClient also seems potentially problematic if the user is trying to limit the behavior of the client they pass in.

colinrtwhite:
I think you're right that `okHttpClient(Builder.())` should be removed as it has behaviour that the other two methods don't, which can be confusing. Going to think about this a little more since ideally I'd like to avoid a Utils method since it's not super discoverable.

I'm likely going to ship `0.6.1` tomorrow, but the conversion to `Call.Factory` would be a good candidate for `0.7.0`.

ZacSweers:
Sounds good, just let me know :+1:

coil-base/src/main/java/coil/ImageLoaderBuilder.kt
```diff
@@ -15,15 +15,16 @@ import coil.util.getDrawableCompat
...
 import okhttp3.OkHttpClient
 
 /** Builder for an [ImageLoader]. */
 @BuilderMarker
 class ImageLoaderBuilder(private val context: Context) {
 
-    private var okHttpClient: OkHttpClient? = null
-    private var okHttpClientBuilder: (OkHttpClient.Builder.() -> Unit)? = null
+    private var callFactory: Call.Factory? = null
+    private var canRebuildWithOptimizations: Boolean = false
```
colinrtwhite: Let's remove `canRebuildWithOptimizations` since we should only apply these optimizations if we need to build the default call factory.

coil-base/src/main/java/coil/ImageLoaderBuilder.kt
```diff
@@ -33,24 +34,12 @@ class ImageLoaderBuilder(private val context: Context) {
...
 
     /**
-     * Set the [OkHttpClient] to be used for network requests.
+     * Set the [Call.Factory] to be used for network requests.
      *
-     * Prefer `okHttpClient(OkHttpClient.Builder.() -> Unit)` if possible,
-     * as the default [OkHttpClient] instance is optimized for Coil.
+     * If using an [OkHttpClient] instance under the hood, use [applyCoilOptimizations] if possible to optimize it for Coil.
      */
-    fun okHttpClient(client: OkHttpClient) = apply {
```
colinrtwhite: Let's keep this (but remove `okHttpClientBuilder`) for discoverability.

coil-base/src/main/java/coil/ImageLoaderBuilder.kt
```diff
@@ -179,26 +168,32 @@ class ImageLoaderBuilder(private val context: Context) {
...
-            .apply { okHttpClientBuilder?.invoke(this) }
+            .applyCoilOptimizations(context)
             .build()
     }
+
+    companion object {
+        /**
+         * Applies the preferred optimizations for Coil (image loading library) to this given Builder instance.
+         */
+        fun OkHttpClient.Builder.applyCoilOptimizations(context: Context): OkHttpClient.Builder = apply {
```
colinrtwhite: Can you move this into `Utils`, set `Utils` to public, and set every other method to `internal`? I'd also change the name to `applyOkHttpClientOptimizations`.

colinrtwhite:
@ZacSweers Left a few comments. Still not 100% sure on having a `Utils` method to apply the optimizations, but I wanted to avoid delaying this PR more.

coil-base/src/main/java/coil/ImageLoaderBuilder.kt
ZacSweers: ad0d666

coil-base/src/main/java/coil/ImageLoaderBuilder.kt
ZacSweers: fbdd36a

coil-base/src/main/java/coil/ImageLoaderBuilder.kt
ZacSweers: By happy coincidence, I forgot to wire this up anywhere. Just had to delete the field. 34b0090

ZacSweers:
Done and rebased

coil-base/src/main/java/coil/util/OkHttpClients.kt
```diff
@@ -0,0 +1,28 @@
...
+
+import android.content.Context
+import okhttp3.Cache
+import okhttp3.Dispatcher
+import okhttp3.OkHttpClient
+
+/**
+ * [OkHttpClient] utilities for Coil.
+ */
+object OkHttpClients {
```
colinrtwhite: Can you remove this + make `applyCoilOptimizations` a top level function?

coil-base/src/main/java/coil/util/OkHttpClients.kt
```diff
@@ -0,0 +1,28 @@
+package coil.util
```
colinrtwhite: @file:JvmName("OkHttpClients")

ZacSweers: Done in 9384a04

coil-base/src/main/java/coil/util/OkHttpClients.kt
ZacSweers: Done in 9384a04

PR #9, opened 2019-08-13
Create new rotate transformation

jeziellago:
Create new rotate transformation similar the Glide and Picasso.


isfaaghyth:
nice idea! maybe you should create a test class for that @jeziellago to make sure it works fine

jeziellago:
> nice idea! maybe you should create a test class for that @jeziellago to make sure it works fine

Yes! It's done. Can you review that, please? Thanks!

colinrtwhite:
Thanks for creating this. I appreciate the work, but in an effort to keep Coil small, we're not looking to add new `Transformations` to the base package at the moment. However, we're working on a separate transformations artifact that this might be better suited for; it'll be public soon.

Issue #10, opened 2019-08-13
Support loading images from AssetManager


ianhanniballake:
**Is your feature request related to a problem? Please describe.**
There's a very specific type of `Uri` on Android in the form of `file:///android_asset/your_asset.jpg` which, despite its `file://` prefix, cannot be opened with `ContentResolver.openInputStream()` but instead must be opened using [`AssetManager`](https://developer.android.com/reference/android/content/res/AssetManager).

This is true for all Uris in the form of `ContentResolver.SCHEME_FILE + ":///android_asset/"`.

**Describe the solution you'd like**
I'd like to be able to use:

    imageView.load("file:///android_asset/starrynight.jpg")

And have it automatically use `AssetManager`. I'd also be fine with using the `load(Uri)` version, although that would be a difference from Picasso and Glide (see examples below).

**Additional context**
Assets are already supported by Picasso:

     Picasso.get().load("file:///android_asset/starrynight.png").into(imageView)

 and Glide

    Glide.with(this).load("file:///android_asset/starrynight.jpg").into(imageView)

colinrtwhite:
Interesting! `imageView.load(String)` is currently reserved for HTTP URLs, but we could support `imageView.load(Uri.parse("file:///android_asset/starrynight.jpg"))` instead. I think we could make the necessary changes in `UriFetcher`. I'll make a quick PR.

ianhanniballake:
`load(Uri)` support would be great, thanks!

It seems like in other image loading libraries, `load(uri: String) = load(Uri.parse(uri))` in all cases and they share the same underlying processing. This ensures that no matter if developers use a `String` or `Uri`, the same rules apply, avoiding all cases where developers need to remember to `htmlUri.toString()` or `Uri.parse(contentUriString)`. Probably worth a separate FR than this as that would be a wider issue than this particular case.

colinrtwhite:
Good point. Currently, Coil uses `url` as the named param for `load(String)` so we'd have to change that to something more general (but hopefully not too general like `data`), but I like the idea of supporting Uris *and* Urls via `load(String)`. We could do this by writing a new `Mapper` or modifying `HttpUrlMapper`.

colinrtwhite:
Created a new issue to support parsing Strings as Uris here: https://github.com/coil-kt/coil/issues/24

PR #12, opened 2019-08-13
Support loading images from AssetManager.

colinrtwhite:
Fixes https://github.com/coil-kt/coil/issues/10.

This was proposed as an enhancement, but is more of a bugfix. There was a test covering asset manager loading, but it used Robolectric. Let's make it an instrumentation test instead.

### DIFF
--- a/coil-base/src/androidTest/java/coil/fetch/UriFetcherTest.kt
+++ b/coil-base/src/androidTest/java/coil/fetch/UriFetcherTest.kt
@@ -9,51 +9,31 @@ import coil.bitmappool.BitmapPool
 import coil.bitmappool.FakeBitmapPool
 import coil.size.PixelSize
 import coil.util.createOptions
-import coil.util.createTestMainDispatcher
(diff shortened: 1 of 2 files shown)


ianhanniballake:
Overall, LGTM.

i think personally I might have preferred a separate `AssetFetcher` that would have a `handles()` that specifically calls out this special case (updating the `handles` of `UriFetcher` to exclude asset Uris if it isn't possible to guarantee ordering between the two) rather than complicate the logic of `UriFetcher` itself, but from an external developer perspective, the chosen approach does fix the issue and future restructuring can be done without changing the external API surface / behavior if deemed useful.

colinrtwhite:
@ianhanniballake Good point. I think I prefer this approach because it avoids creating a new `Fetcher` for a "special" Uri case, but we can always move this logic into a separate `Fetcher` without changing the API if it becomes too unwieldily.

ianhanniballake:
Makes sense to me.

PR #15, opened 2019-08-13
Retrieve java version from const

igorwojda:
use costs representing java version in the root build.gradle


colinrtwhite:
Thanks for the PR. Appreciate the work, but I think I'd prefer to keep it as is since both [Kotlin's official docs](https://kotlinlang.org/docs/reference/using-gradle.html) and [OkHttp](https://square.github.io/okhttp/upgrading_to_okhttp_4/#gradle) recommend using Strings for these values.

Issue #20, opened 2019-08-13
ImageView with one of its dimensions set to wrap_content does not work


Stonos:
**Describe the bug**
Assume the following ImageView:
```xml
<ImageView
        android:id="@+id/imageView"
        android:layout_width="wrap_content"
        android:layout_height="100dp"
        android:adjustViewBounds="true"
        android:scaleType="fitCenter" />
```
If you try to load an image to it, Coil will not calculate the correct width (based on the ImageView's fixed height and the aspect ratio of the image being loaded), and the ImageView will end up with a 0 width.

Using `size()` solves this issue, but it is not ideal because it is not always possible to know beforehand the dimensions of the image being loaded.

**To Reproduce**
Load an image to an ImageView which has one fixed dimension, and one `wrap_content` dimension.
Sample project: [CoilWrapContent.zip](https://github.com/coil-kt/coil/files/3496918/CoilWrapContent.zip)

**Expected behavior**
Coil should automatically calculate the dimension set to `wrap_content` based on the fixed dimension & the dimensions of the image being loaded.

For example: If I load a 400x200 (2:1 aspect ratio) image to an ImageView with `android:layout_width="wrap_content"` and `android:layout_height="100dp"`, then Coil should automatically set the width to 200dp (perhaps this shouldn't apply to all scaleTypes).

**Logs/Screenshots**
Layout Inspector screenshot showing the ImageView having its width set to 0:
![image](https://user-images.githubusercontent.com/42659/62941850-b1cefd80-bddf-11e9-9ed8-c207d8a42d0e.png)

**Library version**
0.6.0

colinrtwhite:
Good catch! I think this is because we force `PixelSize`'s dimensions to be > 0. Relaxing that limitation to >= 0 should solve the issue, but might cause side-effects. Going to work on this later tonight.

Stonos:
Unfortunately #29 does not seem to fully solve the issue. On the sample app the ImageView now has a 1:1 aspect ratio instead of the 2:1 [the image being loaded](https://via.placeholder.com/400x200) has:

![image](https://user-images.githubusercontent.com/42659/63106437-7ffa9a00-bf8b-11e9-8814-07e467b87bb0.png)

colinrtwhite:
[...]

PR #29, opened 2019-08-14
Ensure View dimensions are evaluated to at least 1px.

colinrtwhite:
Fixes: https://github.com/coil-kt/coil/issues/20

This fixes the case where one or more of a View's dimensions resolves to 0px. In this case, we treat the dimension as 1px. Image dimensions must be loaded with all dimensions > 0 so this saves us from handling a lot of complexity later in the image pipline.

### DIFF
--- a/coil-base/src/main/java/coil/size/ViewSizeResolver.kt
+++ b/coil-base/src/main/java/coil/size/ViewSizeResolver.kt
@@ -4,6 +4,7 @@ import android.view.View
 import android.view.ViewTreeObserver
 import kotlinx.coroutines.suspendCancellableCoroutine
 import kotlin.coroutines.resume
+import kotlin.math.max
 
 /**
  * A [SizeResolver] that measures the size of a [View].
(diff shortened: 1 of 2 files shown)


coil-base/src/main/java/coil/size/ViewSizeResolver.kt
```diff
@@ -38,19 +39,11 @@ interface ViewSizeResolver<T : View> : SizeResolver {
...
-                        if (width > 0 && height > 0) {
-                            isResumed = true
-                            viewTreeObserver.removePreDrawListenerSafe(this)
-                            continuation.resume(PixelSize(width, height))
-                        }
-                    }
+                    viewTreeObserver.removePreDrawListenerSafe(this)
+                    val width = max(1, view.width - view.paddingLeft - view.paddingRight)
+                    val height = max(1, view.height - view.paddingTop - view.paddingBottom)
+                    continuation.resume(PixelSize(width, height))
```
Jawnnypoo: `!isResumed` check no longer needed?

colinrtwhite: I don't believe so, but going to double check on a few test projects before merging. I added it to ensure we don't resume the coroutine twice, however it should be fine as long as `viewTreeObserver.removePreDrawListenerSafe(this)` is before `continuation.resume(PixelSize(width, height))`.

colinrtwhite: Actually, I didn't think we did initially, but I'm going to leave it in for safety. It's to prevent resuming the continuation twice, which would throw an exception.

PR #33, opened 2019-08-15
Adds the last modified date to the cache key for file Uris.

rharter:
This is a pretty straightforward change to add the file's last modified timestamp to the cache key for file Uris to accurately determine freshness.

### DIFF
--- a/coil-base/src/androidTest/java/coil/fetch/UriFetcherTest.kt
+++ b/coil-base/src/androidTest/java/coil/fetch/UriFetcherTest.kt
@@ -19,6 +19,7 @@ import org.junit.Test
 import java.io.File
 import kotlin.test.assertEquals
 import kotlin.test.assertFalse
+import kotlin.test.assertNotEquals
 import kotlin.test.assertTrue
 
 class UriFetcherTest {
@@ -44,7 +45,6 @@ class UriFetcherTest {
 
         val uri = file.toUri()
         assertTrue(loader.handles(uri))
-        assertEquals(uri.toString(), loader.key(uri))
 
         val result = runBlocking {
             loader.fetch(pool, uri, PixelSize(100, 100), createOptions())
@@ -58,7 +58,6 @@ class UriFetcherTest {
     fun basicAssetFetch() {
         val uri = Uri.parse("file:///android_asset/normal.jpg")
         assertTrue(loader.handles(uri))
-        assertEquals(uri.toString(), loader.key(uri))
 
         val result = runBlocking {
             loader.fetch(pool, uri, PixelSize(100, 100), createOptions())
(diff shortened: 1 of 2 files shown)


coil-base/src/androidTest/java/coil/fetch/UriFetcherTest.kt
```diff
@@ -88,4 +87,29 @@ class UriFetcherTest {
         val result = loader.extractAssetFileName(uri)
         assertEquals(null, result)
     }
+
+    @Test
+    fun fileCacheKeyWithLastModified() {
+        val file = File(context.filesDir.absolutePath + File.separator + "file.jpg")
+        val uri = Uri.fromFile(file)
```
colinrtwhite: Nit: Can you swap this to `file.toUri()`?

colinrtwhite:
Left one comment, but otherwise LGTM. Thanks for the contribution!

colinrtwhite:
Also does `File.lastModified` work for `file://android_asset` Uris?

coil-base/src/androidTest/java/coil/fetch/UriFetcherTest.kt
rharter: done.

coil-base/src/androidTest/java/coil/fetch/UriFetcherTest.kt
```diff
@@ -88,4 +87,35 @@ class UriFetcherTest {
...
+        val firstKey = loader.key(uri)
+
+        file.setLastModified(4321L)
+        val secondKey = loader.key(uri)
+
+        assertNotEquals(secondKey, firstKey)
+    }
+
+    @Test
+    fun assetCacheKeyWithLastModified() {
```
rharter: I added a test for `file:///android_asset` Uris, as they always return 0 for `lastModified`.

colinrtwhite: Thanks! I think this is good for the moment, but seems like this is a good point to separate `UriFetcher` and `AssetUriFetcher` since we don't need `File.lastModified` for an asset's cache key. I (or another community member) can address this in a separate up PR.

rharter: I agree!  Honestly, you might also want a `ContentUriFetcher`, since it looks like the mime type resolution that is currently being used only works for `content://` Uris.  I ended up updating my custom one like so:

```kotlin
  @SuppressLint("DefaultLocale")
  private fun getMimeType(data: Uri): String? {
    return if (data.scheme == ContentResolver.SCHEME_CONTENT) {
      context.contentResolver.getType(data)
    } else {
      MimeTypeMap.getFileExtensionFromUrl(data.toString())?.let {
        val typeMap = MimeTypeMap.getSingleton()
        when {
          it == "image" -> "image/jpeg" // Handles some stupid iOS conventions.
          typeMap.hasMimeType(it) -> typeMap.getMimeTypeFromExtension(it.toLowerCase())
          else -> null
        }
      }
    }
  }
```

colinrtwhite: Interesting, I'll circle back and take a look at this for 0.7.0.

Issue #6, opened 2019-08-12
Enable native support for SVG format without boilerplate


FrostRocket:
Currently using Glide to load SVG images and the implementation is quite clunky in order to do so. See: https://stackoverflow.com/questions/35507893/does-glide-have-a-method-for-loading-both-png-and-svg

As a fresh new library hot off the presses, I'd like to see Coil gracefully (and optimally) handle these file formats without a ton of boilerplate.

EDIT: WebP is already supported, so this enhancement request is more about formats that require an external library for decoding (and a Decoder).

Thanks!

colinrtwhite:
Android (and Coil) [already supports WebP](https://developer.android.com/guide/topics/media/media-formats) from API 14+. Are you looking for it to support *animated* WebP on pre-P?

For SVGs, I think this might be a good candidate for an external library. However, it should be straightforward to write a custom [Decoder](https://coil-kt.github.io/coil/api/coil-base/coil.decode/-decoder/) for this using [AndroidSVG](http://bigbadaboom.github.io/androidsvg/). Something like:

```kotlin
class SvgDecoder : Decoder {

    override fun handles(source: BufferedSource, mimeType: String?) = mimeType?.endsWith("svg") == true

    override suspend fun decode(pool: BitmapPool, source: BufferedSource, size: Size, options: Options): DecodeResult {
        return DecodeResult(
            drawable = SvgDrawable(SVG.getFromInputStream(source.inputStream())),
            isSampled = false
        )
    }
}

class SvgDrawable(private val svg: SVG) : Drawable() {

    override fun draw(canvas: Canvas) {
        svg.renderToCanvas(canvas)
    }

    override fun setAlpha(alpha: Int) {
        // Unsupported.
    }

    override fun getOpacity() = PixelFormat.TRANSLUCENT

    override fun setColorFilter(colorFilter: ColorFilter?) {
        // Unsupported
    }
}
```

should work well.

Disclaimer: The above code is completely untested.

FrostRocket:
I am currently using SVGDecoder via Glide, so it wasn't a heavy lift to add some of the sample code you wrote and try it out myself. From what I can tell, it's decoding the SVG fine, but they don't appear to be scaling correctly within the ImageView bounds. 

Before:

`@GlideModule
class SvgModule : AppGlideModule() {
    override fun registerComponents(context: Context, glide: Glide, registry: Registry) {
        registry
[...]

PR #38, opened 2019-08-16
Adds support for external SVG decoder.

rharter:
This adds a `coil-svg` module so that users can add SVG support with a single line.

There is probably room for optimization based on the `Options` but I leave that as a future contribution.


coil-sample/src/main/assets/images.json
```diff
@@ -1,15 +1,39 @@
 [
     {
         "urls": {
-            "raw": "https://images.unsplash.com/photo-1550973886-796d048c599f?ixlib=rb-1.2.1&ixid=eyJhcHBfaWQiOjU4MjM5fQ",
-            "full": "https://images.unsplash.com/photo-1550973886-796d048c599f?ixlib=rb-1.2.1&q=85&fm=jpg&crop=entropy&cs=srgb&ixid=eyJhcHBfaWQiOjU4MjM5fQ",
-            "regular": "https://images.unsplash.com/photo-1550973886-796d048c599f?ixlib=rb-1.2.1&q=80&fm=jpg&crop=entropy&cs=tinysrgb&w=1080&fit=max&ixid=eyJhcHBfaWQiOjU4MjM5fQ",
-            "small": "https://images.unsplash.com/photo-1550973886-796d048c599f?ixlib=rb-1.2.1&q=80&fm=jpg&crop=entropy&cs=tinysrgb&w=400&fit=max&ixid=eyJhcHBfaWQiOjU4MjM5fQ",
-            "thumb": "https://images.unsplash.com/photo-1550973886-796d048c599f?ixlib=rb-1.2.1&q=80&fm=jpg&crop=entropy&cs=tinysrgb&w=200&fit=max&ixid=eyJhcHBfaWQiOjU4MjM5fQ"
+            "raw": "https://cdn.coinranking.com/Sy33Krudb/btc.svg",
```
rharter: I figured it made sense to include these in the sample app, so it details how to use these dependencies, but I'd assume the final sample should use different imagery.  AFAIK Unsplash doesn't offer SVGs, so I'm not really sure where to get them from.

coil-svg/build.gradle.kts
```diff
@@ -0,0 +1,46 @@
...
+    libraryVariants.all {
+        generateBuildConfigProvider?.configure { enabled = false }
+    }
+    testOptions {
+        unitTests.isIncludeAndroidResources = true
+    }
+}
+
+afterEvaluate {
+    tasks.withType<DokkaAndroidTask> {
```
rharter: This was copied over from `coil-gif`, but I don't think it's required.  Can you confirm?

rharter:
Fixes #6

coil-sample/build.gradle.kts
```diff
@@ -42,6 +42,7 @@ android {
 
 dependencies {
     implementation(project(":coil-default"))
+    implementation(project(":coil-svg"))
```
colinrtwhite: Thanks for adding this, however I'd like to leave the `coil-svg` module out of the sample app for the moment. Going to work on a separate PR to have new screens for the `gif` and `svg` modules.

coil-svg/build.gradle.kts
colinrtwhite: It's required to create the API docs for https://coil-kt.github.io/coil/.

coil-svg/src/main/java/coil/decode/SvgDecoder.kt
```diff
@@ -0,0 +1,32 @@
...
+ */
+class SvgDecoder : Decoder {
+
+    override fun handles(source: BufferedSource, mimeType: String?) = mimeType == "image/svg+xml"
+
+    override suspend fun decode(pool: BitmapPool, source: BufferedSource, size: Size, options: Options): DecodeResult {
+        try {
+            val svg = source.use { SVG.getFromInputStream(it.inputStream()) }
+            return DecodeResult(
+                    drawable = PictureDrawable(svg.renderToPicture()),
```
colinrtwhite: Nit: 4 space indentation

coil-svg/src/main/java/coil/decode/SvgDecoder.kt
```diff
@@ -0,0 +1,32 @@
...
+                    isSampled = false
+            )
+        } catch (e: SVGParseException) {
+            throw IllegalStateException("Failed to load SVG.", e)
```
colinrtwhite: How come you're wrapping the exception here?

coil-svg/src/main/java/coil/decode/SvgDecoder.kt
```diff
@@ -0,0 +1,32 @@
...
+import coil.bitmappool.BitmapPool
+import coil.size.PixelSize
+import coil.size.Size
+import com.caverock.androidsvg.RenderOptions
+import com.caverock.androidsvg.SVG
+import com.caverock.androidsvg.SVGParseException
+import okio.BufferedSource
+
+/**
+ * A [Decoder] that uses android-SVG to decode SVG files.
```
colinrtwhite: Looks like they style it as `AndroidSVG` + I would link to their site: `A [Decoder] that uses [AndroidSVG](http://bigbadaboom.github.io/androidsvg/) to decode SVG files.`

coil-svg/src/main/java/coil/decode/SvgDecoder.kt
rharter: I've been using this (actually a Glide SVG decoder) in my app for a while and wrap the exception to avoid needing to expose the SVG library as an API dependency.  Also because consumers don't need to know whether it's an SVG that's being loaded to handle an exception (passing a string from an API to the lib, consumers shouldn't need to know or care the file type).

That being said, let me know if you want it changed.

coil-svg/src/main/java/coil/decode/SvgDecoder.kt
rharter: For continuation indents, too? Will update.

coil-svg/src/main/java/coil/decode/SvgDecoder.kt
colinrtwhite: Good point about keeping the decoder as an `implementation` dependency. Looks like `SVGParseException` extends `SAXException` (XML parse exception) w/o any extra metadata so I think listeners can listen for that and we can avoid wrapping the exception. It might make sense to add a `DecodeException` sometime in the future.

coil-sample/src/main/assets/images.json
FrostRocket: Let's use new SVGs please, not sure if we're allowed to use their assets in this way. Found a few samples that could work:

https://dev.w3.org/SVG/tools/svgweb/samples/svg-files/android.svg
https://dev.w3.org/SVG/tools/svgweb/samples/svg-files/mac.svg
https://dev.w3.org/SVG/tools/svgweb/samples/svg-files/photos.svg

colinrtwhite: @FrostRocket See: https://github.com/coil-kt/coil/pull/38#discussion_r314913499

PR #43, opened 2019-08-18
Create test for gray scale transformation

jeziellago:
New test for gray scale transformation, asserting that it transform correctly.


coil-base/src/androidTest/java/coil/transform/GrayscaleTransformationTest.kt
```diff
@@ -0,0 +1,63 @@
...
+
+        val coloredBitmap = getColoredBitmap()
+
+        val grayscaleBitmap = runBlocking {
+            grayscaleTransformation.transform(pool, coloredBitmap)
+        }
+
+        val (_, red, green, blue) = grayscaleBitmap.getPixels()
+
+        assertTrue(red.contentEquals(green) && green.contentEquals(blue))
```
colinrtwhite: To make this test more robust, I would add the output of the grayscale transformation to the test assets as `normal_grayscale.jpg`. Then you can use [Bitmaps.sameAs(other, 0.99)](https://github.com/coil-kt/coil/blob/master/coil-base/src/sharedTest/java/coil/util/Bitmaps.kt#L39) to verify that the result matches the expected output.

coil-base/src/androidTest/java/coil/transform/GrayscaleTransformationTest.kt
```diff
@@ -0,0 +1,63 @@
...
+        assertFalse(grayscaleBitmap.sameAs(coloredBitmap))
+    }
+
+    private fun getColoredBitmap(): Bitmap {
```
colinrtwhite: ^ Can remove this function after moving to [Bitmaps.sameAs(other, 0.99)](https://github.com/coil-kt/coil/blob/master/coil-base/src/sharedTest/java/coil/util/Bitmaps.kt#L39).

colinrtwhite:
Thanks for the contribution! Left a few comments.

coil-base/src/androidTest/java/coil/transform/GrayscaleTransformationTest.kt
jeziellago: Nice! I agree you! Can you review if make sense my last changes? Thanks! :)

jeziellago:
@colinrtwhite Can you review if make sense my last changes? Thanks! :)

coil-base/src/androidTest/java/coil/transform/GrayscaleTransformationTest.kt
```diff
@@ -0,0 +1,71 @@
...
+        assertTrue(red.contentEquals(green) && green.contentEquals(blue))
+        assertTrue(grayscaleBitmap.sameAs(normalGrayscaleBitmap, THRESHOLD))
+    }
+
+    private fun getBitmapFromAssets(
+        fileName: String
+    ): Bitmap {
+        val source = context.assets.open(fileName).source().buffer()
+        val (drawable, _) = runBlocking {
+            decoder.decode(
```
colinrtwhite: To keep this unit test minimal, I would not use `BitmapFactoryDecoder` in this test. Instead, I would remove this method and inline `BitmapFactory.fromInputStream(context.assets.open(fileName))`.

jeziellago: Make sense!

jeziellago:
@colinrtwhite Done!

PR #45, opened 2019-08-19
ArithmeticException: divide by zero bug has been fixed

R12rus:
https://i.pinimg.com/originals/d6/68/ab/d668abc72809303852c27275e6a56775.gif?156

java.lang.ArithmeticException: divide by zero
        at coil.drawable.MovieDrawable.draw(MovieDrawable.kt:65)
        at android.widget.ImageView.onDraw(ImageView.java:1246)


colinrtwhite:
Thanks for the contribution. I'm going to merge this in since it's good to guard against `movie.duration()` being 0, but I think the root cause is somewhere in `Movie` since it seems to only be able to read the first frame of the GIF.

Issue #48, opened 2019-08-19
Creating default OkHttp Cache can trigger strict mode violation.


saschpe:
**Describe the bug**
See strict mode warning below.

**To Reproduce**
Use *Image.load()* from inside a RecyclerView.Adapter. I didn't see it happening when it's used from inside a Fragment.

**Expected behavior**
**Dispatchers.IO** should be used. It also wouldn't hurt to be able to pass a custom lifecycle (actually a different issue) as well.

**Logs/Screenshots**
```
2019-08-19 16:00:08.236 17277-17277/ D/StrictMode: StrictMode policy violation; ~duration=326 ms: android.os.strictmode.DiskReadViolation
        at android.os.StrictMode$AndroidBlockGuardPolicy.onReadFromDisk(StrictMode.java:1504)
        at java.io.UnixFileSystem.checkAccess(UnixFileSystem.java:251)
        at java.io.File.exists(File.java:1152)
        at java.io.File.mkdirs(File.java:1714)
        at coil.util.Utils.getDefaultCacheDirectory(Utils.kt:33)
        at coil.ImageLoaderBuilder.buildDefaultOkHttpClient(ImageLoaderBuilder.kt:189)
        at coil.ImageLoaderBuilder.build(ImageLoaderBuilder.kt:182)
        at coil.Coil.buildDefaultImageLoader(Coil.kt:57)
        at coil.Coil.loader(Coil.kt:20)
...
```

**Library version**
0.6.1

colinrtwhite:
Currently, the OkHttpClient is built synchronously. https://github.com/coil-kt/coil/pull/8 will fix this issue by allowing us to lazily create the OkHttpClient instance on a background thread.

colinrtwhite:
Fix for this has been merged to master. If you need it right away, you can use `0.7.0-SNAPSHOT`: https://github.com/coil-kt/coil/blob/master/docs/faq.md#how-do-i-get-development-snapshots

PR #50, opened 2019-08-20
Support lazy Call.Factory initialization.

colinrtwhite:
- Adds `okHttpClient` and `callFactory` methods that accept a factory.
- Removed `applyCoilOptimizations` in favour of `CoilUtils. createDefaultCache`.
- This removes the code that disabled the max hosts by request limit., however I'm not sure how useful that is.

Still unsure if we should just pass `Lazy<Call.Factory>` to `ImageLoader` vs. wrapping it in a `Call.Factory`.


coil-base/src/main/java/coil/fetch/HttpUrlFetcher.kt
```diff
@@ -12,7 +12,7 @@ import okhttp3.HttpUrl
 import okhttp3.Request
 
 internal class HttpUrlFetcher(
-    private val callFactory: Call.Factory
+    private val callFactory: Lazy<Call.Factory>
```
ZacSweers: Out of curiosity, why was this needed?

ZacSweers: Specifically, worried now that you'll be possibly incurring a double lazy cost now for cases like the dagger example

colinrtwhite: @ZacSweers I added support for the lazy call factory trick [here](https://github.com/coil-kt/coil/pull/50/commits/f6fd48da78d2a6478e747d28d5721bee4fd813fd#diff-a620fd560b7a610a8300f97bc2a41ea7R171), however noticed that that was creating two objects (an extra Call.Factory + the lazy delegate). I figured it would be to balance the cost between `callFactory(Call.Factory)` and `callFactory(() -> Call.Factory)`, as each one would only create one extra object. Do you think it's better to keep the `lazyCallFactory` method?

ZacSweers: I think it depends on the goal. Is it to make it lazy for a functional need or to just make the deferred initialization standard?

colinrtwhite: I'm leaning toward making deferred initialization the preferred approach since creating the disk cache can trigger strict mode warnings on some devices (https://github.com/coil-kt/coil/issues/48).

ZacSweers: Right, I think if that's the case then you have the Call.Factory API only and leave delegation to consumers to implement. If you wrap something lazily, internals shouldn't need to do about that since you can just wrap it in `Call.Factory` again. If you really want to add something as a convenience for consumers, you could do something like this:

```kotlin
inline fun ImageLoaderBuilder.callFactory(
  crossinline body: (Request) -> Call
): ImageLoaderBuilder = callFactory(object : Call.Factory {
  override fun newCall(request: Request): Call = body(request)
})

inline fun ImageLoaderBuilder.delegatingCallFactory(
  delegate: Lazy<OkHttpClient>
): ImageLoaderBuilder = callFactory {
  delegate.get().newCall(it)
}
```

colinrtwhite: Gotcha, I'll move back to using `Call.Factory` directly, however I do really like having a convenience function like:

```kotlin
fun callFactory(initializer: () -> Call.Factory) = apply {
    this.callFactory: Call.Factory = lazyCallFactory(initializer)
}
```

It's a bit more magical than `delegatingCallFactory(Lazy<OkHttpClient>)`, but I think the syntax still makes sense:

```kotlin
val imageLoader(context) {
    // Imperative
    callFactory(OkHttpClient())

    // Lazy
    callFactory { OkHttpClient() }
}
```

hosanaSwaggart:
`<098">`

Issue #42, opened 2019-08-17
android.resource Uris that point to another package's resource ID fail to load


audkar:
**Describe the bug**
`imageView.load(uri)` doesn't load image from `android.resource://` uri.

**To Reproduce**
1. Apply this [patch](https://gist.github.com/audkar/c8d9c34dfc1c4d06accef741a4f3d385) to coil sample app.
2. Run sample app
3. Notice icon image is not loaded.
4. To check expected result comment out `detail.load(uri)` and un-comment `detail.setImageURI(uri)`

**Expected behavior**
Image loaded from uri

**Logs/Screenshots**

| Current | Expected |
| ---------- | ------------ |
| ![Screenshot_1566084096](https://user-images.githubusercontent.com/1947112/63218216-3367b800-c15f-11e9-9706-f8ce8f59f998.png) | ![Screenshot_1566083962](https://user-images.githubusercontent.com/1947112/63218219-44b0c480-c15f-11e9-9419-2912a853100a.png) |

**Library version**
0.6.1

tested on Android 10

colinrtwhite:
Thanks for the report. I'll take a look into this later today.

colinrtwhite:
Confirmed. `android.resource` Uris that point to a vector drawable (or anything that can't be handled by `BitmapFactory` currently won't load correctly. Have an idea in mind of how to fix this and will make sure the fix makes it into 0.7.0.

As a work around you can pass the resource ID directly to load: `.load(R.drawable.image)`

ediTLJ:
> As a work around you can pass the resource ID directly to load: `.load(R.drawable.image)`

Does that work for you? I also tried the resource id directly (with several vector drawables) and it doesn't seem to work.

colinrtwhite:
Closing this as it's fixed by #75. The fix will soon be released in `0.7.0` or you can [depend on the development snapshots](https://github.com/coil-kt/coil/blob/master/docs/faq.md#how-do-i-get-development-snapshots) if you need it before then.

audkar:
@colinrtwhite this still doesn't work if referenced `URI` is from different package.

I updated initial example patch to reproduce issue

colinrtwhite:
Thanks, I'll take a look into this next week (away on holiday this week).

DawnNguyenAhiho:
[...]

PR #51, opened 2019-08-20
Fix android.resource Uris that point to a vector drawable fail to load

pkpdeveloper:
Hi @colinrtwhite,

Please review and merge fix for [android.resource Uri](https://github.com/coil-kt/coil/issues/42) related issue. also added test cases for that.
I have tried many solutions only this one works as expected.

Please let me know for further improvements or any discussion. 

Cheers :beers:  for creating this awesome library :clap: :heart:

### DIFF
--- a/coil-base/src/main/java/coil/fetch/UriFetcher.kt
+++ b/coil-base/src/main/java/coil/fetch/UriFetcher.kt
@@ -2,8 +2,11 @@ package coil.fetch
 
 import android.content.ContentResolver
 import android.content.Context
+import android.content.pm.PackageManager
 import android.content.res.AssetManager
+import android.content.res.Resources
 import android.net.Uri
+import androidx.annotation.DrawableRes
 import androidx.annotation.VisibleForTesting
 import androidx.collection.arraySetOf
 import androidx.core.net.toFile
@@ -11,6 +14,8 @@ import coil.bitmappool.BitmapPool
 import coil.decode.DataSource
 import coil.decode.Options
 import coil.size.Size
+import coil.util.Utils
+import coil.util.getDrawableCompat
 import okio.buffer
 import okio.source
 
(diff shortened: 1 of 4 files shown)


coil-base/src/main/java/coil/fetch/UriFetcher.kt
```diff
@@ -42,6 +44,18 @@ internal class UriFetcher(
         size: Size,
         options: Options
     ): FetchResult {
+        if (data.scheme == ContentResolver.SCHEME_ANDROID_RESOURCE) {
+            val resourceId = extractResourceId(data)
+            resourceId?.let {
+                val drawable = context.getDrawableCompat(resourceId)
```
audkar: If I understand correctly, this code part assumes that `resourceId` belongs to same applications. However `Uri` can point to `resourceId` which belongs to different application:
```
"android.resource://{pkgName}/{resourceId}"
```

pkpdeveloper: Thanks, @audkar for pointing this issue. I got the solution from [Glide source code](https://github.com/bumptech/glide/blob/d8278f43a94484f20822170b71ac7b66269a2aa6/library/src/main/java/com/bumptech/glide/load/resource/drawable/ResourceDrawableDecoder.java) 
I will do the changes ASAP :blush:

audkar:
Implementation logic seems correct and aligned with platform one (private [method](https://android.googlesource.com/platform/frameworks/base/+/9be54d400d68c735013bc8069fbcb66c3f98c3ee/core/java/android/content/ContentResolver.java#470) used for `ImageView.setImageURI(uri)`). Handles boths possible formats for resource uri. Good job

colinrtwhite:
Thanks for the contribution. We're working on a bigger refactor of the mappers + fetchers [here](https://github.com/coil-kt/coil/pull/32), which might conflict with this PR. I'll circle back and review this PR after the one I linked is merged.

PR #55, opened 2019-08-21
Add ImageLoader.clearMemory(key) to clear a single memory cache item

Goddchen:
As discussed/requested by a user on Slack. Being able to clear a single item by its key from memory cache.


colinrtwhite:
Thanks for adding this. I want to spend a bit more time thinking about the design for the moment. This is a breaking change to the base `ImageLoader` interface which is pretty major.

Goddchen:
@Javieraeros merge conflicts resolved 👍

mario:
What do we do with this PR @colinrtwhite ?

Goddchen:
Merge? 😂

coil-base/src/main/java/coil/ImageLoader.kt
```diff
@@ -59,6 +59,11 @@ interface ImageLoader {
      */
     fun clearMemory()
 
+    /**
+     * Remove a single item from memory cache.
+     */
+    fun clearMemory(key: String)
```
colinrtwhite: Can we rename this to `invalidate`? The behaviour of this function is different enough from `clearMemory()` that it should probably have its own name. `clearMemory()` empties the bitmap pool as well as the memory cache whereas `clearMemory(key)` only removes a specific memory cache entry.

colinrtwhite:
Sorry for the delay on this! @Goddchen left one comment. I'm planning to do a quick `0.9.3` release soon, but after that release + the PR feedback, let's merge this in :).

Goddchen:
@colinrtwhite merged coil/master, resolved conflicts and renamed the method as requested. I think it can be merged now 👍

colinrtwhite:
@Goddchen Awesome! Thanks for the contribution.

PR #57, opened 2019-08-21
Added support for directories in android_asset resolver

JaCzekanski:
## Problem
Coil is not able to load images from android_asset that aren't placed in the root directory of assets.

## Example
```
imageBackground.load("file:///android_asset/background.jpg".toUri()) // Works ok
imageBackground.load("file:///android_asset/foo/background.jpg".toUri()) // Fails
```

## Changes
`extractAssetFileName` method in UriFetcher was modified to accept more than 2 path segments. Uri is now verified, stripped from android_asset prefix and joined to a path with subdirectories included.

### DIFF
--- a/coil-base/src/main/java/coil/fetch/UriFetcher.kt
+++ b/coil-base/src/main/java/coil/fetch/UriFetcher.kt
@@ -42,9 +42,9 @@ internal class UriFetcher(
         size: Size,
         options: Options
     ): FetchResult {
-        val assetFileName = extractAssetFileName(data)
-        val inputStream = if (assetFileName != null) {
-            context.assets.open(assetFileName)
+        val assetPath = extractAssetPath(data)
+        val inputStream = if (assetPath != null) {
+            context.assets.open(assetPath)
         } else {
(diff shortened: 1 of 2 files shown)


coil-base/src/main/java/coil/fetch/UriFetcher.kt
```diff
@@ -56,20 +56,26 @@ internal class UriFetcher(
...
-            segments[1].isNotBlank()) {
-            segments[1]
-        } else {
-            null
+        if (segments.count() < 2 || segments[0] != ASSET_FILE_PATH_SEGMENT) {
+            return null
         }
+
+        val path = segments
+                .drop(1)
```
colinrtwhite: Nit: Coil uses 4 spaces for continuation indent.

colinrtwhite:
Looks great; thanks for the fix. Left one minor nit.

coil-base/src/main/java/coil/fetch/UriFetcher.kt
JaCzekanski: Oops, I used IDE defaults for reformatting. Fixed.

Issue #44, opened 2019-08-18
Coil should automatically rotate images based on Exif information


ianhanniballake:
**Describe the bug**
When images have rotation information embedded within their Exif information, Coil should use that to automatically rotate the image correctly.

This is something that other image libraries (Glide, Picasso) do already. Using the [AndroidX ExifInterface library](https://developer.android.com/reference/androidx/exifinterface/media/ExifInterface) allows you to extract the orientation information from any `Uri` / `InputStream`.

**To Reproduce**
You can use [this sample image](https://drive.google.com/file/d/1Ry_kfL37_cqXWeCwLDML9_spCFqiK6Fo/view?usp=sharing).

**Expected behavior**
The image should be rotated correctly.

**Logs/Screenshots**
n/a

**Library version**
0.6.1

colinrtwhite:
Thanks for the report. Coil already uses the AndroidX Exif library as part of [BitmapFactoryDecoder](https://github.com/coil-kt/coil/blob/master/coil-base/src/main/java/coil/decode/BitmapFactoryDecoder.kt#L61) and should be automatically handling rotations, but I'll run the sample image to see what Coil's missing. I have a test [here](https://github.com/coil-kt/coil/blob/master/coil-base/src/androidTest/java/coil/decode/BitmapFactoryDecoderTest.kt#L97) as well to cover all the rotations.

ianhanniballake:
FWIW, the case I was testing was loading the image via `Uri` - specifically, from a local `ContentProvider`. I suppose a `FileProvider` generated `Uri` would be similar enough for testing purposes.

colinrtwhite:
Confirmed. Going to look into fixing this today.

PR #60, opened 2019-08-22
Fix parsing EXIF data longer than 8192 bytes.

colinrtwhite:
Fixes https://github.com/coil-kt/coil/issues/44.

The issue is that Okio will return the buffer's current size (8192 bytes - the size of a `Segment`) for `InputStream.available()`. The AndroidX EXIF parser treats this value as the max length of the stream and will silently throw an `EOFException` and use default EXIF values if it tries to read further than that amount.

@ianhanniballake seems like this is working as intended, but let me know if I should file a bug.

I've worked around this by wrapping the `InputStream` and returning `Int.MAX_VALUE` for the moment.

Will check with the Okio team to see if there's a way to avoid wrapping the `InputStream` + going to add a regression test tomorrow as well.


ianhanniballake:
The [`ExifInterface(InputStream)` constructor](https://developer.android.com/reference/androidx/exifinterface/media/ExifInterface.html?hl=en#ExifInterface(java.io.InputStream)) does specifically say that it is

> not intended to be used with an input stream that performs any networking operations.

So I suspect that the lack of support for streams that don't have the full available bytes `available()` immediately is indeed expected.

PR #63, opened 2019-08-22
Updating README.md to use JVM 1.8

fernandospr:


colinrtwhite:
Thanks for the contribution, however this is already covered in the [getting started section](https://coil-kt.github.io/coil/getting_started/#java-8).

fernandospr:
@colinrtwhite You're welcome. 

Unfortunately, new projects created in Android Studio don't come with Java8 settings configured out of the box. 

I was following the Quick Start step by step and got stuck trying to find the solution to this compilation error.
```
Cannot inline bytecode built with JVM target 1.8 into bytecode that is being built with JVM target 1.6. Please specify proper '-jvm-target' option
```
Therefore, I didn't realize there was a full documentation link below that.

colinrtwhite:
Good point. I've seen other people run into this issue as well. I'd like to keep the quickstart very short so I'll add a note in the readme that links to the getting started section on how to set up Java 1.8.

PR #75, opened 2019-08-29
Refactor Mappers and Fetchers to handle corner cases.

colinrtwhite:
Fixes: https://github.com/coil-kt/coil/issues/24 and https://github.com/coil-kt/coil/issues/42

Refactors Mappers and Fetchers to support treating Strings as Uris as well as fix some edge cases around resource Uris. It also fixes MIME type detection in a lot of cases.

This has a behaviour change where before only the first applicable Mapper would be used. Now, the ImageLoader will fold the data through the entire list of Mappers. Measured Mappers are still evaluated first before Mappers.


coil-base/src/main/java/coil/map/ResourceUriMapper.kt
```diff
@@ -0,0 +1,12 @@
...
+
+import android.content.ContentResolver
+import android.net.Uri
+import androidx.annotation.DrawableRes
+
+class ResourceUriMapper : Mapper<Uri, @DrawableRes Int> {
+
+    override fun handles(data: Uri) = data.scheme == ContentResolver.SCHEME_ANDROID_RESOURCE
+
+    override fun map(data: Uri) = checkNotNull(data.lastPathSegment?.toIntOrNull()) { "Malformed resource Uri: $data" }
```
pkpdeveloper: Hi @colinrtwhite 
Please provide support for another type of Uri pattern as well.

uri_format_exception

colinrtwhite: @pkpdeveloper Good point! I'll add support for this.

PR #78, opened 2019-09-02
add Korean language README

ChoiysApple:
Created Korean version of README which is README-ko


colinrtwhite:
This is great thanks! Do you know if there's a standard way to support multiple languages for READMEs on Github? I wonder if there's a way to default to README-ko if the user's language is set to Korean.

ChoiysApple:
Everybody wants that but I heard it's not available for Github right now. I googled and ask questions about that topic and I found only way is using github pages and develop that function in js

Issue #83, opened 2019-09-06
Disabling the Memory Cache crashes the app


hansenji:
**Describe the bug**
When set the `availableMemoryPercentage` to `0.0` it crashes the app in the LruCache

**To Reproduce**
 ```
Coil.setDefaultImageLoader {
    ImageLoaderBuilder(application)
        .availableMemoryPercentage(0.0)
        .build()
}
```

**Expected behavior**
Setting availableMemoryCache to 0 disables the cache not crashes.

**Logs/Screenshots**
```
java.lang.IllegalArgumentException: maxSize <= 0
        at androidx.collection.LruCache.<init>(LruCache.java:53)
        at coil.memory.MemoryCache$cache$1.<init>(MemoryCache.kt:24)
        at coil.memory.MemoryCache.<init>(MemoryCache.kt:24)
        at coil.RealImageLoader.<init>(RealImageLoader.kt:93)
        at coil.ImageLoaderBuilder.build(ImageLoaderBuilder.kt:177)
```

**Library version**
0.6.1

colinrtwhite:
Thanks for reporting the issue. I'll add the fix to the `0.7.0` release.

PR #84, opened 2019-09-06
Disabling the MemoryCache no longer crashes the app.

hansenji:
Issue #83


colinrtwhite:
Thanks for the contribution. In the case where `maxSize == 0`, I think it might be better to no-op the entire `MemoryCache` since we no longer need to calculate the size of the `Bitmap` or trim any memory.

I'm looking to ship `0.7.0` today so I created a PR [here](https://github.com/coil-kt/coil/pull/87) to do that that reuses some of your code.

PR #90, opened 2019-09-08
Treat WRAP_CONTENT as the largest display dimension.

colinrtwhite:
Fixes: https://github.com/coil-kt/coil/issues/20

### DIFF
--- a/coil-base/src/main/java/coil/size/ViewSizeResolver.kt
+++ b/coil-base/src/main/java/coil/size/ViewSizeResolver.kt
@@ -1,7 +1,10 @@
 package coil.size
 
+import android.util.Log
 import android.view.View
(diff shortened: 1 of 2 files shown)


mario:
@colinrtwhite I still experience the issue in #20, even with Master SNAPSHOT.

mario:
(For full transparency, the issue is visible when setting a placeholder)

colinrtwhite:
@mario Could you open a new bug to track (+ a reproduction case if possible)? Thanks!

mario:
Let me try to come up with a self contained sample - if I manage to do it,
will post along with an issue.

Thank you for your your continued patience.

colinrtwhite:
Thanks! No problem, thanks for helping improve Coil.

PR #94, opened 2019-09-09
Re-organize the buildSrc Library.kt file

itsandreramon:
This helps with managing the projects' dependencies in a cleaner way. All script files that rely on the Library.kt file file have also been updated to make use of the new structure.


colinrtwhite:
I appreciate the work put into this, however I'd prefer to keep the existing dependency format. The existing structure is terser and easier for autocompleting library names.

PR #119, opened 2019-09-22
Add Request.parameters to support passing custom data through the image pipeline.

colinrtwhite:
Adds the ability to add custom params to a request that are passed to `Fetcher`s and `Decoder`s.

I think this will be useful for adding params to extension libraries. For instance, we might eventually want to support loading a frame from a video as an extension library. This gives us a hook to pass in the time code.

EDIT: Ended up going forward with the video frame decoder idea here: https://github.com/coil-kt/coil/pull/122.

### DIFF
--- a/coil-base/src/main/java/coil/RealImageLoader.kt
+++ b/coil-base/src/main/java/coil/RealImageLoader.kt
@@ -43,6 +43,7 @@ import coil.request.BaseTargetRequestDisposable
 import coil.request.EmptyRequestDisposable
 import coil.request.GetRequest
 import coil.request.LoadRequest
+import coil.request.Parameters
 import coil.request.Request
 import coil.request.RequestDisposable
 import coil.request.ViewTargetRequestDisposable
@@ -273,21 +274,25 @@ internal class RealImageLoader(
     private fun <T : Any> computeCacheKey(
         fetcher: Fetcher<T>,
         data: T,
-        parameters: Map<String, Any>,
+        parameters: Parameters,
         transformations: List<Transformation>
     ): String? {
         val baseCacheKey = fetcher.key(data) ?: return null
+
         return buildString {
             append(baseCacheKey)
+
             if (parameters.isNotEmpty()) {
-                parameters.forEach { (key, value) ->
-                    append(';')
+                for ((key, entry) in parameters) {
+                    val cacheKey = entry.cacheKey ?: continue
+                    append('#')
                     append(key)
-                    append(',')
-                    append(value)
+                    append('=')
+                    append(cacheKey)
                 }
-                append(';')
+                append('#')
             }
+
             transformations.forEach { append(it.key()) }
         }
     }
--- a/coil-base/src/main/java/coil/RealImageLoader.kt
+++ b/coil-base/src/main/java/coil/RealImageLoader.kt
@@ -201,7 +201,7 @@ internal class RealImageLoader(
 
             // Compute the cache key.
             val fetcher = registry.requireFetcher(mappedData)
-            val cacheKey = request.key ?: computeCacheKey(fetcher, mappedData, request.transformations)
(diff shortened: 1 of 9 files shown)


coil-base/src/main/java/coil/request/RequestBuilder.kt
```diff
@@ -265,11 +273,38 @@ sealed class RequestBuilder<T : RequestBuilder<T>> {
...
         this.memoryCachePolicy = policy
     }
+
+    /**
+     * Set the parameters for this request.
+     *
+     * Parameters can be used to pass custom data to [Fetcher]s and [Decoder]s.
+     */
+    fun parameters(parameters: Map<String, Any>) {
+        this.parameters = parameters.toMap()
```
colinrtwhite: Performing a defensive copy.

colinrtwhite:
@Jawnnypoo Going to make a couple more changes to this PR so I would hold off reviewing for the moment.

colinrtwhite:
@Jawnnypoo Okay, should be ready. Let me know what you think.

coil-base/src/main/java/coil/RealImageLoader.kt
```diff
@@ -268,16 +269,30 @@ internal class RealImageLoader(
...
+
+            if (parameters.isNotEmpty()) {
+                for ((key, entry) in parameters) {
+                    val cacheKey = entry.cacheKey ?: continue
+                    append('#')
+                    append(key)
+                    append('=')
+                    append(cacheKey)
+                }
+                append('#')
```
Jawnnypoo: You always want the `#` at the end?

coil-base/src/main/java/coil/request/Parameters.kt
```diff
@@ -0,0 +1,73 @@
...
+    override fun toString() = map.toString()
+
+    fun newBuilder() = Builder(this)
+
+    /**
+     * @param value The parameter's value.
+     * @param cacheKey The parameter's cache key. If not null, this value will be added to a request's cache key.
+     */
+    data class Entry(
+        val value: Any,
```
Jawnnypoo: Would it be right to allow value to be nullable too? I guess if we are modeling this off of something like Retrofit and query params, `null` would not be allowed, but I could see some ways in which you might want to be able to pass null to signify that you don't want something to be set, such as if a decoder would automatically perform some behavior and then you want to `null` that out to not perform the automatic behavior. 

In the end, it is a design decision though, so I leave it up to you.

coil-base/src/main/java/coil/request/RequestBuilder.kt
```diff
@@ -247,14 +250,14 @@ sealed class RequestBuilder<T : RequestBuilder<T>> {
     /**
      * Set the cache key for this request.
      *
-     * By default, the cache key is computed by the [Fetcher] and any [Transformation]s.
+     * By default, the cache key is computed by the [Fetcher], any [parameters], and any [Transformation]s.
```
Jawnnypoo: Caps for `parameters`? Or does this work as is?

coil-base/src/main/java/coil/RealImageLoader.kt
colinrtwhite: Just to separate the params and the transformation so we get:

`file:///path/to/file#video_frame=1234#GrayscaleTransformation`

instead of:

`file:///path/to/file#video_frame=1234GrayscaleTransformation`

Probably not required, but it helps visually separate the params.

coil-base/src/main/java/coil/request/Parameters.kt
colinrtwhite: Good point. I can't think of a reason to not allow null so probably better to be more permissive. `Decoder`s will have to handle null anyways if the expected parameter is absent.

Issue #61, opened 2019-08-22
Use placeholderRes/errorRes if url is null


mare011:
It would be cool if passing null url would use placeholder drawable or error drawable if set. Right now it defeats the purpose of fluent api when you have to do something like:

`if (imageUrl != null) {
	it.load(imageUrl) {
		placeholder(placeholderRes)
	}
}
else {
	it.setImageResource(placeholderRes)
}`

mare011:
To me it would make more sense to use error when defined, and placehold as a fallback. Same goes for the case where you do have an url and there's an error.

mlilienberg:
I saw multiple points of view about this topic. Some people think that null is a valid case and should show nothing. Maybe it makes more sense to define your own null strategy.

`
imageView.load(itemModel.imageUrl) {
   placeholder(R.drawable.ic_placeholder)
   error(R.drawable.ic_error)
   isNull(R.drawable.ic_null)
}
`
or

`
imageView.load(itemModel.imageUrl) {
   placeholder(R.drawable.ic_placeholder)
   error(R.drawable.ic_error)
   isNull(null)
}
`

Related topic in glide a few years ago:
https://github.com/bumptech/glide/issues/268

colinrtwhite:
@mlilienberg Great point. What do you think about the keyword `empty` instead of `isNull`? e.g.:

```kotlin
imageView.load(itemModel.imageUrl) {
   placeholder(R.drawable.ic_placeholder)
   error(R.drawable.ic_error)
   empty(R.drawable.ic_error)
}
```

colinrtwhite:
@mlilienberg Great point. I think it makes sense to use the same API as Glide. i.e.:

```kotlin
imageView.load(itemModel.imageUrl) {
   placeholder(R.drawable.ic_placeholder)
   error(R.drawable.ic_error)
   fallback(R.drawable.ic_error)
}
```

Similar to `placeholder` and `error`, you should be able to set the default `fallback` on your `ImageLoader`.

colinrtwhite:
This is now supported in master. To get the fix asap you can [depend on `0.9.0-SNAPSHOT`](https://coil-kt.github.io/coil/faq/#how-do-i-get-development-snapshots).

NOTE: There might be minor behaviour changes to the feature before `0.9.0` stable.

PR #151, opened 2019-10-21
Support fallback drawable when request data is null.

colinrtwhite:
Fixes: #61


spauldhaliwal:
Looking forward to this!

colinrtwhite:
Merging this in, however its behaviour might change while in master and before the stable `0.9.0` release.

bishiboosh:
@colinrtwhite how does this affect `ImageView.clear` ? This uses `loadAny` with null value and would display an error drawable in the image view instead of just canceling requests

colinrtwhite:
@bishiboosh Yep, currently on master this will trigger the `onError` callback. This will be fixed before the `0.9.0` release as part of https://github.com/coil-kt/coil/pull/179.

Issue #99, opened 2019-09-13
Pass `DataSource` to `Target.onSuccess()`


chrisbanes:
The source is useful to know when to run certain types of animation/transition when loading images.

For instance, I apply a fade + saturation animation for images when they're 'first loaded'. The easiest way to implement this (approximately) is to look at where the image is loaded from. This was pretty easy with Glide's `TransitionFactory`, but not so with Coil.

I've added this to my app [here](https://github.com/chrisbanes/tivi/commit/a4d65bbbc9e83c2c0f1d9418d11325032523c7cb), by making the target implement both `Target` and `Request.Listener`.

colinrtwhite:
Added a comment on #98, which also discusses this ticket.

colinrtwhite:
This is now supported! 🎉 The transitions API is part of the `0.9.0-SNAPSHOT`.

Definitely looking for feedback on the API (it'll be marked as experimental for one release) so if you have suggestions let me know!

Both `ImageLoader` and `Request` builders accept a Transition.Factory, which enables you to return a new transition (or null for no transition) for each `Transition.Event`.

[Check out the `Transition` interface for more info.](https://github.com/coil-kt/coil/blob/master/coil-base/src/main/java/coil/transition/Transition.kt)

chrisbanes:
Thanks Colin! I'll take a look at this later today.

colinrtwhite:
Thanks! No rush - probably still a couple weeks away from releasing `0.9.0`.

Issue #98, opened 2019-09-13
Make ImageViewTarget an open class


chrisbanes:
Currently ImageViewTarget is final, meaning it can't be extended. It would be useful to make it `open`.

*Why?*
I want to add a custom animation/transition when the image is loaded. I could add a transformation and wrap the drawable in a custom `Animatable` (although no idea if that would work), but really I just want extend `ImageViewTarget` and add something similar to its crossfade functionality.

colinrtwhite:
Sorry for the late reply on this. `ImageViewTarget` is currently closed since I didn't want consumers relying on the internal behaviour of `onSuccessCrossfade`.

`PoolableViewTarget` requires implementations to stop using the previous `Drawable` as soon as the next lifecycle method is called. This works great when using `setImageDrawable`, but it doesn't account for transitions that need to use both Drawables for the duration of a transition. Currently, `ImageViewTarget` works around this with some special internal behaviour in `onSuccessCrossfade`. This is definitely a shortcut and I want to make this behaviour public, but haven't settled on a design yet.

I want to promote [composition over inheritance](https://en.wikipedia.org/wiki/Composition_over_inheritance) so my goal is to expose something like a `TransitionTarget` where implementations that want to handle transitions have access to a transition-friendly `onSuccess`. I don't think it makes sense to add DataSource to the base Target interface since most targets don't need to suspend the `onSuccess` method while animating.

Here's what I'm thinking for `TransitionTarget` (let me know what you think):

```kotlin
interface TransitionTarget : Target {
    suspend fun onSuccess(result: Drawable, source: DataSource)
}
```

Basically, Coil needs a good way to support custom transitions, which it doesn't have at the moment.

chrisbanes:
Sounds good to me. Making the onSuccess() suspend while the transition/animation takes place is neat API.

PR #162, opened 2019-11-08
Add support for custom transitions.

colinrtwhite:
~Still figuring out if the transitions API should allow crossfading to error/null drawables.~ It does.

~TODO: Add tests.~

Fixes: #98
Fixes: #99

### DIFF
--- a/coil-base/src/main/java/coil/annotation/ExperimentalCoil.kt
+++ b/coil-base/src/main/java/coil/annotation/ExperimentalCoil.kt
@@ -1,0 +1,10 @@
+package coil.annotation
+
+/**
+ * Marks declarations that are still **experimental**.
+ * Targets marked by this annotation design is incubating and may contain breaking changes in the future.
+ */
+@MustBeDocumented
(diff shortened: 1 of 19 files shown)


coil-base/src/main/java/coil/annotation/ExperimentalCoil.kt
```diff
@@ -0,0 +1,10 @@
+package coil.annotation
+
+/**
+ * Marks declarations that are still **experimental**.
+ * Targets marked by this annotation design is incubating and may contain breaking changes in the future.
+ */
+@MustBeDocumented
+@Retention(AnnotationRetention.BINARY)
+@Experimental(Experimental.Level.WARNING)
+annotation class ExperimentalCoil
```
colinrtwhite: Adding this for future use.

colinrtwhite: I lied :). Adding this to the transitions API for one release in case it needs to be modified to support other use cases.

PR #194, opened 2019-12-06
Fix CrossfadeDrawable scaling.

colinrtwhite:
`CrossfadeDrawable` should scale the drawable when drawing instead of just updating the child drawable's bounds.


colinrtwhite:
Closed in favour of: https://github.com/coil-kt/coil/pull/202

PR #196, opened 2019-12-07
Migrate to Github Actions.

colinrtwhite:
The `macOS-latest` image on Github Actions has HAXM installed, which makes emulator tests significantly faster.

This drops the CI build time from \~20 minutes to \~5 minutes. Builds that miss the cache are only slightly longer at \~7:30 minutes.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,47 @@
...
+              run: ./gradlew ktlintCheck
+
+            - name: Unit Tests
+              run: ./gradlew testDebugUnitTest
+
+            - name: Instrumentation Tests
+              uses: reactivecircus/android-emulator-runner@v2
+              with:
+                  api-level: 25
+                  arch: x86_64
```
ychescale9: FYI Google [recommends only using x86_64 images for API 26+](https://www.reddit.com/r/androiddev/comments/e5nulf/emulator_29211_and_amd_hypervisor_12_to_stable/f9l6aos?utm_source=share&utm_medium=web2x) 😃

colinrtwhite: Ah good catch! Also thanks for the great Github Action. Will update.

PR #209, opened 2019-12-22
Add request size to Transformation. Rework RoundedCornersTransformation.

### LINKED ISSUE #126: RoundedCornersTransformation is not working properly
**Describe the bug**
Corners are not rounded on all items of my RecyclerView.
Some pictures are rounded but other ones not.

**To Reproduce**
I have a RecyclerView that I initialize with a GridLayoutManager with three columns.
For every items, it is inflating a view with a layout that will be squared by a constraint into the constraintLayout that I added. The layout is the following one:

```
<?xml version="1.0" encoding="utf-8"?>
<androidx.constraintlayout.widget.ConstraintLayout 
    android:id="@+id/file_preview_layout"
    xmlns:android="http://schemas.android.com/apk/res/android"
    xmlns:app="http://schemas.android.com/apk/res-auto"
    android:layout_width="match_parent"
    android:layout_height="wrap_content"
    android:background="@android:color/transparent"
    android:clickable="true"
    android:focusable="true"
    >

  <ImageView
        android:id="@+id/fileImage"
        android:layout_width="match_parent"
        android:layout_height="0dp"
        android:layout_margin="1dp"
        android:scaleType="centerCrop"
        android:adjustViewBounds="true"
        app:layout_constraintBottom_toBottomOf="parent"
        app:layout_constraintDimensionRatio="H,1:1"
        app:layout_constraintEnd_toEndOf="parent"
        app:layout_constraintStart_toStartOf="parent"
        app:layout_constraintTop_toTopOf="parent"
        />
</androidx.constraintlayout.widget.ConstraintLayout>
```
I am loading the picture when the `onBindViewHolder` adapter method is called. On it I load an url with the following code:
```
fileImage.load(url) { transformations(RoundedCornersTransformation(16)) }
```

**Expected behavior**
All corners should be rounded when they are shown on a RecyclerView

**Logs/Screenshots**
As you can see, only one picture is corners rounded. All items are inflating the same layout
photo_2019-09-25_10-30-31

**Library version**
7.0.0

**Related Issue**
#96

colinrtwhite:
This PR includes two main behaviour / breaking changes:

- `Transformation` is now passed the resolved `Size` of the target. This is often different than the size of the `input` Bitmap. For instance, the target and the source image can have different aspect ratios.
- Any requests with transformations are now ineligible for [image sampling](https://coil-kt.github.io/coil/getting_started/#image-sampling). This is because we can't guarantee that a transformation results in the same image given different target sizes.

This PR also refactors some duplicate code into a `SizeLazy` object, which simplifies the `RealImageLoader.execute` method.

Fixes #126.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -3,7 +3,7 @@ root = true
 [*]
 indent_size=4
 insert_final_newline=true
-max_line_length=135
+max_line_length=130
 
 # ktlint
 disabled_rules=import-ordering,experimental:annotation,experimental:indent
--- a/.editorconfig
+++ b/.editorconfig
@@ -3,7 +3,7 @@ root = true
 [*]
 indent_size=4
 insert_final_newline=true
-max_line_length=130
+max_line_length=135
 
 # ktlint
 disabled_rules=import-ordering,experimental:annotation,experimental:indent
--- a/coil-base/src/androidTest/java/coil/RealImageLoaderIntegrationTest.kt
+++ b/coil-base/src/androidTest/java/coil/RealImageLoaderIntegrationTest.kt
@@ -287,7 +287,6 @@ class RealImageLoaderIntegrationTest {
         val size = PixelSize(100, 100)
         val result = runBlocking {
             imageLoader.applyTransformations(
-                scope = this,
                 result = DrawableResult(
                     drawable = drawable,
                     isSampled = false,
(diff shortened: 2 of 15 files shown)


Jawnnypoo:
Ah, so https://github.com/Commit451/coil-transformations will need to be updated 👍

colinrtwhite:
Yep, will require a new release of the transformations library. This PR isn't 100% finished yet but I'll message you when it's good to go.

PR #213, opened 2019-12-27
feat: coil-firebase-storage module to fetch image from Firebase Storage

b95505017:
Signed-off-by: Yu-Hsuan Lin <b95505017@gmail.com>


b95505017:
@colinrtwhite Is there anything I could help with this PR?

colinrtwhite:
Thanks for writing this; I appreciate the work, though I'm not sure this makes sense as a module to add to the main Coil repo. Loading images via Firebase Storage is a fairly specialized use case and the `Fetcher` is simple to write.

Also `StorageReference.await` is wrapping a network operation to fetch the image. I think it would be better to call `StorageReference.getDownloadUrl()` and pass the `Uri` into Coil where the network operation will be handled by `OkHttp`. This enables all of OkHttp's features like HTTP2 support and disk caching for free.

PR #228, opened 2020-01-06
Add support for decoding Gifs on API v18 and below

### LINKED ISSUE #218: MovieDrawable fails to decode GIFs on API 14-18.
**Describe the bug**
`MovieDrawable` fails to render GIFs on API 14-18. This is because `Movie.decodeStream` requires an `InputStream` that is resettable as shown in the logs:

```
             System.err  W  java.io.IOException
                         W      at java.io.InputStream.reset(InputStream.java:218)
                         W      at android.graphics.Movie.decodeStream(Native Method)
                         W      at coil.decode.GifDecoder.decode(GifDecoder.kt:37)
                         W      at coil.RealImageLoader$loadData$2.invokeSuspend(RealImageLoader.kt:391)
                         W      at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
                         W      at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:56)
                         W      at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:561)
                         W      at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:727)
                         W      at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:667)
                         W      at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:655)
                   skia  D  ------- reset threw an exception
```

**To fix:** we'll need to wrap `BufferedSource` with a custom `InputStream` implementation that buffers the entire GIF into memory until it is closed.

**Expected behavior**
The GIF is decoded and rendered properly.

**To Reproduce**
Attempt to decode any GIF on API 14-18.

NOTE: I've only been able to reproduce this on the emulator since I don't have a hardware device that runs a version of Android pre-KitKat.

**Version**
0.9.0

mario:
Fixes #218. 

A few GIFs are not shown in the sample app but I'm not really sure that's a fault of the code but rather of limitations of native Movie class on those versions of Android.

### DIFF
--- a/coil-gif/src/main/java/coil/decode/GifDecoder.kt
+++ b/coil-gif/src/main/java/coil/decode/GifDecoder.kt
@@ -5,6 +5,7 @@ package coil.decode
 import android.graphics.Bitmap
 import android.graphics.Movie
 import android.os.Build.VERSION.SDK_INT
+import android.os.Build.VERSION_CODES.JELLY_BEAN_MR2
 import android.os.Build.VERSION_CODES.O
 import coil.bitmappool.BitmapPool
 import coil.drawable.MovieDrawable
@@ -34,7 +35,14 @@ class GifDecoder : Decoder {
         options: Options
     ): DecodeResult {
         val drawable = MovieDrawable(
-            movie = source.use { checkNotNull(Movie.decodeStream(it.inputStream())) },
+            movie = if (SDK_INT <= JELLY_BEAN_MR2) {
+                source.use {
+                    val byteArray: ByteArray? = it.readByteArray()
+                    checkNotNull(Movie.decodeByteArray(byteArray, 0, byteArray!!.size)) { " Null response while decoding Gif" }
+                }
+            } else {
+                source.use { checkNotNull(Movie.decodeStream(it.inputStream())) }
+            },
             config = when {
                 options.allowRgb565 -> Bitmap.Config.RGB_565
                 SDK_INT >= O && options.config == Bitmap.Config.HARDWARE -> Bitmap.Config.ARGB_8888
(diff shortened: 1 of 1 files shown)


coil-gif/src/main/java/coil/decode/GifDecoder.kt
```diff
@@ -34,7 +35,14 @@ class GifDecoder : Decoder {
         options: Options
     ): DecodeResult {
         val drawable = MovieDrawable(
-            movie = source.use { checkNotNull(Movie.decodeStream(it.inputStream())) },
+            movie = if (SDK_INT <= JELLY_BEAN_MR2) {
+                source.use {
+                    val byteArray: ByteArray? = it.readByteArray()
+                    checkNotNull(Movie.decodeByteArray(byteArray, 0, byteArray!!.size)) { " Null response while decoding Gif" }
```
colinrtwhite: `it.readByteArray()` always returns non-null. Also could you remove the error message to match the other branch? Thanks

colinrtwhite:
Thanks for the fix. Left a few comments. I think ultimately we'll want to write an `InputStream` implementation that wraps a buffered source and buffers it instead of consuming it (like `BufferedSource.inputStream()` does), as it'll allow `Okio` to avoid allocating a large `ByteArray`. However, this works for the moment.

coil-gif/src/main/java/coil/decode/GifDecoder.kt
mario: Yup, leftover from testing. Will fix.

mario:
Understood - this was a quick fix. If you feel like there's no point in merging it, I might look into wrapping a buffered source instead.

colinrtwhite:
Let's merge this for now and we can leave the custom `InputStream` solution for later. Thanks!

mario:
Awesome, thanks.

PR #256, opened 2020-01-20
Check if view is attached to window before checking hwAcceleration

mario:
I've been able to reproduce the issue *sometimes* which led me to believe this is some kind of a race condition in checking the hw acceleration capabilities for which we need the View to be attached to the window. To be safe, I now check if we're attached to window before checking for the HW acceleration.

Please test and let me know if this works for you.

Signed-off-by: Mario Danic <mario@lovelyhq.com>

### DIFF
--- a/coil-base/src/main/java/coil/memory/RequestService.kt
+++ b/coil-base/src/main/java/coil/memory/RequestService.kt
@@ -142,7 +142,7 @@ internal class RequestService {
         if (!request.allowHardware) return false
 
         // Prevent hardware bitmaps for non-hardware accelerated targets.
-        if (request.target.run { this is ViewTarget<*> && !view.isHardwareAccelerated }) return false
+        if (request.target.run { this is ViewTarget<*> && (!view.isAttachedToWindow || !view.isHardwareAccelerated) }) return false
 
         return true
     }


colinrtwhite:
Thanks for taking a look, though it looks like `View.isHardwareAccelerated` already requires the view to be attached to return true (`View.isAttached` checks `mAttachInfo != null`).

mario:
Ah indeed, sorry about that.  Let me take a further look, but it's tricky since it can't be reproduced every time (at least I can't).

mario:
@colinrtwhite do you have a fool-proof way to reproduce? I tried several times with https://d2lnr5mha7bycj.cloudfront.net/product-image/file/primary_2a461463-ca35-44b6-82b7-426375c8f70c.png on API 26 and it seems to work as expected now.

mario:
(Without this PR)

colinrtwhite:
No worries, thanks for taking a look. It might be an issue in the emulator we can't work around in which case we should report on the Android bug tracker. Though, if it's possible to work around it that'd be great.

colinrtwhite:
@mario I'm able to reproduce it on a Pixel 2 API 26 emulator (29.3.0) by replacing `detail.load(screen.image.url)` with `detail.load("https://d2lnr5mha7bycj.cloudfront.net/product-image/file/primary_2a461463-ca35-44b6-82b7-426375c8f70c.png")`.

mario:
@colinrtwhite I use 29.3.6. and can't reproduce with master. Let me see if I can downgrade.

mario:
Do you mind updating the emulator and trying again please?

colinrtwhite:
@mario I see `29.3.4` as the latest and am able to reproduce it on that version. Do you know how I can get access to `29.3.6`? Also are you running on MacOS with HAXM?

mario:
@colinrtwhite yes, I use MacOS with HAMX with AS 4.0 Canary 8. Updating AS might help with the emulator update option(s)? :)

colinrtwhite:
@mario Hmm after updating to `29.3.6` and recreating my emulator I'm still able to reproduce the crash.

mario:
I will recreate the emulator - strange that we have different results.

Thanks for testing it.

colinrtwhite:
Indeed! Going to give this another shot on a different computer tomorrow as well.

mario:
So I've tried it two more times with a recreated emulator and it work well (without this PR as well, obviously, so I'll close it) - I use x86 image. Is that the same one you use?

mario:
Screenshot 2020-01-27 at 14 23 40

colinrtwhite:
@mario Yep, I'm using the x86 emulator with Google APIs.

PR #295, opened 2020-02-16
Add WeakMemoryCache implementation.

colinrtwhite:
This adds a `WeakMemoryCache` implementation, which holds weak references to bitmaps.

The plan is `MemoryCache` holds strong references and when they're evicted, pass the bitmap to the `WeakMemoryCache` where it will hold a weak reference until either the bitmap is no longer referenced or it is added to the `BitmapPool`.

The use case for this is now you can rely on an `ImageLoader` having a value in its memory cache if the image still has a strong reference outside of it (i.e. an `ImageView`). I'd argue this is much more intuitive - basically if it's on screen, it's cached. This should make shared element transitions easier to implement and will increase the overall cache hit rate of the memory cache (at the slight cost of extra overhead to manage the weak references).

The behaviour can be enabled/disabled with `ImageLoaderBuilder.trackWeakReferences`. Planning to enable this by default since it should be a fairly safe optimization.


coil-base/src/main/java/coil/ImageLoader.kt
```diff
@@ -63,6 +64,7 @@ interface ImageLoader {
     /**
      * Completely clear this image loader's memory cache and bitmap pool.
      */
+    @MainThread
     fun clearMemory()
```
colinrtwhite: These functions have always called through to data structures that aren't thread safe, however now I'm making this assumption explicit. We could support calling these from any thread, however it would slightly slow down every image request due to synchronization so let's optimize for the common case.

You can technically still call these methods off the main thread, however it's not 100% safe due to race conditions.

coil-base/src/main/java/coil/ImageLoaderBuilder.kt
```diff
@@ -165,6 +167,18 @@ class ImageLoaderBuilder(context: Context) {
...
 
+    /**
+     * Enables weak reference tracking of loaded images.
+     *
+     * This allows the image loader to hold weak references to loaded images.
+     * This ensures that if an image is still in memory it will be returned from the memory cache.
+     *
+     * Default: true
+     */
+    fun trackWeakReferences(enable: Boolean) = apply {
```
colinrtwhite: Definitely open to better name suggestions for this.

coil-base/src/main/java/coil/RealImageLoader.kt
```diff
@@ -474,7 +479,7 @@ internal class RealImageLoader(
         private var size: Size? = null
 
         @MainThread
-        suspend inline fun size(cached: BitmapDrawable? = null): Size = scope.run {
```
colinrtwhite: Pretty sure this `inline` isn't actually being inlined because `var size` is private.

PR #301, opened 2020-02-25
Introduce ImageLoaderProvider.

colinrtwhite:
Looking for feedback on this!

This introduces `ImageLoaderProvider` which is used to initialize the default `ImageLoader`. This mirrors the way [`WorkManager`](https://developer.android.com/topic/libraries/architecture/workmanager/advanced/custom-configuration#implement-configuration-provider) and [`CameraX`](https://developer.android.com/jetpack/androidx/releases/camera#camera-core-1.0.0-alpha07) handle initialization by allowing you to implement the interface on your `Application`. Overall, I think this should make the singleton set up clearer.

This PR also renames a couple `Coil` methods:

`loader` -> `imageLoader`
`setDefaultImageLoader` -> `setImageLoader`

for symmetry.

### DIFF
--- a/coil-default/src/main/java/coil/ImageLoaderProvider.kt
+++ b/coil-default/src/main/java/coil/ImageLoaderProvider.kt
@@ -1,0 +1,17 @@
+package coil
+
+import android.app.Application
+
+/**
+ * A class that can provide the default [ImageLoader].
+ *
+ * Either implement [ImageLoaderProvider] on your [Application] **or** call [Coil.setImageLoader]
+ * with your [ImageLoaderProvider] to provide the default [ImageLoader].
+ */
+interface ImageLoaderProvider {
+
+    /**
+     * Return the default [ImageLoader].
+     */
+    fun getImageLoader(): ImageLoader
+}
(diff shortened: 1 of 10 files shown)


coil-default/src/main/java/coil/ImageLoaderProvider.kt
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * Either implement [ImageLoaderProvider] on your [Application] **or** call [Coil.setImageLoader]
+ * with your [ImageLoaderProvider] to provide the default [ImageLoader].
+ */
+interface ImageLoaderProvider {
+
+    /**
+     * Return the default [ImageLoader].
+     */
+    fun getImageLoader(): ImageLoader
```
Jawnnypoo: I see that the CameraX and WorkManager implementations use this `get` syntax. Personally, I would like `create` better than `get`, but for the sake of consistency with other libraries, 👍

colinrtwhite: Yeah, `get` doesn't feel 100% right in this instance, however I'd rather be consistent with AndroidX. Also this method doesn't necessarily have to create a new instance each time it's called.

PR #311, opened 2020-03-12
Rename ImageLoaderProvider to ImageLoaderFactory.

colinrtwhite:
Thought about this some more and `Provider` doesn't feel right even though it's in line with `AndroidX`. Rename to `ImageLoaderFactory` and `newImageLoader` to imply creation. This is continues follows the API naming convention in `OkHttp`. @Jawnnypoo you were right!


coil-default/src/main/java/coil/Coil.kt
```diff
@@ -26,29 +26,29 @@ object Coil {
...
-     * Using this method to set an explicit [provider] takes precedence over an [Application] that
-     * implements [ImageLoaderProvider].
+     * Using this method to set an explicit [factory] takes precedence over an [Application] that
+     * implements [ImageLoaderFactory].
      */
     @JvmStatic
     @Synchronized
-    fun setImageLoader(provider: ImageLoaderProvider) {
-        imageLoaderProvider = provider
+    fun setImageLoader(factory: ImageLoaderFactory) {
```
Jawnnypoo: Should the method be `setImageLoaderFactory` instead?

Jawnnypoo: Actually, I get it now. Probably best to keep it like this so the consumer gets a choice of how they want to provide the image loader 👍

colinrtwhite: Yep, I think it'll make more sense too when we can make `ImageLoaderFactory` a `fun interface` in Kotlin 1.4 so it's callable using `Coil.setImageLoader { ImageLoader(context) }`. Mostly I wanted to give both methods the same name to show they overwrite each other.

PR #318, opened 2020-03-20
Inline Build.VERSION_CODES constants.

### LINKED ISSUE #236: Replace imported SDK_INT constants with actual numbers
Some of my favorite candidates for fix are in `RealBitmapPool`:

    if (SDK_INT >= O) {
        configs += Bitmap.Config.RGBA_F16
    }

    require(SDK_INT < O || config != Bitmap.Config.HARDWARE) { "Cannot create a mutable hardware Bitmap." }

Using actual API level values will promote readability, i.e. nobody will have to ponder for minutes why the hell do we check whether SDK_INT is less than, greater than or equal to zero.

See https://twitter.com/jakewharton/status/1049421501559128065 for discussion. Main points:
* [JWh] Just use the numbers. They already convey what you want. `@RequiresApi(21)`, `minSdkVersion 21`, `setWhateverApiLevel(21)`. `CAN_USE_THING_API_LEVEL = 21`.
* [JWa] Agree, though this doesn't solve the practical issue of being unable to find API-specific code reliably. Searching for two digits in a large project is still painful. Searching for `@RequiresApi`, `@SdkSuppress`, `Build.VERSION.SDK_INT`, and comments is less painful but not obvious.
* [JWh] True! Although I rarely need to find all of those at once. Usually it's just one and you always need to use the annotation/field because someone may have used a number or a semantic constant like `FINGERPRINT_ENABLED`.

colinrtwhite:
Closes #236

### DIFF
--- a/coil-base/src/main/java/coil/ImageLoaderBuilder.kt
+++ b/coil-base/src/main/java/coil/ImageLoaderBuilder.kt
@@ -152,7 +152,7 @@ class ImageLoaderBuilder(context: Context) {
      *
      * If false, any use of [Bitmap.Config.HARDWARE] will be treated as [Bitmap.Config.ARGB_8888].
      *
-     * NOTE: Setting this to false this will reduce performance on Android O and above. Only disable if necessary.
+     * NOTE: Setting this to false this will reduce performance on API 26 and above. Only disable if necessary.
      *
      * Default: true
      */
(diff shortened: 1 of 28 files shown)


Jawnnypoo:
Closes https://github.com/coil-kt/coil/issues/236

PR #322, opened 2020-03-24
Deprecate Coil and ImageLoader extension functions.

colinrtwhite:
Looking for feedback on this!

This PR deprecates the `ImageLoader.load(context, data)`, `ImageLoader.get(data)`, `Coil.load(context, data)`, and `Coil.get(data)` functions in favour of a more standard builder syntax:

```kotlin
// LoadRequest
val request = LoadRequest.Builder(context)
    .target(imageView)
    .build()
val disposable = imageLoader.launch(request)

// GetRequest
val request = GetRequest.Builder()
    .size(1080, 1920)
    .build()
val drawable = imageLoader.launch(request)
```

NOTE: This does not deprecate `ImageView.load`.

Why deprecate these functions? I covered a lot of my thoughts on builders [here](https://github.com/coil-kt/coil/pull/267) that I think apply here as well.

Upsides:
- Remove DSL syntax
- Java compatibility

Downsides:
- Type safety (`data(Any)`)

Overall, I think this makes it clearer that requests are value objects that can be constructed now and executed later:

```kotlin
val request = LoadRequest.Builder(context)
    .transformations(GrayscaleTransformation())
    .build()

// Then later...
val disposable = imageLoader.launch(request.newBuilder().target(imageView).build())
```

Open questions:
- Should there be a `target` + `launch` function similar to Glide and Picasso's `into`?
- Is it clear enough that you need to call `data` when constructing a request?

### DIFF
--- a/coil-base/src/main/java/coil/ImageLoader.kt
+++ b/coil-base/src/main/java/coil/ImageLoader.kt
@@ -14,7 +14,7 @@ import coil.request.RequestDisposable
 import coil.target.Target
 
 /**
- * Loads images using [launch] and [launch].
+ * A service class that fetches and decodes image data.
  */
 interface ImageLoader {
 
@@ -91,7 +91,7 @@ interface ImageLoader {
 
     /** @see launch */
     @Deprecated(
-        message = "Migrate to launch().",
+        message = "Migrate to launch(request).",
         replaceWith = ReplaceWith("launch(request)"),
         level = DeprecationLevel.ERROR
     )
@@ -99,7 +99,7 @@ interface ImageLoader {
 
     /** @see launch */
     @Deprecated(
-        message = "Migrate to launch().",
+        message = "Migrate to launch(request).",
         replaceWith = ReplaceWith("launch(request)"),
         level = DeprecationLevel.ERROR
     )
--- a/coil-base/src/main/java/coil/ImageLoader.kt
+++ b/coil-base/src/main/java/coil/ImageLoader.kt
@@ -12,7 +12,11 @@ import coil.request.RequestDisposable
 import coil.target.Target
 
 /**
- * A service class that fetches and decodes image data.
+ * A service class that loads images by executing [Request]s. Image loaders handle caching, source fetching,
+ * image decoding, request management, bitmap pooling, memory management, and more.
+ *
+ * Image loaders are designed to be shareable and work best when you create a single instance and
+ * share it throughout your app.
  */
 interface ImageLoader {
 
@@ -39,30 +43,28 @@ interface ImageLoader {
     }
 
     /**
-     * The default options for any [Request]s created by this image loader.
+     * The default options that are used to fill in unset [Request] values.
      */
     val defaults: DefaultRequestOptions
 
     /**
-     * Launch an asynchronous operation that executes the [request] and sets the result on its [Target].
-     *
-     * If the request's target is null, this method preloads the image.
+     * Launch an asynchronous operation that executes the [LoadRequest] and sets the result on its [Target].
      *
      * @param request The request to execute.
      * @return A [RequestDisposable] which can be used to cancel or check the status of the request.
      */
-    fun launch(request: LoadRequest): RequestDisposable
+    fun execute(request: LoadRequest): RequestDisposable
 
     /**
-     * Executes the [request] and suspend until the operation is complete. Return the loaded [Drawable].
+     * Suspends and executes the [GetRequest]. Returns the loaded [Drawable] when complete.
      *
      * @param request The request to execute.
      * @return The [Drawable] result.
      */
-    suspend fun launch(request: GetRequest): Drawable
+    suspend fun execute(request: GetRequest): Drawable
 
     /**
-     * Completely clear this image loader's memory cache and bitmap pool.
+     * Clear this image loader's memory cache and bitmap pool.
      */
     @MainThread
     fun clearMemory()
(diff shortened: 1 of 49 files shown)


coil-base/src/main/java/coil/ImageLoader.kt
```diff
@@ -43,6 +45,16 @@ interface ImageLoader {
...
 
+    /**
+     * Create a new [LoadRequestBuilder].
+     */
+    fun load(context: Context) = LoadRequestBuilder(context, this)
+
+    /**
+     * Create a new [GetRequestBuilder].
+     */
+    fun get() = GetRequestBuilder(this)
```
colinrtwhite: These are declared as part of the interface instead of extension functions so they're more discoverable and are easily callable from Java.

coil-base/src/main/java/coil/ImageLoader.kt
```diff
@@ -51,15 +63,15 @@ interface ImageLoader {
      * @param request The request to execute.
      * @return A [RequestDisposable] which can be used to cancel or check the status of the request.
      */
-    fun load(request: LoadRequest): RequestDisposable
+    fun launch(request: LoadRequest): RequestDisposable
```
Jawnnypoo: Could be similar to OkHttp and do `enqueue` but `launch` is good too.

Jawnnypoo:
> Is it clear enough that you need to call data when constructing a request?
Is there a way to "block" the call until they call `data`, ie. they cannot continue with the chain until they do so? Do we throw exceptions if `data` is not called? That would probably be best, so that they know right away that they are not passing the data correctly.

coil-base/src/main/java/coil/ImageLoader.kt
colinrtwhite: Good point. I mostly landed on `launch` since it's the keyword to start a new coroutine. Also feels weird to me to have `enqueue`, but no `execute`.

Jawnnypoo: Very true. Makes sense to me!

colinrtwhite:
@Jawnnypoo Yep, ideally we'd want to force `data` to be called as well. Unfortunately, I don't think it's possible to force it at compile time without adding an intermediary class that creates the `Request`.

Verifying it's called at runtime can be done one of two ways:
- Throw an exception as soon as `build` is called. This is how `okhttp3.Request` works as it requires you to call `url` before `build`. `data(null)` is technically correct usage so we'll have to check that the method was called instead of `data` being null. This also throws the exception at the callsite which is more explicit, but could be unexpected and/or cause crashes.
- Throw an exception when the request is launched. [This is how it works already](https://github.com/coil-kt/coil/blob/master/coil-base/src/main/java/coil/RealImageLoader.kt#L168) if `Request.data` is null. It also throws the exception into the coroutine scope so it won't crash, calls `onError`, and sets the `fallback` drawable. The drawback is this is more "silent".

I'm leaning towards the second since it's easier to go from 2 -> 1 (more restrictive) than it is to go from 1 -> 2 (less restrictive) if we decide we want to change. Let me know what you think!

Jawnnypoo:
Ahhh yes, option 2 makes much more sense based on what is in place currently 👍

coil-base/src/main/java/coil/ImageLoader.kt
```diff
@@ -51,15 +63,15 @@ interface ImageLoader {
...
 
     /**
      * Load the [request]'s data and suspend until the operation is complete. Return the loaded [Drawable].
      *
      * @param request The request to execute.
      * @return The [Drawable] result.
      */
-    suspend fun get(request: GetRequest): Drawable
+    suspend fun launch(request: GetRequest): Drawable
```
swankjesse: Is it idiomatic to overload coroutines names like launch? Is there a situation where this could be ambiguous?

Otherwise I like this name,  though execute() would also work! With coroutines `execute()` is the suspending form and you never need to choose between blocking and non-blocking!

swankjesse: Another word for this is `load()` since that’s what an ImageLoader does.

colinrtwhite: @swankjesse Great point. Overloading `launch` could be confusing. There might be cases where you'd do something like:

```kotlin
CoroutineScope(Dispatchers.Main.immediate).launch {
    val request = GetRequest.Builder().data(url).build()
    val drawable = imageLoader.launch(request)
}
```

`execute` doesn't collide with anything (that I'm aware of) and is usable for both `GetRequest`s and `LoadRequest`s.

`load` works well for `LoadRequest`s, but I think it's awkward to also `load` a `GetRequest`. I'm currently using `ImageLoader.load` and `ImageLoader.get` to execute requests but I'm starting to think having one name to execute both is preferrable...

gabrielittner:
> Should there be a target + launch function similar to Glide and Picasso's into?

Did you make a decision on this?

colinrtwhite:
@gabrielittner Yep, though definitely still looking for feedback if you agree/disagree. Originally in this PR request creation was going to be coupled with the `ImageLoader` - something like:

```kotlin
val disposable = imageLoader.load(context)
    .target(imageView)
    .execute()
```

However, I ended up revising the PR based on feedback to keep the request separate from the `ImageLoader`:

```kotlin
val request = LoadRequest.Builder(context)
    .target(imageView)
    .build()
val disposable = imageLoader.execute(request)
```

Since the `RequestBuilder` doesn't have a reference to an `ImageLoader` any more it doesn't make sense to have a target + execute function.

PR #335, opened 2020-03-31
Inline assertNotShutdown.

colinrtwhite:


coil-base/src/main/java/coil/RealImageLoader.kt
```diff
@@ -484,7 +484,6 @@ internal class RealImageLoader(
 
     override fun invalidate(key: String) = memoryCache.invalidate(key)
 
-    @Synchronized
     override fun shutdown() {
         if (isShutdown) return
         isShutdown = true
```
colinrtwhite: This is safe because `shutdown` should only be called from the main thread. This will be enforced in a later release.

PR #340, opened 2020-04-01
Fix links in FAQ

scana:
They were pointing to root directory instead of `docs/`


colinrtwhite:
Thanks, but I think we actually need to point to root. If I test out your change locally (with `mkdocs serve`) the link redirects to `/faq/getting_started.md/#preloading` which is invalid.

PR #349, opened 2020-04-05
Return sealed RequestResult type from ImageLoader.execute(GetRequest).

### LINKED ISSUE #278: [FR] Expand return type on imperative APIs
Currently all of the imperative APIs return a `Drawable`, which is great for drawing but it would be handy to know more information about the load itself. Things like: where did the image come from? (cache? network?)

I imagine something like the transition API:

``` kotlin
enum class Source { DISK, MEMORY_CACHE, NETWORK }

sealed class LoadResult {
    class Success(
        val drawable: Drawable,
        val source: Source
    ) : LoadResult()

    class Error(
        val drawable: Drawable?
    ) : LoadResult()
}
    
fun Coil.getResult(data: Any): LoadResult
```

This would be super useful for when Coil is being used in outside of View [environments](https://github.com/chrisbanes/tivi/blob/cb/episode-compose/common-ui-compose/src/main/java/app/tivi/common/compose/imageloading.kt) 🚀

WDYT?

colinrtwhite:
This has two main benefits:

- `imageLoader.execute(GetRequest)` is now safe by default. Consumers must handle the success/error case. This is more intuitive than having to surround the call in a try/catch.
- We can provide more info in the returned value (e.g. the source of the image).

This PR also adds `DataSource.MEMORY_CACHE` to indicate if a request is fulfilled from an `ImageLoader`'s internal memory cache.

Fixes #278.


coil-base/src/main/java/coil/request/RequestResult.kt
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Indicates that an error occurred while executing the request.
+ *
+ * @param drawable The error drawable.
+ * @param throwable The error that failed the request.
+ */
+data class ErrorResult(
+    override val drawable: Drawable?,
+    val throwable: Throwable
+) : RequestResult()
```
Jawnnypoo, line 36: Are these still usable in Java?

colinrtwhite: Yep! Sealed classes are Java compatible.

PR #357, opened 2020-04-09
Fix memory cache miss when Transformation reduces input bitmap's size.

colinrtwhite:
This fixes a bug where a request could miss the memory cache if it uses a transformation that decreases the output bitmap's height (e.g. `CircleCropTransformation`). It's reproducible on master if you add a `CircleCropTransformation` in `ImageListAdapter`.

This requires a broader refactor to convert the memory cache's key from `String` to a custom object so we can hold onto the request's resolved size. Overall, I've been planning to move away from using a String cache key for a while as it has gotchas (like `Parameters` having to always order its entries the same way irrespective of insertion order).

Added regression tests to guard against this going forward.

### DIFF
--- a/coil-base/src/main/java/coil/memory/MemoryCache.kt
+++ b/coil-base/src/main/java/coil/memory/MemoryCache.kt
@@ -7,10 +7,16 @@ import android.content.ComponentCallbacks2.TRIM_MEMORY_UI_HIDDEN
 import android.graphics.Bitmap
 import android.util.Log
 import androidx.collection.LruCache
+import coil.fetch.Fetcher
+import coil.memory.MemoryCache.Key
 import coil.memory.MemoryCache.Value
+import coil.request.Parameters
+import coil.size.Size
+import coil.transform.Transformation
 import coil.util.Logger
 import coil.util.getAllocationByteCountCompat
 import coil.util.log
+import coil.util.mapIndices
 
 /** An in-memory cache for [Bitmap]s. */
 internal interface MemoryCache {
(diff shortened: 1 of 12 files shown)


coil-base/src/main/java/coil/memory/MemoryCache.kt
```diff
@@ -43,14 +49,80 @@ internal interface MemoryCache {
...
+    fun invalidate(key: Key)
 
     /** Remove all values from this cache. */
     fun clearMemory()
 
     /** @see ComponentCallbacks2.onTrimMemory */
     fun trimMemory(level: Int)
 
+    /** Cache key for [MemoryCache] and [WeakMemoryCache]. */
+    class Key {
```
colinrtwhite, line 61: No data class, unfortunately, since we need to be able to control when `size` is null.

coil-base/src/main/java/coil/request/Parameters.kt
```diff
@@ -62,14 +85,14 @@ class Parameters private constructor(
 
     class Builder {
 
-        private val map: SortedMap<String, Entry>
+        private val map: MutableMap<String, Entry>
```
Jawnnypoo, line 88: Why was this sorted before?

colinrtwhite: @Jawnnypoo The iteration order had to be consistent irrespective of insertion order so the cache key would match up. If we didn't sort by keys, the parameters would produce two difference cache keys. E.g.

```kotlin
val parameters = Parameters.Builder()
    .set("key1", "first_value")
    .set("key2", "second_value")
    .build()
// would produce the string cache key "key1=first_value#key2=second_value"

val parameters = Parameters.Builder()
    .set("key2", "second_value")
    .set("key1", "first_value")
    .build()
// would produce the string cache key "key2=second_value#key1=first_value"
```

One of the gotchas of using a string cache key. Now order doesn't matter since `Parameters` are compared using `Map.equals`!

Jawnnypoo: Ahhhh gotcha. Thanks for clarifying!

PR #380, opened 2020-04-30
Convert SparseIntArraySet.size to a val.

colinrtwhite:
This is more in line with Kotlin's collections.

### DIFF
--- a/coil-base/src/main/java/coil/collection/SparseIntArraySet.kt
+++ b/coil-base/src/main/java/coil/collection/SparseIntArraySet.kt
@@ -16,22 +16,25 @@ import coil.util.growAndInsert
 class SparseIntArraySet @JvmOverloads constructor(initialCapacity: Int = 10) {
 
     private var elements = IntArray(initialCapacity)
-    private var size = 0
+    private var _size = 0
 
-    /** Adds an element to the set. */
+    /** Returns the number of elements that this set currently stores. */
+    val size: Int get() = _size
+
+    /** Adds the element to the set. */
     fun add(element: Int): Boolean {
-        val i = elements.binarySearch(element, toIndex = size)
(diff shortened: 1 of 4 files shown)


coil-base/src/main/java/coil/extension/SparseIntArraySets.kt
```diff
@@ -5,31 +5,48 @@ package coil.extension
...
+/** Performs the given [action] for each element in the set. */
 inline fun SparseIntArraySet.forEach(action: (element: Int) -> Unit) {
-    for (index in 0 until size()) {
+    for (index in 0 until size) {
         action(elementAt(index))
     }
 }
+
+/** Return an iterator over the set's values. */
+operator fun SparseIntArraySet.iterator(): IntIterator = object : IntIterator() {
```
colinrtwhite, line 48: Added this as `SparseIntArray` ktx has it.

coil-base/src/main/java/coil/collection/SparseIntArraySet.kt
```diff
@@ -16,45 +16,52 @@ import coil.util.growAndInsert
 class SparseIntArraySet @JvmOverloads constructor(initialCapacity: Int = 10) {
 
     private var elements = IntArray(initialCapacity)
-    private var size = 0
+    private var _size = 0
 
-    /** Adds an element to the set. */
+    /** Returns the number of elements that this set currently stores. */
+    val size: Int @JvmName("size") get() = _size
```
Jawnnypoo, line 22: Does `JvmName` just allow it to be a method for Java callers? Or is it needed otherwise?

colinrtwhite: @Jawnnypoo It sets the callable name in Java so Java callers call `set.size()` instead of `set.getSize()`.

PR #385, opened 2020-05-02
Proxy system callbacks to RealImageLoader.

colinrtwhite:
Proxy `onTrimMemory` and `onConnectivityChange` callbacks through a `SystemCallbacks` class. This allows us to avoid the Android framework keeping a strong reference to the `ImageLoader`. This makes `ImageLoader.shutdown` optional - though still recommended.


coil-base/src/main/java/coil/network/NetworkObserver.kt
```diff
@@ -1,51 +1,146 @@
...
 
 /**
- * This class caches the network state from [NetworkObserverStrategy].
+ * Observes the device's network state and calls [Listener] if any state changes occur.
  *
- * Instances should be cleaned up with [shutdown].
+ * This class provides a raw stream of updates from the network APIs. The [Listener] can be
+ * called multiple times for the same network state.
  */
-internal class NetworkObserver(
```
colinrtwhite, line 13: This diff isn't accurate as `NetworkObserver` was deleted and `NetworkObserverStrategy` was renamed to `NetworkObserver`.

coil-base/src/main/java/coil/network/NetworkObserver.kt
```diff
@@ -1,51 +1,146 @@
...
+    private val listener: Listener
+) : NetworkObserver {
+
+    private val networkCallback = object : ConnectivityManager.NetworkCallback() {
+        override fun onAvailable(network: Network) = onConnectivityChange(network, true)
+        override fun onLost(network: Network) = onConnectivityChange(network, false)
+    }
+
+    override val isOnline: Boolean
+        get() = connectivityManager.allNetworks.any { it.isOnline() }
```
Jawnnypoo, line 89: Will this throw an exception if the permission isn't allowed, or is it safe?

colinrtwhite: It should be safe since we ensure we have the permissions [here](https://github.com/coil-kt/coil/pull/385/files#diff-6a3f731cb10b6dd03ddc46d173999f1dR37).

PR #395, opened 2020-05-08
Interrupt thread while decoding if coroutine is cancelled.

### LINKED ISSUE #378: Support thread interruption in cancel
I'm working on a file manager with fast scroll support. In a directory with hundreds of images, if the user fast scrolls the list of files, the thumbnail loading isn't as smooth as what Glide used to provide. I think this is because Coil's decoders (mainly `BitmapFactoryDecoder`) isn't getting interrupted upon cancel, which I heard Glide does (via `Future.cancel(true)`).

Coil decoders should interrupt its thread upon job cancellation. An implementation of this has been proposed at Kotlin/kotlinx.coroutines#1947, however I'm not sure when it will land, so maybe we should also consider rolling our own since this is important for smooth image loading.

**Additional context**
I didn't find any usage of `Future` or `interrupt` in this project, and I also looked at `BitmapFactoryDecoder` where I didn't find anything special, so I'm assuming that the current Coil job cancellation is just a coroutine cancenllation. Sorry in advance if I overlooked anything, because I'm not a coroutine expert yet.

colinrtwhite:
This allows for much faster cancellation as we can cancel in-flight decode operations.

Fixes #378

### DIFF
--- a/coil-base/src/main/java/coil/decode/BitmapFactoryDecoder.kt
+++ b/coil-base/src/main/java/coil/decode/BitmapFactoryDecoder.kt
@@ -14,6 +14,7 @@ import coil.size.PixelSize
 import coil.size.Size
 import coil.util.toDrawable
 import coil.util.toSoftware
+import kotlinx.coroutines.runInterruptible
 import okio.Buffer
 import okio.BufferedSource
 import okio.ForwardingSource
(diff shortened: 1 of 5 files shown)


Jawnnypoo:
Do we still need the deferred stuff too? https://github.com/coil-kt/coil/blob/9930c2a1fd9258607dbdf2759d2b07823cd7b264/coil-base/src/main/java/coil/RealImageLoader.kt#L253

colinrtwhite:
> Do we still need the deferred stuff too?
> 
> https://github.com/coil-kt/coil/blob/9930c2a1fd9258607dbdf2759d2b07823cd7b264/coil-base/src/main/java/coil/RealImageLoader.kt#L253

Yep, `invokeOnCompletion` will still be called if the thread is interrupted. We also still need all the logic in `invokeOnCompletion` since it handles setting the error drawable, some bitmap pooling work, notifying listeners, etc.

colinrtwhite:
Seeing a performance regression when testing this. Going to hold off merging for the moment and investigate.

colinrtwhite:
I think the performance regression I was seeing on my Pixel 1 was a local config issue and haven't been able to reproduce. Going to merge this in and keep an eye on it.

Issue #387, opened 2020-05-03
Loading images into "gone" Views cause downsampling to 1x1 BitmapDrawable


littledot:
**Describe the bug**

Consider the following layout:
```xml
<FrameLayout xmlns:android="http://schemas.android.com/apk/res/android"
    android:layout_width="match_parent"
    android:layout_height="match_parent"
    android:orientation="vertical">

    <ImageView
        android:id="@+id/img1"
        android:layout_width="match_parent"
        android:layout_height="match_parent"
        android:layout_gravity="center"
        android:visibility="gone" />
</FrameLayout>
```
When Coil loads an image into a View whose visibility is "gone", it will downsample the image into a 1x1 BitmapDrawable.
```kotlin
class MainActivity : AppCompatActivity() {

    override fun onCreate(savedInstanceState: Bundle?) {
        super.onCreate(savedInstanceState)
        setContentView(R.layout.main)

        val src =
            "https://pbs.twimg.com/profile_images/378800000666093975/cbe970ce9a19d2c6bec0d06c6004ccf0_400x400.png"

        Coil.loadAny(this, src) {
            target(img1)
            listener { request, source ->
                img1.visibility = View.VISIBLE
            }
        }
    }
}
```

**Expected behavior**

I think other libraries won't even start loading the image until the target View has valid dimensions.

I think it has something to do with this [size resolver.](https://github.com/coil-kt/coil/blob/master/coil-base/src/main/java/coil/size/ViewSizeResolver.kt#L62)

**To Reproduce**

Run the project: https://gitlab.com/littledot/android-bug-report

**Logs/Screenshots**

![](https://gitlab.com/littledot/android-bug-report/-/raw/master/screenshot/coil387.png)

**Version**
Coil 0.10.1

colinrtwhite:
Thanks for the repro project! I think this is due to `ViewSizeResolver` waiting at most one layout pass for valid dimensions when instead we should wait as many as needed. Will have a fix up soon.

PR #397, opened 2020-05-09
Allow suspending through multiple predraw passes.

colinrtwhite:
Previously, we waited for at most one `onPreDraw` pass for the view to be measured. Now we wait for any number until the view is measured.

Fixes: #387

### DIFF
--- a/coil-base/src/main/java/coil/size/ViewSizeResolver.kt
+++ b/coil-base/src/main/java/coil/size/ViewSizeResolver.kt
@@ -45,14 +45,9 @@ interface ViewSizeResolver<T : View> : SizeResolver {
             val viewTreeObserver = view.viewTreeObserver
 
             val preDrawListener = object : ViewTreeObserver.OnPreDrawListener {
-                private var isResumed = false
-
                 override fun onPreDraw(): Boolean {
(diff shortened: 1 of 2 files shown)


coil-base/src/main/java/coil/size/ViewSizeResolver.kt
```diff
@@ -34,34 +34,26 @@ interface ViewSizeResolver<T : View> : SizeResolver {
...
                 private var isResumed = false
 
                 override fun onPreDraw(): Boolean {
-                    if (!isResumed) {
-                        isResumed = true
+                    if (isResumed) return true
 
+                    val size = getSize(false)
+                    if (size != null) {
+                        isResumed = true
```
littledot, line 55: `isResumed` becomes unnecessary because the listener is removed at the same time, meaning `onPreDraw` will never be called again.

colinrtwhite: Good point - I remember adding it a while back for safety, but considering `androidx.core.view.OneShotPreDrawListener` doesn't guard against `onPreDraw` being called again after removing the listener it's probably safe to remove.

PR #407, opened 2020-05-18
Optimize RealBitmapPool and guard against re-adding the same bitmap.

colinrtwhite:
This PR:
- Removes `SizeConfigStrategy`. This should increase the pool's hit rate on API 19-22. Glide requires this strategy to work around a framework bug with bitmaps with null configs on API 19-22. We can safely remove it since Coil rejects bitmaps with null configs from the pool.
- Simplifies `SizeStrategy` and increases its performance by refactoring away an extra `HashMap`.
- Guards against adding the same bitmap to the pool more than once. It has a very slight performance hit to maintain a `HashSet`, but it makes using the pool much safer.


coil-base/src/main/java/coil/decode/EmptyDecoder.kt
```diff
@@ -14,10 +14,7 @@ import okio.blackholeSink
  */
 internal object EmptyDecoder : Decoder {
 
-    private val result = DecodeResult(
-        drawable = EMPTY_DRAWABLE,
-        isSampled = false
-    )
+    private val result = DecodeResult(EMPTY_DRAWABLE, false)
```
Jawnnypoo, line 17: Could the empty decode result be a static singleton, in case someone needed to check against it?

colinrtwhite: It's actually already a singleton since `EmptyDecoder` is an object. Also `result` is never returned publicly so it should be safe to keep private.

PR #408, opened 2020-05-19
Fix bitmap casting in recipes's wiki snippets

Pavneet-Sing:
Accessing drawable directly and removed the unwanted result object


Pavneet-Sing:
> Thanks!

Glad to contribute!

PR #411, opened 2020-05-21
Check for duplicates after initial checks in RealBitmapPool.

colinrtwhite:


coil-base/src/main/java/coil/collection/LinkedMultimap.kt
```diff
@@ -105,12 +105,12 @@ internal class LinkedMultimap<K, V>(sorted: Boolean = false) {
         entry.next.prev = entry.prev
     }
 
-    private class LinkedEntry<K, V>(@JvmField val key: K?) {
```
colinrtwhite, line 108: Removing these since if we add them here I think it's a slippery slope to add them elsewhere internally.

PR #414, opened 2020-05-23
Support thread interruption while decoding.

colinrtwhite:
Fixes: #378

An initial attempt at supporting thread interruption while decoding.

The implementation is based on [`runInterruptible`](https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-core/jvm/src/Interruptible.kt) from the Coroutines library. The main difference is we need to prevent interruption while reading from a `Source` due to [this reason](https://github.com/coil-kt/coil/issues/378#issuecomment-628225348).

I didn't add interruption support to `VideoFrameFetcher` as it seems like even if we interrupt the thread subsequent requests still seem to wait for the previous request to finish. This is likely a hardware limitation.

~TODO: Benchmark this to see how much overhead `InterruptionSource` adds.~ According to my benchmark it adds negligible (too small to measure) overhead to decoding. In the typical case we're getting and setting a volatile int <20 times which is very fast.

### DIFF
--- a/coil-base/src/main/java/coil/annotation/InternalCoilApi.kt
+++ b/coil-base/src/main/java/coil/annotation/InternalCoilApi.kt
@@ -1,0 +1,15 @@
+package coil.annotation
+
+/**
+ * Marks declarations that are part of Coil's internal API. They should not be used outside of the `coil` package
+ * as their signatures and semantics will change between future releases without any warnings and without providing
+ * any migration aids.
+ */
+@MustBeDocumented
+@Retention(value = AnnotationRetention.BINARY)
+@RequiresOptIn(
+    level = RequiresOptIn.Level.ERROR,
+    message = "This is an internal Coil API that should not be used from outside of the `coil` package. " +
+        "No compatibility guarantees are provided."
+)
+annotation class InternalCoilApi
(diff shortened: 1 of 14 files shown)


coil-base/src/main/java/coil/annotation/InternalCoilApi.kt
```diff
@@ -0,0 +1,15 @@
...
+ * any migration aids.
+ */
+@MustBeDocumented
+@Retention(value = AnnotationRetention.BINARY)
+@RequiresOptIn(
+    level = RequiresOptIn.Level.ERROR,
+    message = "This is an internal Coil API that should not be used from outside of the `coil` package. " +
+        "No compatibility guarantees are provided."
+)
+annotation class InternalCoilApi
```
colinrtwhite, line 15: This is necessary to avoid duplicating the interruption code in `coil-gif` and `coil-svg`. Maybe `withInterruptibleSource` can be public API one day, but probably doesn't make sense.

colinrtwhite:
@Jawnnypoo Ran the change through my benchmarking project and wasn't able to measure a change with enough accuracy so I'd say it has almost no decode time overhead.

Issue #405, opened 2020-05-13
`LoadRequest` and `GetRequest` do not support object equality


chrisbanes:
```
val request1 = GetRequest.Builder(...).data("blah").build()
val request2 = GetRequest.Builder(...).data("blah").build()
assertEquals(request1, request2) // this fails
```

Can they just be data classes?

chrisbanes:
For context, my integration in [Accompanist](https://github.com/chrisbanes/accompanist) heavily depends on equality checks to only request once per request. I added a very bad workaround in https://github.com/chrisbanes/accompanist/pull/4 which just keys off `data` instead, but that it is obviously not great for complex requests.

colinrtwhite:
I'd want to avoid using data classes for the requests as they create copy and component functions that make it difficult to modify the classes in the future without breaking binary compatibility, however we could auto generate `equals` and `hashCode` implementations.

I've mostly avoided adding them so far since it means custom `Target`, `SizeResolver`, `Transformation`, `Transition`, and `Listener` implementations will need to implement equals as well.

Does Compose overall depend on concrete `equals/hashCode` implementations? Is it possible to use `memoize` instead?

chrisbanes:
We are memoizing, keyed on the `request` instance (via `stateFor(key)`). This enables automatic dropping of the cached result if the request changes. But that requires the key to implement equals/hashCode correctly.

colinrtwhite:
Hmm going to think about this for a bit... We could implement a standard `equals/hashCode` for the request classes - that would at least open the door for users to implement their own `Target`, `SizeResolver`, `Transformation`, `Transition`, and `Listener` `equals/hashCode`. Though, this would also require implementing `equals/hashCode` on `Decoder`s + `Fetcher`s since you can set them on a per-request basis. Internally, we'd basically have to implement `equals/hashCode` for all our components.

colinrtwhite:
This is fixed in `0.12.0` and up as `ImageRequest` (which replaces `LoadRequest` and `GetRequest`) has `equals`/`hashCode`/`toString` implementations.

PR #424, opened 2020-06-06
Merge LoadRequest and GetRequest into ImageRequest. Remove aliasKeys.

colinrtwhite:
**This PR merges `LoadRequest` and `GetRequest` into one request type: `ImageRequest`.** `ImageRequest` is basically the same `LoadRequest` except it can be executed asynchronously or imperatively (via `suspend fun`). Check out [here](https://github.com/coil-kt/coil/pull/424/files#diff-e06c426f53a4ab7d67b85d0a8e45eb81R50) for how this affects the `ImageLoader` interface.

For context, `GetRequest` was initially added to restrict setting a target on the request while also returning the result from `suspend fun execute()`. This limitation was artificial as both request types worked similarly internally. Ultimately I think it makes sense to combine the two request types for the following reasons:

- `ImageRequest` is a much clearer name. I considered going with just `Request`, but that overlaps with **a lot** of other libraries (including OkHttp).
- Easier to understand one request type instead of two.
- Less code to maintain.
- `ImageRequest` can have its `Builder` as a nested class instead of having to link to it with a `companion object` function.
- `ImageRequest` is more flexible than `GetRequest` when executed using `suspend fun execute()` as it can support a target.

As there's only one request type, the `ImageLoader` methods need to have different names. I ended up going with `enqueue`/`execute` as that continues the trend of following OkHttp's API, but let me know if you have a better idea for the method names!

There are some gotchas associated with setting a target on a "get" request, however I think it's worth the change as long as it's called out in the `ImageLoader.execute` docs.

---

During the conversion I also removed the `aliasKeys` API. I'm not sure it's used and it's not a great API; I'll add a better alternative to it in a separate PR!


coil-base/src/main/java/coil/ImageLoader.kt
```diff
@@ -52,26 +50,28 @@ import java.io.File
...
      */
-    fun execute(request: LoadRequest): RequestDisposable
+    fun enqueue(request: ImageRequest): RequestDisposable
 
     /**
-     * Suspends and executes the [GetRequest]. Returns either [SuccessResult] or [ErrorResult] depending
-     * on how the request completes.
+     * Execute the [request] in the current coroutine scope.
+     *
+     * NOTE: If [ImageRequest.target] is a [ViewTarget], this request will automatically be cancelled
```
colinrtwhite, line 68: Added a warning for the potential gotcha here.

coil-base/src/main/java/coil/memory/ViewTargetRequestManager.kt
```diff
@@ -67,7 +67,7 @@ internal class ViewTargetRequestManager : View.OnAttachStateChangeListener {
         currentRequestJob = null
 
         pendingClear?.cancel()
-        pendingClear = CoroutineScope(Dispatchers.Main.immediate).launch { setCurrentRequest(null) }
+        pendingClear = GlobalScope.launch(Dispatchers.Main.immediate) { setCurrentRequest(null) }
```
colinrtwhite, line 70: Using `GobalScope` isn't ideal, but it's basically equivalent to `CoroutineScope(Dispatchers.Main.immediate).launch` so might as well save an allocation.

coil-base/src/main/java/coil/request/ImageRequest.kt
```diff
@@ -0,0 +1,769 @@
...
+
+    /** @see Builder.fallback */
+    val fallback: Drawable? get() = getDrawableCompat(fallbackDrawable, fallbackResId)
+
+    @JvmOverloads
+    fun newBuilder(context: Context = this.context) = Builder(this, context)
+
+    override fun equals(other: Any?): Boolean {
+        if (this === other) return true
+        return other is ImageRequest &&
```
colinrtwhite, line 138: Added a proper `equals`/`hashCode`/`toString` to begin addressing https://github.com/coil-kt/coil/issues/405. Still need to add equals/hashCode for a couple other classes.

coil-base/src/main/java/coil/request/ImageRequest.kt
```diff
@@ -0,0 +1,769 @@
...
+import coil.transition.Transition
+import coil.util.EMPTY_DRAWABLE
+import coil.util.getDrawableCompat
+import coil.util.orEmpty
+import kotlinx.coroutines.CoroutineDispatcher
+import okhttp3.Headers
+import okhttp3.HttpUrl
+import java.io.File
+
+class ImageRequest private constructor(
```
colinrtwhite, line 47: This class is largely the same as `LoadRequest`.

coil-base/src/main/java/coil/ImageLoader.kt
```diff
@@ -52,26 +50,28 @@ import java.io.File
...
 
     /**
-     * The default options that are used to fill in unset [Request] values.
+     * The default options that are used to fill in unset [ImageRequest] values.
      */
     val defaults: DefaultRequestOptions
 
     /**
-     * Launch an asynchronous operation that executes the [LoadRequest] and sets the result on its [Target].
+     * Execute the [request] asynchronously.
```
Jawnnypoo, line 58: Is it confusing to use the term `execute` in the docs of the `enqueue` function?

colinrtwhite: Good point, updated the docs.

PR #432, opened 2020-06-14
Introduce public interface for the MemoryCache.

colinrtwhite:
Adds a public `MemoryCache` interface that can be accessed with `imageLoader.memoryCache`. This allows users to query an `ImageLoader`'s memory cache directly. Spec:

```kotlin
interface MemoryCache {

    /** The current size of the cache in bytes. */
    val size: Int

    /** The maximum size of the cache in bytes. */
    val maxSize: Int

    /** Get the [Bitmap] associated with [key]. */
    operator fun get(key: Key): Bitmap?

    /** Remove the [Bitmap] referenced by [key]. */
    fun remove(key: Key): Boolean

    /** Remove all values from the memory cache. */
    fun clear()
}
```

Currently it's read-only (except `remove`/`clear`), but I may add `operator fun set(key: Key, bitmap: Bitmap)` later (need to make sure `set` wouldn't violate any internal assumptions).

`MemoryCache.Key` comes in two variants:
- Simple, which wraps a `String`.
- Complex, only created by Coil's image pipeline so the semantics can be changed in the future if necessary. Returned as part of `SucessResult` and `ImageRequest.Listener.onSuccess`.

Example usage:

```kotlin
    .target(imageView)
    .listener { request, metadata ->
        // Now or later...
        val bitmap = imageLoader.memoryCache[metadata.key]
    }
    .build()
imageLoader.enqueue(request)
```

### DIFF
--- a/coil-base/src/main/java/coil/memory/BitmapReferenceCounter.kt
+++ b/coil-base/src/main/java/coil/memory/BitmapReferenceCounter.kt
@@ -18,8 +18,6 @@ import java.lang.ref.WeakReference
  *
  * This class uses [System.identityHashCode] to determine bitmap identity as it provides a "unique-enough" key
  * for a [Bitmap] and it allows us to avoid using [WeakReference]s.
- *
- * NOTE: This class is not thread safe and its methods should only be called from the main thread.
  */
 internal class BitmapReferenceCounter(
     private val weakMemoryCache: WeakMemoryCache,
@@ -33,6 +31,7 @@ internal class BitmapReferenceCounter(
     /**
      * Increase the reference count for this [Bitmap] by one.
      */
+    @Synchronized
     fun increment(bitmap: Bitmap) {
         val key = bitmap.identityHashCode
         val count = counts[key]
@@ -48,6 +47,7 @@ internal class BitmapReferenceCounter(
      *
      * @return True if [bitmap] was added to [bitmapPool] as a result of this decrement operation.
      */
+    @Synchronized
     fun decrement(bitmap: Bitmap): Boolean {
         val key = bitmap.identityHashCode
         val count = counts[key]
@@ -73,6 +73,7 @@ internal class BitmapReferenceCounter(
      * Mark this bitmap as invalid so it is not returned to the bitmap pool
      * when it is no longer referenced.
      */
+    @Synchronized
     fun invalidate(bitmap: Bitmap) {
         invalidKeys += bitmap.identityHashCode
     }
--- a/coil-base/src/main/java/coil/memory/BitmapReferenceCounter.kt
+++ b/coil-base/src/main/java/coil/memory/BitmapReferenceCounter.kt
@@ -45,7 +45,7 @@ internal class BitmapReferenceCounter(
      *
      * If the reference count is now zero, add the [Bitmap] to [bitmapPool].
      *
-     * @return True if [bitmap] was added to [bitmapPool] as a result of this decrement operation.
+     * @return `true` if [bitmap] was added to [bitmapPool] as a result of this decrement operation.
      */
     @Synchronized
     fun decrement(bitmap: Bitmap): Boolean {
--- a/coil-base/src/main/java/coil/memory/BitmapReferenceCounter.kt
+++ b/coil-base/src/main/java/coil/memory/BitmapReferenceCounter.kt
@@ -19,7 +19,7 @@ import java.lang.ref.WeakReference
  * This class uses [System.identityHashCode] to determine bitmap identity as it provides a "unique-enough" key
  * for a [Bitmap] and it allows us to avoid using [WeakReference]s.
  *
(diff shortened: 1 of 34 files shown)


coil-base/src/main/java/coil/memory/BitmapReferenceCounter.kt
```diff
@@ -33,6 +31,7 @@ internal class BitmapReferenceCounter(
     /**
      * Increase the reference count for this [Bitmap] by one.
      */
+    @Synchronized
     fun increment(bitmap: Bitmap) {
```
colinrtwhite, line 35: This PR also makes `StrongMemoryCache`, `WeakMemoryCache`, and `BitmapReferenceCounter` thread safe (so `MemoryCache` is thread safe. This also also required for another feature coming later!

coil-base/src/main/java/coil/memory/MemoryCache.kt
```diff
@@ -1,251 +1,63 @@
...
-    private class InternalValue(
-        override val bitmap: Bitmap,
-        override val isSampled: Boolean,
-        val size: Int
-    ) : Value
-
-    companion object {
-        private const val TAG = "RealMemoryCache"
+@ExperimentalCoilApi
+interface MemoryCache {
```
colinrtwhite, line 14: The existing `MemoryCache` on master was renamed to `StrongMemoryCache` (because it holds strong references). `MemoryCache` is now a public interface that wraps `StrongMemoryCache` + `WeakMemoryCache`.

coil-base/src/main/java/coil/memory/StrongMemoryCache.kt
```diff
@@ -0,0 +1,176 @@
...
+import coil.annotation.ExperimentalCoilApi
+import coil.memory.MemoryCache.Key
+import coil.memory.RealMemoryCache.Value
+import coil.util.Logger
+import coil.util.allocationByteCountCompat
+import coil.util.log
+
+/** An in-memory cache that holds strong references [Bitmap]s. */
+@OptIn(ExperimentalCoilApi::class)
+internal interface StrongMemoryCache {
```
colinrtwhite, line 19: This is basically a renamed version of `MemoryCache` on master.

coil-base/src/main/java/coil/request/Metadata.kt
```diff
@@ -0,0 +1,19 @@
...
+import coil.memory.MemoryCache
+
+/**
+ * Supplemental information about a successful image request.
+ *
+ * @param key The cache key for the image in the memory cache.
+ *  It is null if the image was not written to the memory cache.
+ * @param dataSource The data source that the image was loaded from.
+ */
+data class Metadata(
```
colinrtwhite, line 16: Not sure `Metadata` is the best name for this... Created this type to be able to return more things as needed in the future through `ImageRequest.Listener.onSuccess`.

coil-base/src/main/java/coil/util/Extensions.kt
```diff
@@ -192,3 +196,31 @@ internal inline fun ComponentRegistry.mapData(data: Any, lazySize: () -> Size):
     }
     return mappedData
 }
+
+@OptIn(ExperimentalCoilApi::class)
+internal inline operator fun MemoryCache.Key.Companion.invoke(
```
colinrtwhite, line 201: Making these "constructors" extension functions to keep the `MemoryCache` interface as simple as possible.

coil-base/src/test/java/coil/memory/StrongMemoryCacheTest.kt
```diff
@@ -0,0 +1,121 @@
...
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+import kotlin.test.assertEquals
+import kotlin.test.assertNotNull
+import kotlin.test.assertNull
+
+@RunWith(RobolectricTestRunner::class)
+@OptIn(ExperimentalCoilApi::class)
+class StrongMemoryCacheTest {
```
colinrtwhite, line 17: `MemoryCacheTest` -> `StrongMemoryCacheTest`. Looks like I fumbled the rename 🤦.

coil-base/src/main/java/coil/request/Metadata.kt
Jawnnypoo: Makes sense to me. I could see there being a nice feature too where you can pass raw data via the request and it gets returned in the metadata. Something like 
```
val request = LoadRequest.Builder(context)
    .additionalData(someObject)
```
similar to what you can do with Android views via `setTag`, and this gets returned via `Metadata` (without type safety). Just a potential idea, not a change request to this PR 😄

colinrtwhite: Good point! OkHttp has a `tag` system as well. Might be a good addition...

PR #439, opened 2020-06-22
Support calling SizeResolvers from a non-main context.

colinrtwhite:


colinrtwhite:
Closing as this adds overhead without much gain.

PR #441, opened 2020-06-23
Remove data modifier from several public classes.

colinrtwhite:
These classes are likely to have attributes added in the future. To be able to change the API safely, they can't be data classes. See: https://jakewharton.com/public-api-challenges-in-kotlin/

### DIFF
--- a/coil-base/src/main/java/coil/util/Extensions.kt
+++ b/coil-base/src/main/java/coil/util/Extensions.kt
@@ -5,6 +5,7 @@ package coil.util
 
 import android.app.ActivityManager
 import android.content.res.Configuration
+import android.graphics.Bitmap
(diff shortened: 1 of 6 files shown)


coil-base/src/main/java/coil/util/Extensions.kt
```diff
@@ -195,3 +201,20 @@ internal inline operator fun MemoryCache.Key.Companion.invoke(
         parameters = parameters.cacheKeys()
     )
 }
+
+@OptIn(ExperimentalCoilApi::class)
+internal fun DefaultRequestOptions.copy(
```
colinrtwhite, line 206: This is private to avoid having to maintain binary compatibility for it. It's also not likely to be used outside of the library.

Issue #435, opened 2020-06-19
How to pause request and resume request like Glide 


WuBingHui:
I have a list.
 when scroll was stoped  to do resume requests ,on the contrary scrolling to do pause requests
On Glide  like this. 
Glide.with(context).pauseRequests()
Glide.with(context).resumeRequests()

colinrtwhite:
It's currently not possible to pause/resume requests globally with Coil.

WuBingHui:
thank you !

PR #447, opened 2020-06-29
Add support for Interceptors.

colinrtwhite:
This PR adds support for interceptors in Coil. Its design is heavily based on the [interceptor design in OkHttp](https://square.github.io/okhttp/interceptors/). Its interface:

```kotlin
interface Interceptor {

    suspend fun intercept(chain: Chain): RequestResult

    interface Chain {

        val request: ImageRequest

        val size: Size

        fun withSize(size: Size): Chain

        suspend fun proceed(request: ImageRequest): RequestResult
    }
}
```

Benefits:
- Significantly greater control of how requests are executed in Coil. Users can use interceptors to:
    - Add their own custom caching layers to short circuit requests.
    - Transform requests (beyond just their data) based on custom logic.
    - Buffer, delay, or apply any other coroutines operators to the request since `intercept` is a suspend function. For example, you could use an interceptor to [globally pause/resume requests](https://github.com/coil-kt/coil/issues/435) using a `StateFlow`.
- Consistent event ordering. `LazySizeResolver` is removed and request `Size` is always resolved up front.
- Performance improvements by moving the memory cache check to a background thread.
- Internally, this makes the code much cleaner - especially in `RealImageLoader`.

Downsides:
- The memory cache check no longer occurs on the main thread. This means cached images aren't resolved synchronously (though they already weren't for a couple cases). This means automatic image sampling is no more :(, however this PR also introduces `placeholderKey`, which lets you use a memory cache key as the placeholder. `placeholderKey` is always resolved synchronously.

To make synchronous placeholders easy (though not completely free like before), `metadata` for a successful request is now attached to a view. So you can do something like this:

```kotlin
// In list view
listImage.load("https://www.example.com/image.jpg")

// In detail view
detailImage.load("https://www.example.com/larger_image.jpg") {
    placeholderKey(listImage.metadata?.cacheKey)
}
```


coil-base/src/main/java/coil/ImageLoader.kt
```diff
@@ -261,18 +259,6 @@ interface ImageLoader {
...
 
-        /**
-         * Enables adding [File.lastModified] to the memory cache key when loading an image from a [File].
-         *
-         * This allows subsequent requests that load the same file to miss the memory cache if the file has been updated.
-         * To ensure a cached image is returned synchronously, [File.lastModified] is called on the main thread.
-         *
-         * Default: true
-         */
-        fun addLastModifiedToFileCacheKey(enable: Boolean) = apply {
```
colinrtwhite, line 272: This isn't required any more since cache key computation is done on a background thread.

coil-base/src/main/java/coil/ComponentRegistry.kt
```diff
@@ -12,71 +15,103 @@ import coil.util.MultiMutableList
...
+        }
+
         /** Register a [Mapper]. */
-        inline fun <reified T : Any> add(mapper: Mapper<T, *>) = add(T::class.java, mapper)
+        inline fun <reified T : Any> add(mapper: Mapper<T, *>) = add(mapper, T::class.java)
 
         @PublishedApi
-        internal fun <T : Any> add(type: Class<T>, mapper: Mapper<T, *>) = apply {
-            mappers += type to mapper
+        internal fun <T : Any> add(mapper: Mapper<T, *>, type: Class<T>) = apply {
```
colinrtwhite, line 65: Changed the parameter order so the mapper/fetcher is `component0` instead of `component1`. This change is transparent for Kotlin users.

coil-base/src/main/java/coil/EventListener.kt
```diff
@@ -30,50 +29,41 @@ import coil.transition.TransitionTarget
...
 
     /**
-     * Called before any [Mapper]s and/or [MeasuredMapper]s are called to convert the request's data.
-     *
-     * @param input The data that will be converted.
+     * Called before [SizeResolver.size].
      */
     @MainThread
-    fun mapStart(request: ImageRequest, input: Any) {}
+    fun resolveSizeStart(request: ImageRequest) {}
```
colinrtwhite, line 41: `input` is now always equal to `request.data` now that `request.data` is not null.

coil-base/src/main/java/coil/fetch/HttpUrlFetcher.kt
```diff
@@ -16,40 +14,17 @@ import okhttp3.HttpUrl
...
-
-    override fun handles(data: Uri) = data.scheme == "http" || data.scheme == "https"
-
-    override fun key(data: Uri) = data.toString()
-
-    override fun Uri.toHttpUrl(): HttpUrl = HttpUrl.get(toString())
-}
-
-internal class HttpUrlFetcher(callFactory: Call.Factory) : HttpFetcher<HttpUrl>(callFactory) {
+internal class HttpUrlFetcher(private val callFactory: Call.Factory) : Fetcher<HttpUrl> {
```
colinrtwhite, line 17: Removed this optimization and re-added `HttpUriMapper` since mappers are now called on a background thread.

coil-base/src/main/java/coil/interceptor/EngineInterceptor.kt
```diff
@@ -0,0 +1,333 @@
...
+import coil.util.takeIf
+import coil.util.toDrawable
+import kotlinx.coroutines.CancellationException
+import kotlinx.coroutines.ensureActive
+import kotlin.coroutines.coroutineContext
+import kotlin.math.abs
+
+/** The last interceptor in the chain which executes the [ImageRequest]. */
+@OptIn(ExperimentalCoilApi::class)
+internal class EngineInterceptor(
```
colinrtwhite, line 55: This is a combination of existing code from `RealImageLoader` and `MemoryCacheService`. It takes in an `ImageRequest` and returns a `RequestResult` without handling lifecycles or setting the result on the target (that's still handled in `RealImageLoader`.

coil-base/src/main/java/coil/memory/MemoryCache.kt
```diff
@@ -12,7 +11,6 @@ import kotlinx.android.parcel.Parcelize
 /**
  * An in-memory cache of recently loaded images.
  */
-@ExperimentalCoilApi
```
colinrtwhite, line 15: Removing the experimental status on `MemoryCache` so `placeholderKey` doesn't have to be experimental.

coil-base/src/main/java/coil/memory/ViewTargetRequestManager.kt
```diff
@@ -30,14 +31,17 @@ internal class ViewTargetRequestManager : View.OnAttachStateChangeListener {
     @Volatile var currentRequestJob: Job? = null
         private set
 
+    // Metadata of the successful request attached to this view.
+    @Volatile var metadata: Metadata? = null
```
colinrtwhite, line 35: Caching the metadata on the view so it can be grabbed with `imageView.metadata.cacheKey`.

coil-base/src/test/java/coil/RealImageLoaderBasicTest.kt
```diff
@@ -1,299 +0,0 @@
...
-@RunWith(RobolectricTestRunner::class)
-@OptIn(ExperimentalCoilApi::class)
-class RealImageLoaderBasicTest {
-
-    private lateinit var context: Context
-    private lateinit var strongMemoryCache: StrongMemoryCache
-    private lateinit var imageLoader: RealImageLoader
-
-    @Before
-    fun before() {
```
colinrtwhite, line 61: Most of these tests were moved around to different classes, but still exist.

coil-base/src/main/java/coil/interceptor/EngineInterceptor.kt
Jawnnypoo: Not sure I understand the naming convention, but makes sense 👍

coil-base/src/main/java/coil/map/MeasuredMapper.kt
```diff
@@ -14,6 +14,7 @@ import coil.target.Target
  *
  * @see Mapper
  */
+@Deprecated("Superseded by `coil.interceptor.Interceptor` which offers a more versatile API.")
```
Jawnnypoo, line 17: Probably no straightforward way to do a `ReplaceWith`, eh?

coil-base/src/main/java/coil/memory/RequestService.kt
```diff
@@ -158,18 +84,18 @@ internal class RequestService(
...
     private fun isConfigValidForTransformations(request: ImageRequest): Boolean {
-        return request.transformations.isEmpty() || (request.bitmapConfig ?: defaults.bitmapConfig) in VALID_TRANSFORMATION_CONFIGS
+        return request.transformations.isEmpty() || request.bitmapConfig in VALID_TRANSFORMATION_CONFIGS
     }
 
     companion object {
-        /** A whitelist of valid bitmap configs for the input and output bitmaps of [Transformation.transform]. */
-        @JvmField internal val VALID_TRANSFORMATION_CONFIGS = if (SDK_INT >= 26) {
+        /** An allowlist of valid bitmap configs for the input and output bitmaps of [Transformation.transform]. */
+        @JvmField val VALID_TRANSFORMATION_CONFIGS = if (SDK_INT >= 26) {
```
Jawnnypoo, line 98: Purposefully made public?

coil-base/src/main/java/coil/request/DefaultRequestOptions.kt
```diff
@@ -32,6 +33,22 @@ class DefaultRequestOptions(
...
+        allowRgb565: Boolean = this.allowRgb565,
+        placeholder: Drawable? = this.placeholder,
+        error: Drawable? = this.error,
+        fallback: Drawable? = this.fallback,
+        memoryCachePolicy: CachePolicy = this.memoryCachePolicy,
+        diskCachePolicy: CachePolicy = this.diskCachePolicy,
+        networkCachePolicy: CachePolicy = this.networkCachePolicy
+    ) = DefaultRequestOptions(dispatcher, transition, precision, bitmapConfig, allowHardware, allowRgb565, placeholder,
+        error, fallback, memoryCachePolicy, diskCachePolicy, networkCachePolicy)
+
```
Jawnnypoo, line 51: Can't make this a data class I guess?

coil-base/src/main/java/coil/map/MeasuredMapper.kt
colinrtwhite: Unfortunately the replace is a little too complex for autocomplete, but a `MeasuredMapper` -> `Interceptor` conversion will end up looking like this:

```kotlin
class SizeInterceptor : Interceptor {

    suspend fun intercept(chain: Chain): RequestResult {
        val request = chain.request
        if (handles(request)) {
            val newRequest = request.newBuilder().data(map(request.data, chain.size)).build()
            return chain.proceed(newRequest)
        } else {
            return chain.proceed(request)
        }
    }
}
```

coil-base/src/main/java/coil/memory/RequestService.kt
colinrtwhite: It's actually still `internal` since `RequestService` is internal.

coil-base/src/main/java/coil/request/DefaultRequestOptions.kt
colinrtwhite: Technically this could be a `data` class, however I'm starting to avoid `data` for public api classes due to the reasons [here](https://jakewharton.com/public-api-challenges-in-kotlin/). Mostly want to avoid creating component functions so I can still control property ordering in the constructor in future versions.

coil-base/src/main/java/coil/interceptor/EngineInterceptor.kt
colinrtwhite: Ah yeah I was going for a [software engine](https://en.wikipedia.org/wiki/Software_engine), though definitely could be improved (open to suggestions!). Fortunately, it's not public API so it can be changed in the future.

Jawnnypoo: Ahhh, nice, I like that.

coil-base/src/main/java/coil/memory/RequestService.kt
Jawnnypoo: Oh, my bad!

coil-base/src/main/java/coil/request/DefaultRequestOptions.kt
Jawnnypoo: Ah good call 👍

colinrtwhite:
Good call out - going to do a sizable docs update over the next couple days to account for the changes.

PR #453, opened 2020-07-06
Move `coil.api` functions to `coil` and move `coil.extension` functions to respective packages.

colinrtwhite:
- Moves the extensions in `coil.extension` to their correct packages.
- Moves `coil.api.load` -> `coil.load`. This is going to cause a lot of deprecation warnings unfortunately, but I think it's worthwhile to make this change before 1.0.0 as the `coil.api` package is left over from when Coil had a DSL.

Unfortunately we can't use `ReplaceWith()` as it will cause a conflicting import when it adds the new import without removing the old one.


coil-base/src/main/java/coil/extension/Parameters.kt
```diff
@@ -4,12 +4,15 @@
 package coil.extension
 
 import coil.request.Parameters
+import coil.request.count as _count
+import coil.request.get as _get
+import coil.request.isNotEmpty as _isNotEmpty
```
colinrtwhite, line 9: The deprecated functions delegate to the functions in the correct package.

PR #461, opened 2020-07-15
Prefer immutable bitmaps on API 24+.

colinrtwhite:
Talked with a member of the Android graphics team and they recommended preferring immutable bitmaps on Android N+ as [this change](https://cs.android.com/android/_/android/platform/frameworks/base/+/775873a66a946fae2b0535abb51df9817bd1b20c) makes bitmap pooling much less important. Immutable software bitmaps also "don't need to stay pinned in the GPU texture cache like mutable ones do, since we know we're free to re-upload whenever we want/need."

This disables bitmap reuse on N+, however we still call `Bitmap.recycle()` to eagerly free memory when a bitmap's reference count reaches zero. Disabling bitmap reuse doesn't affect decode performance, but will result in creating more bitmaps - though each bitmap will be the exact minimum required size (and they're much cheaper on N+). Overall, I'm following the recommendation of the graphics team here.

Bitmap pool size can still be set manually by consumers using `ImageLoader.Builder.bitmapPoolPercentage` - this PR just changes the default.

This also moves common bitmap pooling-related classes to the `coil.bitmap` package for consistency.

### DIFF
--- a/coil-base/src/main/java/coil/bitmap/BitmapPoolStrategy.kt
+++ b/coil-base/src/main/java/coil/bitmap/BitmapPoolStrategy.kt
@@ -1,4 +1,4 @@
-package coil.bitmappool
+package coil.bitmap
 
 import android.graphics.Bitmap
 import android.os.Build.VERSION.SDK_INT
--- a/coil-base/src/main/java/coil/bitmap/BitmapPoolStrategy.kt
+++ b/coil-base/src/main/java/coil/bitmap/BitmapPoolStrategy.kt
@@ -95,7 +95,7 @@ internal class SizeStrategy : BitmapPoolStrategy {
     override fun toString() = "SizeStrategy: entries=$entries, sizes=$sizes"
 
     companion object {
-        private const val MAX_SIZE_MULTIPLE = 8
+        private const val MAX_SIZE_MULTIPLE = 4
     }
 }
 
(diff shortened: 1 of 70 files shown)


coil-base/src/main/java/coil/bitmap/BitmapPoolStrategy.kt
```diff
@@ -95,7 +95,7 @@ internal class SizeStrategy : BitmapPoolStrategy {
     override fun toString() = "SizeStrategy: entries=$entries, sizes=$sizes"
 
     companion object {
-        private const val MAX_SIZE_MULTIPLE = 8
+        private const val MAX_SIZE_MULTIPLE = 4
```
colinrtwhite, line 98: This means only bitmaps at most 2x the size of the target size will be reused (we wouldn't want to reuse a 4mb bitmap for a 20kb image).

coil-sample/src/main/java/coil/sample/Application.kt
```diff
@@ -21,6 +21,7 @@ class Application : MultiDexApplication(), ImageLoaderFactory {
 
     override fun newImageLoader(): ImageLoader {
         return ImageLoader.Builder(this)
+            .allowHardware(false)
```
Jawnnypoo, line 24: Purposeful change?

colinrtwhite: Good catch! Removing the debug code.

PR #470, opened 2020-07-23
Rename RequestResult and RequestDisposable.

colinrtwhite:
Rename:

`RequestResult` -> `ImageResult`
`RequestDisposable` -> `Disposable`.

`Disposable` is also used by RxJava, however I think it should be ok. Curious what you think!

Also converts `Metadata` to be a data class again. This locks the parameter order, but generating `component` functions for this data type should be ok (property order isn't likely to change).

### DIFF
--- a/coil-base/src/main/java/coil/request/Deprecated.kt
+++ b/coil-base/src/main/java/coil/request/Deprecated.kt
@@ -32,3 +32,15 @@ typealias GetRequestBuilder = ImageRequest.Builder
     replaceWith = ReplaceWith("ImageRequest", "coil.request.ImageRequest")
 )
 typealias Request = ImageRequest
+
+@Deprecated(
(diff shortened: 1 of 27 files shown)


coil-base/src/main/java/coil/request/Deprecated.kt
```diff
@@ -32,3 +32,15 @@ typealias GetRequestBuilder = ImageRequest.Builder
     replaceWith = ReplaceWith("ImageRequest", "coil.request.ImageRequest")
 )
 typealias Request = ImageRequest
+
+@Deprecated(
+    message = "Replace with Disposable.",
+    replaceWith = ReplaceWith("Disposable", "coil.request.Disposable")
+)
+typealias RequestDisposable = Disposable
```
andreygorilla, line 40: really nice

PR #480, opened 2020-07-31
Fix WebP decoding on API < 19 devices

cbeyls:
On older devices (mostly API < 19), WebP decoding is half broken: decoding from an `InputStream` causes the image to be only partially decoded (the bitmap contains transparent areas), or throws a JNI Exception which crashes the app. This can be reproduced at least on the API 17 and API 18 emulators.

On these devices, it turns out that executing the first decoding pass (computing image size and MIME type without actually decoding the image) will return a `null` MIME type. When the MIME type is `null`, we already skip decoding the EXIF metadata.

Switch to decoding the image from a byte array instead of a stream when the MIME type is `null` (which is the best indicator that something is broken with the decoder). This fixes the WebP decoding problems mentioned above on the older devices.


coil-base/src/main/java/coil/decode/BitmapFactoryDecoder.kt
```diff
@@ -162,7 +162,12 @@ internal class BitmapFactoryDecoder(private val context: Context) : Decoder {
         var outBitmap: Bitmap? = null
         try {
             outBitmap = safeBufferedSource.use {
-                BitmapFactory.decodeStream(it.inputStream(), null, this)
+                if (shouldDecodeFromByteArray(outMimeType)) {
```
colinrtwhite, line 165: Can you inline this method here + add an `SDK_INT < 19` check? Thanks!

coil-sample/src/main/assets/webps.json
```diff
@@ -0,0 +1,2966 @@
+[
```
colinrtwhite, line 1: Can you remove these? I don't think we need a separate section for webps. If you're up for it it would be better to add a regression test in `BitmapFactoryDecoderTest` (though no worries if you'd rather not).

coil-base/src/main/java/coil/decode/BitmapFactoryDecoder.kt
cbeyls: I'm not entirely sure this issue never happens on any API 19 device, but everytime I saw it the MIME type was null so maybe as extra safety we could just fallback to byte array decoding every time we get a null MIME type no matter the API version or image format?

coil-sample/src/main/assets/webps.json
cbeyls: Not sure how the test would work. Most of the time the app does not crash but images are just partially decoded. Bitmaps have the correct size so the test would have to look at the pixels directly.

colinrtwhite: Coil's tests have a `Bitmap.isSimilarTo` extension function to test the visual similarity of images. You could use that to compare a jpg with a webp of the same image.

coil-base/src/main/java/coil/decode/BitmapFactoryDecoder.kt
colinrtwhite: Hmm I'd guess that it's probably constrained to < API 19 since there was a lot of work done to the graphics system in API 19. Also I'm not sure we should always fall back to byte array decoding if mime type is null since it has worse performance as it requires us to buffer the entire source into memory before decoding.

cbeyls: Okay, I added the condition to only use byte array decoding on API < 19

coil-sample/src/main/assets/webps.json
cbeyls: I wrote a simple test for lossy WebP decoding. I had a problem because `Bitmap.isSimilarTo` returned true even when the WebP decoding failed on API 17 and the bitmaps were definitely not similar. It turns out there was a bug in the `crossCorrelation()` function so I changed it too.

PR #508, opened 2020-08-26
Implement circle stroke transformation

t-kurimura:
I came up with just an idea, circle transformation with STROKE. We often meet such cases that needs to draw stroke and it's east to be achieved with Glide according to my experience.

This is just proposal, so Im going to fix test and Readme if you agree.
Thank you !

red_stroke_small_circle


colinrtwhite:
Thanks for adding this, however I'd like to avoid adding new transformations to the base coil artifact. I'd check out https://github.com/Commit451/coil-transformations, which is the Coil counterpart to glide-transformations.

PR #509, opened 2020-08-27
Update Kotlin (1.4.0) and Gradle (6.6.1).

colinrtwhite:
NOTE: This does not add the `jvm-default=all` compiler flag which is present on the `1.0.0` branch.


coil-singleton/src/main/java/coil/ImageLoaderFactory.kt
```diff
@@ -9,7 +9,7 @@ import android.app.Application
  * - Implement [ImageLoaderFactory] in your [Application].
  * - **Or** call [Coil.setImageLoader] with your [ImageLoaderFactory].
  */
-interface ImageLoaderFactory {
+fun interface ImageLoaderFactory {
```
Jawnnypoo, line 12: Love this.

Issue #280, opened 2020-02-05
Images inside recyclerview flash when notifyDataSetChanged() is called


dafi:
I'm migrating my code from Picasso to Coil and discovered a strange behaviour non present with Picasso.

A recycler view contains images but when I call `notifyDataSetChanged()` on the adapter some images flash as shown in attached GIF

I created a project useful to reproduce the bug, [CoilFlashing](https://github.com/dafi/coil_flashing)

**Version**
Coil 0.9.4
Android 9 and 10 under Android emulator, Xiaomi MiA1, Pixel 3XL

![screenshoot](https://user-images.githubusercontent.com/84500/73835865-77826e00-480e-11ea-8ad0-378212939fc1.gif)

colinrtwhite:
Thanks for the report + sample project. I just added a fix here: https://github.com/coil-kt/coil/pull/283.

dafi:
0.12.0 contains a regression and the bug popped up again

The problem to reproduce the issue is still valid, you can download and use it, with version 0.11.0 everything works fine

colinrtwhite:
This is likely intended behaviour in `0.12.0` as the memory cache no longer resolves values synchronously on the main thread. This change was made for several reasons:

- Resolving synchronously on the main thread didn't work unexpectedly in some edge cases. For instance, if the request has a `Transformation` the request likely won't check the memory cache synchronously.
- It requires running `Mapper`s on the main thread, which can also be unexpected (since they're part of the image pipeline) and is slower on the main thread.
- `Interceptors` can't run on the main thread without likely causing performance issues.

Both Glide and Picasso also check the memory cache off of the main thread.

mattlogan:
Hey Colin - I understand your reasoning above for checking the cache asynchronously, but I noticed that Picasso doesn't have this issue despite also checking the memory cache off the main thread. I'm not sure I understand why checking the memory cache asynchronously necessitates clearing & reloading the image as it appears to be doing now. Perhaps you could wait to clear the image until the memory cache check completes?

It would be awesome if you have an idea for another solution here, otherwise I think this will lead to a slightly degraded UX for situations like the one described above. Thanks for the awesome library!

colinrtwhite:
[...]

PR #415, opened 2020-05-24
Add option to disable adding File.lastModified to the cache key.

colinrtwhite:
Fixes #383 by adding an option to disable the `lastModified` check.

PR #513, opened 2020-08-29
Launch the Interceptor chain on the main thread by default.

colinrtwhite:
This PR changes the default behaviour of the interceptor chain to be launched on the main thread by default. Why do this:

- It allows the memory cache check to occur synchronously if the request is started from the main thread.
- It allows `Interceptors` to be execute on the main thread synchronously if preferred. `Interceptors` can also control which requests check the memory cache on the main thread by using `withContext` to change the current dispatcher when calling `Interceptor.Chain.proceed`.
- Synchronously checking the memory cache was the default behaviour pre-`0.12.0`. This realigns with that behaviour, but allows consumers to move the check off the main thread if they want maximum performance.

This PR adds two new flags to `ImageLoader.Builder`:

- `launchInterceptorChainOnMainThread`: Enables/disables launching the interceptor chain on the main thread.
- `addLastModifiedToFileCacheKey`: If the memory cache check occurs on the main thread, calling `File.lastModified` can cause a strict mode violation. This flag disables calling `File.lastModified`. See https://github.com/coil-kt/coil/pull/415 for context.

Fixes: #280


coil-base/src/main/java/coil/fetch/HttpFetcher.kt
```diff
@@ -15,17 +17,38 @@ import okhttp3.HttpUrl
 import okhttp3.Request
 import okhttp3.ResponseBody
 
-internal class HttpUrlFetcher(private val callFactory: Call.Factory) : Fetcher<HttpUrl> {
+internal class HttpUriFetcher(callFactory: Call.Factory) : HttpFetcher<Uri>(callFactory) {
```
colinrtwhite, line 20: Restoring the optimization to call `HttpUrl.get` as part of `Fetcher.fetch` instead of part of `Mapper.map` as `Fetcher.fetch` is always called on a background thread.

PR #516, opened 2020-09-01
Replace all (integer / 2) operations to (integer shr 1) for better performance.

phantomVK:


Tolriq:
https://jakewharton.com/which-is-better-on-android-divide-by-two-or-shift-by-one/

phantomVK:
> https://jakewharton.com/which-is-better-on-android-divide-by-two-or-shift-by-one/


Thank you so much for the reply!

PR #530, opened 2020-09-11
Work around compiler bug in Lifecycles.

colinrtwhite:
Noticed this while adding more tests. Currently the "catch" that removes the `observer` won't work if the function is inlined as it will be `null` if the method is inlined.

Working on reproducing this in a small project to file a bug.


Jawnnypoo (approved):
Strange...

PR #537, opened 2020-09-17
Add ImageLoader.executeBlocking extension function.

### LINKED ISSUE #529: Feature Request: Add ImageLoader.executeBlocking Implementation
Yes.  I have a pure Java android project and I'm using rxjava worker threads for background image retrieval.  I would like to fetch images synchronously on these threads with coil.

Rather than [document](https://coil-kt.github.io/coil/java_compatibility/) the (admittedly trivial) solution for performing blocking requests from Java, add this to the core library as `ImageLoader.executeBlocking` and update the documentation to reflect this.

**Additional context**
For now I've added a kotlin module to my project for implementing this behavior, but it seems kind of silly to add an entire kotlin module for this functionality.  If the use case is common enough to call out in the documentation, it seems like it'd make sense to add it to the core library.

In any event, kudos on an amazing library.  I've thoroughly evaluated Glide, Fresco, and Coil, and Coil is already a clear winner.

colinrtwhite:
Adds the `ImageLoader.executeBlocking` extension function to the main artifact.

Also updates some docs from `default image loader` -> `singleton image loader` to be clearer.

Closes #529

### DIFF
--- a/coil-base/api/coil-base.api
+++ b/coil-base/api/coil-base.api
@@ -120,6 +120,10 @@ public final class coil/ImageLoader$Companion {
 	public final fun create (Landroid/content/Context;)Lcoil/ImageLoader;
 }
 
+public final class coil/ImageLoaders {
+	public static final fun executeBlocking (Lcoil/ImageLoader;Lcoil/request/ImageRequest;)Lcoil/request/ImageResult;
+}
+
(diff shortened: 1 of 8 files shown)


Jawnnypoo (approved):
I think this is a good call 👍

Issue #549, opened 2020-10-07
HTTP 404 responses not cached by OkHttp


clhols:
**Describe the bug**
The issue is described here:
https://github.com/square/okhttp/issues/6310

**Expected behavior**
HTTP responses that OkHttp can cache should be cached.

**To Reproduce**
Try any request that returns 404 with a Cache-Control header set to allow caching.

**Version**
1.0.0-rc3

sakayori-daisuke:
有事対応を施した形跡がライブラリ側にあるとし、取り込みを判断
https://github.com/coil-kt/coil/issues/549

PR #551, opened 2020-10-08
Allow OkHttp to cache unsuccessful responses (#549)

clhols:
Fixes: #549

### DIFF
--- a/coil-base/src/test/java/coil/fetch/HttpFetcherTest.kt
+++ b/coil-base/src/test/java/coil/fetch/HttpFetcherTest.kt
@@ -122,7 +122,7 @@ class HttpFetcherTest {
 
     @Test
     fun `not found response is cached`() {
-        val uri = createMockWebServer(context).url("/normal.jpg").toString().toUri()
+        val uri = createMockWebServer(context).url("/notfound.jpg").toString().toUri()
         val fetcher = HttpUriFetcher(callFactory)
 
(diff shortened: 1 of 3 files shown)


coil-base/src/test/java/coil/fetch/HttpFetcherTest.kt
```diff
@@ -113,4 +119,17 @@ class HttpFetcherTest {
...
+    fun `not found response is cached`() {
+        val uri = createMockWebServer(context).url("/normal.jpg").toString().toUri()
+        val fetcher = HttpUriFetcher(callFactory)
+
+        assertFailsWith<HttpException> {
+            runBlocking {
+                fetcher.fetch(pool, uri, PixelSize(100, 100), createOptions(context))
+            }
+        }
+        assertTrue(cache.urls().hasNext())
```
colinrtwhite, line 133: Can you change this to assert that the `url` we fetched is the same as the one in the cache?

colinrtwhite:
Left one comment, but otherwise LGTM. Thanks for reporting + fixing this!

coil-base/src/test/java/coil/fetch/HttpFetcherTest.kt
clhols: Done

colinrtwhite (approved):

PR #556, opened 2020-10-20
Updated documentation around ImageLoaderFactory

matteinn:
The code snippet that demonstrates `ImageLoaderFactory` usage  was referencing a `context` property which is not defined in that scope. Replaced it with the application-scoped `applicationContext`.


colinrtwhite (approved):
Good catch, thanks!

PR #560, opened 2020-10-25
Test against API 30.

colinrtwhite:


colinrtwhite:
The Google API emulators are significantly slower than the non-Google API emulators. It's not worth upgrading to test against API 30 until a non-Google API emulator variant is released for API 30.

PR #561, opened 2020-10-25
Add an empty bitmap pool implementation.

colinrtwhite:
Adds an empty bitmap pool implementation. This should be slightly more efficient on API 24 and up or if `bitmapPoolPercentage` is set to 0 since it avoids locking + all of the pool management code in `RealBitmapPool`.

### DIFF
--- a/coil-base/src/main/java/coil/bitmap/RealBitmapPool.kt
+++ b/coil-base/src/main/java/coil/bitmap/RealBitmapPool.kt
@@ -92,25 +92,24 @@ internal class RealBitmapPool(
         return getDirtyOrNull(width, height, config) ?: createBitmap(width, height, config)
     }
 
+    @Synchronized
     override fun getDirtyOrNull(@Px width: Int, @Px height: Int, config: Bitmap.Config): Bitmap? {
         require(!config.isHardware) { "Cannot create a mutable hardware bitmap." }
 
-        synchronized(this) {
-            val result = strategy.get(width, height, config)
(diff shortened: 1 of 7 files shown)


Jawnnypoo (approved):
Great improvement!

coil-base/src/main/java/coil/bitmap/RealBitmapPool.kt
```diff
@@ -92,24 +92,25 @@ internal class RealBitmapPool(
...
-        if (result == null) {
-            logger?.log(TAG, Log.VERBOSE) { "Missing bitmap=${strategy.stringify(width, height, config)}" }
-            misses++
-        } else {
-            bitmaps -= result
-            currentSize -= result.allocationByteCountCompat
-            hits++
-            normalize(result)
-        }
+        synchronized(this) {
```
Jawnnypoo, line 98: Whats the advantage to using this `synchronized` vs. the annotation on the method?

colinrtwhite: It's pretty minor, but using `synchronized` lets us grab the lock after we've checked if the config is `HARDWARE`. `@Synchronized` is just syntactic sugar that wraps the method in a `synchronized` block

colinrtwhite: Actually, ended up reverting this part since the benefit is probably imperceptibly small.

Issue #568, opened 2020-10-28
Add the ability to control alpha pre-multiplication


romainguy:
Android loads Bitmaps in premultiplied format by default (correlating the alpha channel with the rgb channels). This is however not desirable in a number of situations (lookup tables, channel packing, data textures, etc.), especially for games and 3D engines. It would be useful to have an API to control whether a bitmap loaded by an image request is premultiplied or not.

PR #569, opened 2020-10-29
Add ability to control alpha pre-multiplication

romainguy:
The new `ImageRequest` API maps to the equivalent API in `BitmapFactory.Options` and `ImageDecoder`.
This PR addresses issue #568.


coil-base/src/main/java/coil/request/DefaultRequestOptions.kt
```diff
@@ -26,6 +26,7 @@ class DefaultRequestOptions(
     val bitmapConfig: Bitmap.Config = Utils.DEFAULT_BITMAP_CONFIG,
     val allowHardware: Boolean = true,
     val allowRgb565: Boolean = false,
+    val premultipliedAlpha: Boolean = true,
```
colinrtwhite, line 29: I would only add this attribute to `ImageRequest` and `Options` unless you feel there's a good use case for setting the `ImageLoader`'s default. For context, we don't support setting the default `ColorSpace` for an `ImageLoader` either.

coil-base/src/main/java/coil/request/ImageRequest.kt
```diff
@@ -114,9 +114,12 @@ class ImageRequest private constructor(
     /** @see Builder.allowHardware */
     val allowHardware: Boolean,
 
-    /** @see Builder.allowHardware */
+    /** @see Builder.allowRgb565 */
```
colinrtwhite, line 117: Nice catch!

coil-base/src/main/java/coil/decode/Options.kt
```diff
@@ -50,13 +54,14 @@ class Options(
...
         allowRgb565: Boolean = this.allowRgb565,
+        premultipliedAlpha: Boolean = this.premultipliedAlpha,
         headers: Headers = this.headers,
         parameters: Parameters = this.parameters,
         memoryCachePolicy: CachePolicy = this.memoryCachePolicy,
         diskCachePolicy: CachePolicy = this.diskCachePolicy,
         networkCachePolicy: CachePolicy = this.networkCachePolicy
-    ) = Options(context, config, colorSpace, scale, allowInexactSize, allowRgb565, headers, parameters,
-        memoryCachePolicy, diskCachePolicy, networkCachePolicy)
+    ) = Options(context, config, colorSpace, scale, allowInexactSize, allowRgb565, premultipliedAlpha,
```
colinrtwhite, line 63: We'll need to add `@Deprecated(HIDDEN)` versions of the constructor + copy method for binary compatibility.

colinrtwhite:
Looks good - thanks for adding this! Left a few comments.

coil-base/src/main/java/coil/request/DefaultRequestOptions.kt
romainguy: 👍

coil-base/src/main/java/coil/decode/Options.kt
romainguy: Will do!

colinrtwhite (approved):
Great, thanks!

PR #572, opened 2020-10-31
Mutate crossfade start/end drawables

### LINKED ISSUE #571: Using crossfade together with placeholder changes placeholder drawable alpha
**Describe the bug**
Using placeholder drawable together with crossfade results in modifying alpha of all instances of that drawable.
So alpha of that drawable will be in the same state as current drawable used in CrossfadeTransition:

crossfade_bug_compressed

**Expected behavior**
Crossfade animation shouldn't change original drawable alpha

**To Reproduce**
I used sample project to illustrate the issue.
```kotlin
// coil/sample/ImageListAdapter.kt:44
            load(item.uri) {
                crossfade(2000)
                placeholder(R.drawable.ic_launcher_foreground)
                parameters(item.parameters)
            }
```
```xml
// layout/activity_main.xml:25
<ImageView
       android:id="@+id/test_image"
       app:srcCompat="@drawable/ic_launcher_foreground"
       app:layout_constraintTop_toBottomOf="@id/app_bar"
       app:layout_constraintStart_toStartOf="parent"
       app:layout_constraintEnd_toEndOf="parent"
       android:layout_width="wrap_content"
       android:layout_height="wrap_content"/>

    <androidx.recyclerview.widget.RecyclerView
        android:id="@+id/list"
        android:layout_width="0dp"
        android:layout_height="0dp"
        app:layout_constraintBottom_toBottomOf="parent"
        app:layout_constraintEnd_toEndOf="parent"
        app:layout_constraintStart_toStartOf="parent"
        app:layout_constraintTop_toBottomOf="@id/test_image"/>
```

**Version**
1.0.0

**Solution**
Since we don't work directly with ImageView canvas, then alpha for drawable still needs to be modified.
All drawables from same resource share [constant state](https://android-developers.googleblog.com/2009/05/drawable-mutations.html), which includes its alpha.

Mutating the `start: Drawable?` for `CrossfadeDrawable`  solves this problem. There shouldn't be much overhead, as they still share underlying bitmap.

So there are 2 ways to go about this:
1. Mutate start drawable when creating `CrossfadeDrawable` - then users dont have to know about this thing and it works as expected. I've have added test and solution in https://github.com/coil-kt/coil/pull/572
2. State in docs that placeholder drawable should be mutated to avoid this effect. But i don't think that such gotchas are good idea.

Raenar4k:
Fixes #571

### DIFF
--- a/coil-base/src/main/java/coil/drawable/CrossfadeDrawable.kt
+++ b/coil-base/src/main/java/coil/drawable/CrossfadeDrawable.kt
@@ -24,7 +24,7 @@ import kotlin.math.roundToInt
  *
  * NOTE: The animation can only be executed once as the [start] drawable is dereferenced at the end of the transition.
  *
- * @param start The [Drawable] to crossfade from. Will be mutated to prevent modifying alpha of original drawable
+ * @param start The [Drawable] to crossfade from.
  * @param end The [Drawable] to crossfade to.
  * @param scale The scaling algorithm for [start] and [end].
  * @param durationMillis The duration of the crossfade animation.
@@ -32,7 +32,7 @@ import kotlin.math.roundToInt
  */
 class CrossfadeDrawable @JvmOverloads constructor(
     start: Drawable?,
-    private val end: Drawable?,
+    end: Drawable?,
     val scale: Scale = Scale.FIT,
     val durationMillis: Int = DEFAULT_DURATION,
     val fadeStart: Boolean = true
@@ -43,7 +43,8 @@ class CrossfadeDrawable @JvmOverloads constructor(
     private val intrinsicWidth = computeIntrinsicDimension(start?.intrinsicWidth, end?.intrinsicWidth)
     private val intrinsicHeight = computeIntrinsicDimension(start?.intrinsicHeight, end?.intrinsicHeight)
 
-    private var start: Drawable? = null
+    private var start: Drawable?
+    private val end: Drawable?
     private var startTimeMillis = 0L
     private var maxAlpha = 255
     private var state = STATE_START
@@ -52,8 +53,9 @@ class CrossfadeDrawable @JvmOverloads constructor(
         require(durationMillis > 0) { "durationMillis must be > 0." }
 
         this.start = start?.mutate()
-        start?.callback = this
-        end?.callback = this
+        this.end = end?.mutate()
+        this.start?.callback = this
+        this.end?.callback = this
     }
 
     override fun draw(canvas: Canvas) {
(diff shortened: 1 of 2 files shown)


coil-base/src/main/java/coil/drawable/CrossfadeDrawable.kt
```diff
@@ -24,14 +24,14 @@ import kotlin.math.roundToInt
  *
  * NOTE: The animation can only be executed once as the [start] drawable is dereferenced at the end of the transition.
  *
- * @param start The [Drawable] to crossfade from.
+ * @param start The [Drawable] to crossfade from. Will be mutated to prevent modifying alpha of original drawable
```
colinrtwhite, line 27: I don't think we need to mention mutation in the docs.

coil-base/src/main/java/coil/drawable/CrossfadeDrawable.kt
```diff
@@ -24,14 +24,14 @@ import kotlin.math.roundToInt
...
  * @param end The [Drawable] to crossfade to.
  * @param scale The scaling algorithm for [start] and [end].
  * @param durationMillis The duration of the crossfade animation.
  * @param fadeStart If false, the start drawable will not fade out while the end drawable fades in.
  */
 class CrossfadeDrawable @JvmOverloads constructor(
-    private var start: Drawable?,
+    start: Drawable?,
     private val end: Drawable?,
```
colinrtwhite, line 35: We should probably mutate `end` as well since it could be loaded from resources.

coil-base/src/main/java/coil/drawable/CrossfadeDrawable.kt
```diff
@@ -43,13 +43,15 @@ class CrossfadeDrawable @JvmOverloads constructor(
...
     private var startTimeMillis = 0L
     private var maxAlpha = 255
     private var state = STATE_START
 
     init {
         require(durationMillis > 0) { "durationMillis must be > 0." }
 
+        this.start = start?.mutate()
         start?.callback = this
         end?.callback = this
```
colinrtwhite, line 56: Should this be `this.start` and `this.end`? `drawable.mutate` is supposed to return the same drawable, but there's nothing to prevent a bad drawable from returning a new drawable.

colinrtwhite:
Thanks for finding + fixing this!

coil-base/src/main/java/coil/drawable/CrossfadeDrawable.kt
Raenar4k: Yep, didn't think about that! Since we can provide something like error placeholder. Tested it in sample, same bug
https://github.com/coil-kt/coil/issues/571#issuecomment-722013475

Raenar4k: Fixed. Not sure if tests are good enough, they are testing drawable changes, but with a bit of a hack  - since percent of animation is tied to `SystemClock` and can't be influenced directly.

colinrtwhite (approved):
Awesome, thanks!

Issue #230, opened 2020-01-06
Image not displayed when loading into CircleImageView


cioccarellia:
**Describe the bug**
After updating coil to version 0.9.1, icons downloaded from network are not loading anymore.

**Expected behavior**
Icons should be loading normally, instead, just a transparent element is left (icons targets are not `ImageViews` but `CircularImageView`s instead ([here the library](https://github.com/hdodenhof/CircleImageView)), inside a `RecyclerView`

**To Reproduce**
Try loading URL images inside a `RecyclerView` using a `CircularImageView`

**Logs/Screenshots**
I'm not able to report any logs yet, but I will be following up with more details

**Version**
Android 10, OxygenOS 10.0.7

colinrtwhite:
Thanks for the report. Sounds related to this issue: https://github.com/coil-kt/coil/issues/223. Taking a look

mario:
After testing this, let me just clarify that the images do actually load if you scroll down and back up. However there are indeed issues with the crossfading.

```
2020-01-13 22:11:37.219 8534-8534/coil.sample W/System.err: java.lang.IllegalArgumentException: width and height must be > 0
2020-01-13 22:11:37.219 8534-8534/coil.sample W/System.err:     at android.graphics.Bitmap.createBitmap(Bitmap.java:1113)
2020-01-13 22:11:37.219 8534-8534/coil.sample W/System.err:     at android.graphics.Bitmap.createBitmap(Bitmap.java:1080)
2020-01-13 22:11:37.220 8534-8534/coil.sample W/System.err:     at android.graphics.Bitmap.createBitmap(Bitmap.java:1030)
2020-01-13 22:11:37.220 8534-8534/coil.sample W/System.err:     at android.graphics.Bitmap.createBitmap(Bitmap.java:991)
2020-01-13 22:11:37.220 8534-8534/coil.sample W/System.err:     at de.hdodenhof.circleimageview.CircleImageView.getBitmapFromDrawable(CircleImageView.java:313)
2020-01-13 22:11:37.220 8534-8534/coil.sample W/System.err:     at de.hdodenhof.circleimageview.CircleImageView.initializeBitmap(CircleImageView.java:330)
2020-01-13 22:11:37.220 8534-8534/coil.sample W/System.err:     at de.hdodenhof.circleimageview.CircleImageView.setImageDrawable(CircleImageView.java:263)
[...]

Issue #223, opened 2020-01-02
PhotoView + Coil + crossfade = no pinch-to-zoom


JeremiahStephenson:
**Describe the bug**
When using Chris Banes' PhotoView library with Coil + crossfade the pinch-to-zoom is broken. This appears to be due to CrossfadeDrawable updating the loaded drawable's bounds. Unless I'm mistaken I only see two options to get around this. Either disable crossfade for PhotoView or create my own transition.

**Expected behavior**
Pinch-to-zoom continues to function with crossfade enabled and without creating custom transition.

**To Reproduce**
Load an image into a PhotoView using Coil with crossfade enabled.

**Version**
This is with version 0.9.1 of Coil

colinrtwhite:
Thanks for the report. @Jawnnypoo any ideas (since you're maintaining `PhotoView`)?

The behaviour of `CrossfadeDrawable` changed slightly in `0.9.0` to better support crossfading between drawables with no intrinsic dimensions (like a `ColorDrawable`) and drawables with dimensions.

`CrossfadeDrawable` will have a width/height of -1 if either of its children have that dimension as -1. [Code](https://github.com/coil-kt/coil/blob/master/coil-base/src/main/java/coil/drawable/CrossfadeDrawable.kt#L283).

The child drawable's bounds will be centered in the middle of the `CrossfadeDrawable`'s bounds. They're scaled at draw-time if it's not a `BitmapDrawable` (since `BitmapDrawable`s do their own scaling). [Code](https://github.com/coil-kt/coil/blob/master/coil-base/src/main/java/coil/drawable/CrossfadeDrawable.kt#L244).

Jawnnypoo:
Hey @JeremiahStephenson , I added Coil to the PhotoView sample on `master`, and I don't see any issues currently. Can you potentially make a PR on that repo that shows the issue, or post code changes in here needed to make this break? https://github.com/chrisbanes/PhotoView/blob/master/sample/src/main/java/com/github/chrisbanes/photoview/sample/CoilSampleActivity.kt

JeremiahStephenson:
@Jawnnypoo The sample works because the `multiplier` value in `CrossfadeDrawable.updateBounds` is always 1. In my scenario that's coming out as 1.2 and that leads to the issue. I'll try and duplicate it in your sample so you can see what I mean.

JeremiahStephenson:
[...]

PR #585, opened 2020-11-15
Support preferring exact intrinsic size in CrossfadeDrawable.

colinrtwhite:
Fixes: #223 
Fixes: #230 

By default, `CrossfadeDrawable`'s intrinsic size is -1 if **either** of its child drawables have a -1 intrinsic size. This PR adds an option so its intrinsic size will only be -1 if **both** of its child drawables have an intrinsic size of -1.

This allows a user to use crossfade with `PhotoView` and `CircleImageView` since they both require an exact (>0) intrinsic size.


Jawnnypoo (approved):
Nice!

coil-base/api/coil-base.api
```diff
@@ -231,7 +231,8 @@ public final class coil/drawable/CrossfadeDrawable : android/graphics/drawable/D
 	public fun <init> (Landroid/graphics/drawable/Drawable;Landroid/graphics/drawable/Drawable;Lcoil/size/Scale;)V
 	public fun <init> (Landroid/graphics/drawable/Drawable;Landroid/graphics/drawable/Drawable;Lcoil/size/Scale;I)V
 	public fun <init> (Landroid/graphics/drawable/Drawable;Landroid/graphics/drawable/Drawable;Lcoil/size/Scale;IZ)V
-	public synthetic fun <init> (Landroid/graphics/drawable/Drawable;Landroid/graphics/drawable/Drawable;Lcoil/size/Scale;IZILkotlin/jvm/internal/DefaultConstructorMarker;)V
```
colinrtwhite, line 234: Argh, this is binary incompatible.

PR #587, opened 2020-11-16
Change CENTER and MATRIX scale types to resolve to OriginalSize.

### LINKED ISSUE #545: SIGABRT crash with scaleType and precision set (0.13.0)
**Describe the bug**
Hello, I do have the following crash when loading this image which is 2783 * 87 with Coil v0.13.0 : 
image_causing_crash 

This is my code :
```kotlin
class NavigationActivity : AppCompatActivity() {

    override fun onCreate(savedInstanceState: Bundle?) {
        super.onCreate(savedInstanceState)

        val content = FrameLayout(this).apply {
            val imageView = ImageView(context).apply {
                layoutParams = FrameLayout.LayoutParams(
                        ViewGroup.LayoutParams.MATCH_PARENT,
                        246.dp
                ).apply { gravity = Gravity.CENTER_VERTICAL }

                scaleType = ImageView.ScaleType.CENTER

                load("https://user-images.githubusercontent.com/59887260/95083839-51a2c580-071d-11eb-99f5-5054cd395cf8.jpg") {
                    precision(Precision.EXACT)
                }
            }

            addView(imageView)
        }
        setContentView(content)
    }

}

private val Int.dp
    get() = (this * Resources.getSystem().displayMetrics.density).toInt()
```

Logcat :

```
2020-10-05 15:15:41.037 22871-22983/com.lanr.com.dev E/OpenGLRenderer: GL error:  GL_INVALID_VALUE
2020-10-05 15:15:41.038 22871-22983/com.lanr.com.dev A/OpenGLRenderer: glEGLImageTargetTexture2DOES error! GL_INVALID_VALUE (0x501)
2020-10-05 15:15:41.038 22871-22983/com.lanr.com.dev A/libc: Fatal signal 6 (SIGABRT), code -6 (SI_TKILL) in tid 22983 (RenderThread), pid 22871 (om.lanr.com.dev)
```

I reproduce this issue on a Samsung S8, with Android 9. I tried to reproduce on other physical devices (Android 5, 6 and Android 10) and on emulator (Android 9 and Android 10) and I never crash, but the image didn't load neither. But on my Samsung S8, it's always crashing so I'm able to reproduce this crash every time.

If I remove `precision` from the `load` function or `scaleType` from the ImageView, then it works like a charm. 

And a last thing, if my ImageView doesn't have a fixed height (i.e. if I use `ViewGroup.LayoutParams.WRAP_CONTENT` instead of `246.dp`), then it doesn't crash, but it doesn't load the image neither and I have this in my logcat :

```
2020-10-05 15:35:44.622 27125-27202/com.lanr.com.dev W/GraphicBufferAllocator: over alloc limit : width(69575) x height(2175) = 151325625  Bitmap::allocateSkiaHardwareBitmap pid [27125]
2020-10-05 15:35:44.622 27125-27202/com.lanr.com.dev W/OpenGLRenderer: createGraphicBuffer() failed in GraphicBuffer.create()
2020-10-05 15:35:44.622 27125-27215/com.lanr.com.dev D/skia: --- Failed to allocate a hardware bitmap
```

Thanks a lot

colinrtwhite:
Fixes #545 

This changes the default behaviour for the `CENTER` and `MATRIX` scale types to [better match](https://thoughtbot.com/blog/android-imageview-scaletype-a-visual-guide) `ImageView.setImageDrawable`. Before, the image would be scaled to fit in the `ImageView`'s bounds. Now we default to `OriginalSize` for those scale types.

NOTE: This only changes the default behaviour. Explicitly setting the size is not affected by this change.

### DIFF
--- a/coil-base/src/main/java/coil/request/ImageRequest.kt
+++ b/coil-base/src/main/java/coil/request/ImageRequest.kt
@@ -863,7 +863,7 @@ class ImageRequest private constructor(
 
         private fun resolveSizeResolver(): SizeResolver {
             val target = target
-            return if (target is ImageViewTarget && target.view.scaleType.let { it == MATRIX || it == CENTER }) {
+            return if (target is ImageViewTarget && target.view.scaleType.let { it == CENTER || it == MATRIX }) {
                 SizeResolver(OriginalSize)
             } else if (target is ViewTarget<*>) {
                 ViewSizeResolver(target.view)
--- a/coil-base/src/main/java/coil/request/ImageRequest.kt
+++ b/coil-base/src/main/java/coil/request/ImageRequest.kt
@@ -10,6 +10,8 @@ import android.graphics.drawable.Drawable
 import android.net.Uri
 import android.os.Build.VERSION.SDK_INT
 import android.widget.ImageView
+import android.widget.ImageView.ScaleType.CENTER
+import android.widget.ImageView.ScaleType.MATRIX
 import androidx.annotation.DrawableRes
 import androidx.annotation.MainThread
 import androidx.annotation.Px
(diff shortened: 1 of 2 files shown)


Jawnnypoo (approved):
Should we add this to the CHANGELOG as a "breaking change"?

colinrtwhite:
@Jawnnypoo Yep, will make sure to call it out in the changelog. It's not technically breaking, but is a moderately important behaviour change.

PR #600, opened 2020-11-28
Add PDF support in coil-pdf.

zhanghai:
I got PDF preview working in [zhanghai/MaterialFiles](https://github.com/zhanghai/MaterialFiles) so I was thinking of upstreaming it here.

Somehow I couldn't successfully complete the Gradle sync for Coil locally, so test wasn't written yet and the API text was manually edited. Just wanted to know if you would be interested in merging this so I'm uploading it now regardless.


colinrtwhite:
Sorry for the late response on this! I really appreciate the work you put into this, but I think I want to keep first party extensions to image formats/animated image formats only. I think this is best fit as an external library that we can link to similar to OkHttp's ["Works with OkHttp"](https://square.github.io/okhttp/works_with_okhttp/) section in their docs.

colinrtwhite:
Sorry for the late response on this! I really appreciate the work you put into this, but I think we should limit first party extensions to image formats/animated image formats only. I think this is best fit as an external library that we can link to similar to OkHttp's ["Works with OkHttp"](https://square.github.io/okhttp/works_with_okhttp/) section in their docs.

zhanghai:
I see, thanks for the reply! I'll just keep it as a closed PR here, since I don't want to maintain a standalone library just for two files either.

PR #621, opened 2020-12-18
Fix HttpFetcher blocking forever if called from the main thread.

### LINKED ISSUE #615: Calling execute() on the main thread freezes the app
**Describe the bug**

Attempts to run execute() on the main thread cause freezes the app.

**Expected behavior**

I expect to be able to run execute() on the main thread (even though yes, I know this is not good practice).

**To Reproduce**

Try to call `execute()` inside of `runBlocking {}` while on the main thread.

Here's a small sample:

```
class MainActivity : AppCompatActivity() {

  override fun onCreate(savedInstanceState: Bundle?) {
    super.onCreate(savedInstanceState)

    val imageLoader = ImageLoader.Builder(this)
      .build()

    Log.i("test", "start")

    runBlocking {
      imageLoader.execute(
        ImageRequest.Builder(this@MainActivity)
          .data(R.drawable.image)
          .build()
      )
    }

    Log.i("test", "end") // Never prints
  }
}
```

**Version**

Coil 1.1.0.

colinrtwhite:
Fixes #615.

Fixes `HttpFetcher` blocking forever if called from the main thread.

- Requests that don't explicitly disable reading from the network using `ImageRequest.networkCachePolicy(CachePolicy.DISABLED)` will fail due to a `NetworkOnMainThreadException`.
- Requests that disable reading from the network and only read from the disk cache will work as expected.

### DIFF
--- a/coil-base/src/test/java/coil/fetch/HttpFetcherTest.kt
+++ b/coil-base/src/test/java/coil/fetch/HttpFetcherTest.kt
@@ -1,15 +1,19 @@
 package coil.fetch
 
 import android.content.Context
+import android.os.NetworkOnMainThreadException
 import android.webkit.MimeTypeMap
 import androidx.core.net.toUri
 import androidx.test.core.app.ApplicationProvider
 import coil.bitmap.BitmapPool
+import coil.decode.DataSource
 import coil.decode.Options
 import coil.network.HttpException
+import coil.request.CachePolicy
 import coil.size.PixelSize
 import coil.util.createMockWebServer
 import coil.util.createTestMainDispatcher
+import coil.util.runBlockingTest
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.runBlocking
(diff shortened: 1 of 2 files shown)


coil-base/src/test/java/coil/fetch/HttpFetcherTest.kt
```diff
@@ -51,7 +55,7 @@ class HttpFetcherTest {
         context = ApplicationProvider.getApplicationContext()
         mainDispatcher = createTestMainDispatcher()
         server = createMockWebServer(context, "normal.jpg")
-        cache = Cache(File("build/cache"), 1024).apply { evictAll() }
+        cache = Cache(File("build/cache"), Long.MAX_VALUE).apply { evictAll() }
```
Jawnnypoo, line 58: Is this change okay?

Jawnnypoo (approved):

coil-base/src/test/java/coil/fetch/HttpFetcherTest.kt
colinrtwhite: Yep, the cache is evicted at the end of the test.

PR #622, opened 2020-12-21
Optimize gradle config

Goooler:
- add `extra.gradle.kts` under buildSrc which contains repositories and rootProject's extra properties for buildSrc's and rootProject's `build.gradle.kts` to use with `apply(from = "extra.gradle.kts")`
- based on last step, add extension function `setupBase()` in `Extensions.kt`, to replace `android {}` block in all library modules


build.gradle.kts
```diff
@@ -11,19 +11,21 @@ import org.jlleitschuh.gradle.ktlint.KtlintExtension
 import java.net.URL
 
 buildscript {
+    apply(from = "buildSrc/extra.gradle.kts")
```
colinrtwhite, line 14: I don't think we get much from extracting the `androidGradlePlugin` and `kotlinPlugin` into a separate file. Also `extra.gradle.kts` will cause us to set the `repositories` block multiple times. Can you remove this part?

buildSrc/src/main/kotlin/Extensions.kt
```diff
@@ -54,49 +57,73 @@ private inline fun <T> List<T>.sumByIndexed(selector: (Int, T) -> Int): Int {
     return sum
 }
 
-private fun DependencyHandler.testImplementation(dependencyNotation: Any): Dependency? {
-    return add("testImplementation", dependencyNotation)
-}
+private fun DependencyHandler.testImplementation(vararg names: Any): Array<Dependency?> =
```
colinrtwhite, line 60: This is personal preference, but I'd rather leave this part as is (even if the `testImplementation` and `androidTestImplementation` statements are a bit repetitive).

coil-base/build.gradle.kts
```diff
@@ -14,20 +12,7 @@ plugins {
...
-        testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
-    }
-    libraryVariants.all {
-        generateBuildConfigProvider?.configure { enabled = false }
-    }
-    testOptions {
-        unitTests.isIncludeAndroidResources = true
-    }
-}
+setupBase()
```
colinrtwhite, line 15: This looks good, but could we rename `setupBase` -> `setupLibraryModule` to be more explicit?

colinrtwhite:
Thanks for the contribution! Left a few comments.

build.gradle.kts
Goooler: Removed.
But for using `Gradle` and `Kotlin` api in `buildSrc`, should add theirs dependencies and repos in buildSrc's `build.gradle.kts`, therefore repos have been declared many times. This is the reason why I add a `extra.gradle.kts` and import with `apply(from = ...)`

buildSrc/src/main/kotlin/Extensions.kt
Goooler: Reverted.

coil-base/build.gradle.kts
Goooler: Renamed.

build.gradle.kts
colinrtwhite: @Goooler Ah sorry I missed that. I'll factor out the versions into gradle variables in a separate PR.

colinrtwhite:
Awesome, thanks for the clean up!

colinrtwhite (approved):

PR #626, opened 2020-12-29
Add squared constructor to Size

mati1:
Add squared constructor to `PixelSize` class.


colinrtwhite:
Thanks for the contribution, but I'd rather not add this to the API. `PixelSize` should be kept as simple as possible and size convenience methods should be added to `ImageRequest`.

PR #653, opened 2021-02-02
Add support for ImageLoader.newBuilder().

### LINKED ISSUE #614: Allow multiple ImageLoaders to share the same memory caches

If you want to use multiple ImageLoaders in an app, you run into memory cache problems. Namely, each one has its own cache, which uses a percentage of the app's memory allocation. Without sharing, enough ImageLoaders will end up filling up all available memory.

I want ImageLoaders to be able to share caches, so you can have one single cache (with a reasonable memory limit).

Either you could provide a cache to the builder, or there could be a way to build upon an existing ImageLoader such that it reuses the cache.

Note that I do not need (or even want) to customize the cache. I *only* want to be able to share it between ImageLoaders.

**Additional context**

Some context on *why* one might want multiple ImageLoaders.

In our app, image requests require authentication headers (for security). The authentication headers are setup as interceptors in OkHttp.

Our app allows for multiple users, each with their own unique auth tokens. Thus, each user gets their own `OkHttpClient`; they have the same base, but their interceptors vary.

With DI, it's a lot easier to create a single ImageLoader for each user, built on top of the user's `OkHttpClient`. It might be possible to work around this problem by using `Call.Factory` to provide the different `OkHttpClient` as necessary, but that would be much trickier than just sharing a single cache.

colinrtwhite:
Resolves #614.

Supports building a new `ImageLoader` instance using an existing instance as a base. This shares resources + configuration between the two instances.

~TODO: Needs more tests.~

### DIFF
--- a/coil-base/src/main/java/coil/ImageLoader.kt
+++ b/coil-base/src/main/java/coil/ImageLoader.kt
@@ -19,6 +19,7 @@ import coil.intercept.Interceptor
 import coil.map.Mapper
 import coil.memory.EmptyWeakMemoryCache
 import coil.memory.MemoryCache
+import coil.memory.RealMemoryCache
 import coil.memory.RealWeakMemoryCache
 import coil.memory.StrongMemoryCache
 import coil.request.CachePolicy
@@ -34,6 +35,7 @@ import coil.target.ViewTarget
 import coil.transition.CrossfadeTransition
 import coil.transition.Transition
 import coil.util.CoilUtils
+import coil.util.ImageLoaderOptions
 import coil.util.Logger
 import coil.util.Utils
 import coil.util.getDrawableCompat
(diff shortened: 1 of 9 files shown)


coil-base/src/main/java/coil/ImageLoader.kt
```diff
@@ -52,9 +54,6 @@ import java.io.File
  *
  * Image loaders are designed to be shareable and work best when you create a single instance and
  * share it throughout your app.
- *
- * It's recommended, though not required, to call [shutdown] when you've finished using an image loader.
```
colinrtwhite, line 56: Removed this and clarified that you probably don't need to shut down image loaders in the `shutdown` method doc.

Jawnnypoo (approved):
Nice improvement!

Issue #649, opened 2021-02-01
Coil doesn't load URLs with query parameters


alirezanazari:
**Describe the bug**
I try to load images which protected by signatures and token to show them by coil , it seems  query parameters removed from URL and this library doesn't load them. We used headers and setParameter() methods but doesn't work. 

Sample URL:
http://test.com/images/svg/xxxx.svg?SignedHeaders=host&Signature=bd1628825bb8e67a56edb9b

**Version**
Android 9, 10
Coil 1.1.1

imherrera:
Try logging out the url with okHttp interceptor

alirezanazari:
@imherrera actually downloaded and URL is correct but Coil doesn't load the image : 
Note : I checked and Glide loaded this URL.

```

 <-- 200 OK http://1.1.1.1/images/svg/image.svg?mz-Signature=68e87d4a08414aa0307abc1a38b1c84bae63bb9564e13 (109ms)
 Accept-Ranges: bytes
 Content-Length: 2806
 Content-Security-Policy: block-all-
 Content-Type: application/octet-stream
 ETag: "c6d8cb78952c7682b32ff0c"
 Last-Modified: Sat, 30 Jan 2021 16:24:25 GMT
 Server: MinIO/RELEASE.2020-50Z
 Vary: Origin
 X-Amz-Request-Id: 165FA684AEE1
 X-Xss-Protection: 1; mode=block
 Date: Mon, 01 Feb 2021 14:38:00 GMT
 <?xml version="1.0" encoding="UTF-8"?>
 <svg width="44px" height="44px" viewBox="0 0 44 44" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
     <title>ACB3B96C-FB03-4F91-95A6-022E76CD7108</title>
     <defs>
         <linearGradient x1="0%" y1="50%" x2="100%" y2="50%" id="linearGradient-1">
             <stop stop-color="#9E2373" offset="0%"></stop>
             <stop stop-color="#E8CAE1" offset="100%"></stop>
         </linearGradient>
     </defs>
     <g id="Payment" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
         <g id="Sim-List" transform="translate(-315.000000, -574.000000)" fill-rule="nonzero">
             <g id="Sim-Operator" transform="translate(0.000000, 306.000000)">
[...]

PR #654, opened 2021-02-03
Support automatic SVG detection using the file data.

colinrtwhite:
Adds parsing logic to `SvgDecoder` to automatically detect SVGs based on the contents of the `Source`.

This should greatly improve SVG autodetection and cover 99% of cases since the previous solution only checked the mime type. It's not possible to cover all cases without reading the whole file + a bunch of extra processing.

Should help avoid issues like: https://github.com/coil-kt/coil/issues/649

### DIFF
--- a/coil-svg/README.md
+++ b/coil-svg/README.md
@@ -16,4 +16,4 @@ val imageLoader = ImageLoader.Builder(context)
     .build()
 ```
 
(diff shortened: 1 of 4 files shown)


Jawnnypoo (approved):
LGTM!

PR #662, opened 2021-02-07
Enable explicit API mode, add explicit visibility and typing to comply

zsmb13:
This PR enables [explicit API mode](https://kotlinlang.org/docs/reference/whatsnew14.html#explicit-api-mode-for-library-authors), which requires explicit visibility modifiers for all declarations, and explicit types for public declarations. I only made existing visibility explicit, and did not change the visibilty of any declarations (though this is a good time to review that there's nothing that's accidentally public).

Since this (unfortunately) [also affects test sources](https://youtrack.jetbrains.com/issue/KT-37652), everything in tests has been marked `internal`.


zsmb13:
Didn't update tests yet, gonna fix that

colinrtwhite:
@ zsmb13 Thanks for this, however I'm not the biggest fan of adding `public` to all our public APIs (and `internal` to all the tests). We already validate binary compatibility changes on CI using the Kotlin's [`binary-compatibility-validator`](https://github.com/Kotlin/binary-compatibility-validator); is there an advantage to using the explicit API mode as well?

zsmb13:
Explicit API mode basically has the following advantages:

- Classes can no longer be accidentally added to public API (with implicit public visibility)
- You can no longer expose the wrong type somewhere via type inference, as anything public requires explicit typing

If you use the binary compat validator very seriously (have reviewed the entire generated initial `.api` file, and review every change), it will solve these for most cases. However, there are things that are public API that it doesn't list, e.g. reified functions, as they focus on binary compatibility, and don't care about addressing all source-incompatible changes.

The two things that explicit API mode enforces are also part of the [official Kotlin coding conventions for libraries](https://kotlinlang.org/docs/coding-conventions.html#coding-conventions-for-libraries), if you care about that sort of thing.

---

All the `internal` stuff required for tests really is annoying, but it should eventually go away when the bug gets fixed.

Gryzor:
> @zsmb13 Thanks for this, however I'm not the biggest fan of adding `public` to all our public APIs (and `internal` to all the tests). We already validate binary compatibility changes on CI using the Kotlin's [`binary-compatibility-validator`](https://github.com/Kotlin/binary-compatibility-validator); is there an advantage to using the explicit API mode as well?

But isn't adding `public` (or whatever visibility) the whole point of *explicit*? :)

colinrtwhite:
Appreciate the work on this 🙏 but going to close this out. We might want to revisit this in the future, but personally I still think using `binary-compatibility-validator` is enough (I might be in the minority). If we do come back to this we'll need to open a new PR as there have been a lot of API changes due to the 2.x work.

PR #666, opened 2021-02-10
Add gradle-versions-plugin for update checking

Goooler:
- Add a plugin named [gradle-versions-plugin](https://github.com/ben-manes/gradle-versions-plugin) to check dependency updates.
- Rules by `isStableVersion` to limit check stable versions only

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,6 +1,8 @@
 import coil.by
 import coil.groupId
+import coil.isStableVersion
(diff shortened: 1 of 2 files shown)


colinrtwhite:
Thanks, however I don't think we want to add new plugins to the repository unless needed.

colinrtwhite:
Thanks, however I don't think we want to add new plugins to the repository unless necessary.

PR #669, opened 2021-02-13
Bump ktlint plugin to 10.0.0

Goooler:


build.gradle.kts
```diff
@@ -46,7 +46,6 @@ allprojects {
     version = project.versionName
 
     extensions.configure<KtlintExtension>("ktlint") {
-        version by "0.40.0"
```
colinrtwhite, line 49: Please leave the `ktlint` version here - we'll probably want to bump the `ktlint` version independently of the plugin in the future.

colinrtwhite (approved):

PR #670, opened 2021-02-14
Enable the JVM IR backend.

colinrtwhite:
https://blog.jetbrains.com/kotlin/2021/02/the-jvm-backend-is-in-beta-let-s-make-it-stable-together/

Blocked by: https://youtrack.jetbrains.com/issue/KT-44993 (or a similar `inline` bug). **This bug is fixed in Kotlin `1.5.0`.**

### DIFF
--- a/coil-base/api/coil-base.api
+++ b/coil-base/api/coil-base.api
@@ -347,6 +347,22 @@ public final class coil/memory/MemoryCache$Key$Companion {
 	public final fun create (Ljava/lang/String;)Lcoil/memory/MemoryCache$Key;
 }
 
+public final class coil/memory/MemoryCache$Key$Complex$Creator : android/os/Parcelable$Creator {
+	public fun <init> ()V
+	public final fun createFromParcel (Landroid/os/Parcel;)Lcoil/memory/MemoryCache$Key$Complex;
+	public synthetic fun createFromParcel (Landroid/os/Parcel;)Ljava/lang/Object;
+	public final fun newArray (I)[Lcoil/memory/MemoryCache$Key$Complex;
+	public synthetic fun newArray (I)[Ljava/lang/Object;
+}
+
+public final class coil/memory/MemoryCache$Key$Simple$Creator : android/os/Parcelable$Creator {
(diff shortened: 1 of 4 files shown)


coil-base/api/coil-base.api
```diff
@@ -660,7 +676,7 @@ public final class coil/size/OriginalSize : coil/size/Size {
 	public fun writeToParcel (Landroid/os/Parcel;I)V
 }
 
-public class coil/size/OriginalSize$Creator : android/os/Parcelable$Creator {
+public final class coil/size/OriginalSize$Creator : android/os/Parcelable$Creator {
```
colinrtwhite, line 679: Interestingly this is not a binary incompatible change, however classes that extend this class with throw a `VerifyError` when instantiated. Assuming the Kotlin team has taken this into account.

Goooler: See [here](https://github.com/coil-kt/coil/pull/671)

coil-video/api/coil-video.api
```diff
@@ -4,6 +4,7 @@ public abstract class coil/fetch/VideoFrameFetcher : coil/fetch/Fetcher {
 	public static final field VIDEO_FRAME_OPTION_KEY Ljava/lang/String;
 	public fun <init> (Landroid/content/Context;)V
 	public fun fetch (Lcoil/bitmap/BitmapPool;Ljava/lang/Object;Lcoil/size/Size;Lcoil/decode/Options;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
+	public static final synthetic fun fetch$suspendImpl (Lcoil/fetch/VideoFrameFetcher;Lcoil/bitmap/BitmapPool;Ljava/lang/Object;Lcoil/size/Size;Lcoil/decode/Options;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
```
colinrtwhite, line 7: It's a bug in `1.4.31` that this is declared public, but it can't be called by Java and it's a Kotlin implementation detail so I wouldn't consider it part of the public API.

PR #676, opened 2021-02-21
Callback support for animated drawables

### LINKED ISSUE #606: Provide callback when animated GIF is done playing
Provide a mechanism for the caller to be notified when an animated GIF has completed all loops. ( Similar to Glide V4's onAnimationEnd listener - see [https://github.com/bumptech/glide/pull/3438](https://github.com/bumptech/glide/pull/3438) )

For example,
```
binding.myImageView.load(R.drawable.some_gif_animation) {
   repeatCount(0)
   onAnimationComplete = { /* do something */ }
}
```

sagar-viradiya:
- Fixes #606 
- Add start and end callback support for animated drawables.


coil-gif/src/main/java/coil/decode/GifDecoder.kt
```diff
@@ -54,6 +58,21 @@ class GifDecoder : Decoder {
 
         drawable.setRepeatCount(options.parameters.repeatCount() ?: MovieDrawable.REPEAT_INFINITE)
 
+        // Set the start and end animation callbacks if any one is supplied through the request.
+        if (options.parameters.animationStartCallback() != null || options.parameters.animationEndCallback() != null) {
+            drawable.registerAnimationCallback(object : Animatable2Compat.AnimationCallback() {
+                override fun onAnimationStart(drawable: Drawable?) {
+                    super.onAnimationStart(drawable)
```
colinrtwhite, line 65: The `super` implementations for `Animatable2` and `Animatable2Compat` are all no-ops so we don't need to call them.

coil-gif/src/main/java/coil/decode/ImageDecoderDecoder.kt
```diff
@@ -108,6 +112,22 @@ class ImageDecoderDecoder : Decoder {
             val drawable = if (baseDrawable is AnimatedImageDrawable) {
                 baseDrawable.repeatCount = options.parameters.repeatCount() ?: AnimatedImageDrawable.REPEAT_INFINITE
 
+                // Set the start and end animation callbacks if any one is supplied through the request.
+                if (options.parameters.animationStartCallback() != null ||
```
colinrtwhite, line 116: Nit: I would set `options.parameters.animationStartCallback()` to a local variable here (and in `GifDecoder`) so Kotlin infers that it's not-null inside the callback.

coil-gif/src/main/java/coil/request/Gifs.kt
```diff
@@ -35,3 +37,25 @@ fun ImageRequest.Builder.animatedTransformation(animatedTransformation: Animated
...
     return value(ANIMATED_TRANSFORMATION_KEY) as AnimatedTransformation?
 }
+
+/** Set the callback to be invoked at the start of the animation if the result is an animated [Drawable]. */
+fun ImageRequest.Builder.onAnimationStart(callback: (() -> Unit)?): ImageRequest.Builder {
+    return setParameter(ANIMATION_START_CALLBACK_KEY, callback)
+}
+
+/** Get the callback to be invoked at the start of the animation if the result is an animated [Drawable]. */
+@Suppress("UNCHECKED_CAST")
```
colinrtwhite, line 47: Nit: I would ignore this for the whole file to avoid cluttering the method signature.

coil-gif/src/main/java/coil/decode/ImageDecoderDecoder.kt
sagar-viradiya: Auto infer won't work here as there is `||` condition.

coil-gif/src/main/java/coil/decode/GifDecoder.kt
sagar-viradiya: Noted

coil-gif/src/main/java/coil/request/Gifs.kt
sagar-viradiya: Noted

coil-gif/src/androidTest/java/coil/callback/AnimationCallbacksTest.kt
```diff
@@ -0,0 +1,79 @@
...
+            }
+            .onAnimationEnd {
+                isEndCalled = true
+            }
+            .target(imageView)
+            .decoder(decoder)
+            .data("${ContentResolver.SCHEME_FILE}:///android_asset/animated.gif")
+            .build()
+        imageLoader.enqueue(imageRequest)
+        delay(5000)
```
colinrtwhite, line 75: `delay` could make the test flaky if the emulator is slow. I'd recommend converting `isStartCalled` and `isEndCalled` to `MutableStateFlow(false)`. Then you can start the image request and await them being set to true like [here](https://github.com/coil-kt/coil/blob/master/coil-base/src/androidTest/java/coil/request/DisposableTest.kt#L232).

coil-gif/src/main/java/coil/decode/GifDecoder.kt
```diff
@@ -66,5 +83,7 @@ class GifDecoder : Decoder {
     companion object {
         const val REPEAT_COUNT_KEY = "coil#repeat_count"
         const val ANIMATED_TRANSFORMATION_KEY = "coil#animated_transformation"
+        const val ANIMATION_START_CALLBACK_KEY = "coil#aimation_start_callback"
+        const val ANIMATION_END_CALLBACK_KEY = "coil#aimation_end_callback"
```
colinrtwhite, line 87: Typo here: it should be `animation`

coil-gif/src/androidTest/java/coil/callback/AnimationCallbacksTest.kt
sagar-viradiya: I have made the suggested changes. I have one concern here though. What if in the future due to some changes callback won't get triggered, Wouldn't it lead to suspension forever?

colinrtwhite: Yep! We could probably add a listener for `onSuccess/onError/onCancel`, but it awaiting forever is ok too. After ~25 minutes the tests will timeout. By migrating to use the state flow we guarantee that the test will only fail/stall if it's been broken.

colinrtwhite (approved):
Great work, thanks 🙏

rundavidrun:
Thanks again for this @sagar-viradiya. Works like a charm!

PR #685, opened 2021-03-03
Add params in gradle.properties

Goooler:
- Add params to enhance the speed of build

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,13 +1,25 @@
-org.gradle.jvmargs=-Xmx4g
+org.gradle.caching=true
+org.gradle.configureondemand=true
+org.gradle.jvmargs=-Xmx4g -Dfile.encoding=UTF-8
+org.gradle.parallel=true
 
 # AndroidX
 android.useAndroidX=true
+
+# Enable install debug apk manually
+android.injected.testOnly=false
 
 # Disable welcome message.
 systemProp.org.gradle.internal.launcher.welcomeMessageEnabled=false
 
 # Disable adding the standard library by default.
 kotlin.stdlib.default.dependency=false
+
+# Enhance speed of kotlin build
+kotlin.code.style=official
+kotlin.incremental=true
+kotlin.incremental.usePreciseJavaTracking=true
+kotlin.parallel.tasks.in.project=true
 
 # Config
 minSdk=14
--- a/gradle.properties
+++ b/gradle.properties
@@ -19,7 +19,6 @@ kotlin.stdlib.default.dependency=false
 kotlin.code.style=official
 
 # Enhance speed of kotlin build
-kotlin.incremental=true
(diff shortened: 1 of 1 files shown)


gradle.properties
```diff
@@ -1,14 +1,26 @@
-org.gradle.jvmargs=-Xmx4g
+org.gradle.caching=true
```
colinrtwhite, line 1: I think most of these are enabled by default.

gradle.properties
```diff
@@ -1,14 +1,26 @@
...
+org.gradle.configureondemand=true
+org.gradle.jvmargs=-Xmx4g -Dfile.encoding=UTF-8
+org.gradle.parallel=true
 
 # AndroidX
 android.useAndroidX=true
 
+# Enable install debug apk manually
+android.injected.testOnly=false
```
colinrtwhite, line 10: What does this do?

gradle.properties
```diff
@@ -1,14 +1,26 @@
...
+
 # Disable welcome message.
 systemProp.org.gradle.internal.launcher.welcomeMessageEnabled=false
 
 # Disable adding the standard library by default.
 kotlin.stdlib.default.dependency=false
 
+# Enhance speed of kotlin build
+kotlin.code.style=official
```
colinrtwhite, line 19: How does enabling the official kotlin code style improve build speed?

gradle.properties
```diff
@@ -1,14 +1,26 @@
...
+kotlin.incremental=true
```
colinrtwhite, line 20: These are all enabled by default.

gradle.properties
Goooler: I was negligent, it should be separated

gradle.properties
Goooler: image
Default value is false

gradle.properties
Goooler: Could let debuggale apk installed manually

gradle.properties
Goooler: Not sure this properity is enabled by default, bucause not found default value in gradle source

gradle.properties
colinrtwhite: Hmm according to [this blog](https://blog.jetbrains.com/kotlin/2017/03/kotlin-1-1-1-is-out/) it was enabled by default in Kotlin `1.1.1`.

gradle.properties
colinrtwhite: How does that increase build speed?

Goooler: No increasement, just convient for test

gradle.properties
Goooler: Ok, removed

colinrtwhite:
Thanks, however I'd prefer to stick to the defaults to prioritize stability over build speed.

PR #688, opened 2021-03-04
Use the view bounds as the intrinsic size of an SVG in SvgDecoder.

colinrtwhite:
Currently we use an SVG's `width`/`height` to determine its intrinsic size and its aspect ratio. However, we should actually be using its `viewBox` since that's the actual size of the rendered content. e.g. we could have an SVG that's 500x500, but a view box that's only the center 100x200 pixels. In that case the aspect ratio of the rendered content is 1x2 and not 1x1.

Also adds a `useViewBoundsAsIntrinsicSize` constructor param to opt out of the changes if it causes issues.

### DIFF
--- a/coil-svg/src/androidTest/assets/coil_logo.svg
+++ b/coil-svg/src/androidTest/assets/coil_logo.svg
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
-<svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" x="0px" y="0px" width="200px"
-    height="200px" viewBox="332.5 200 140 140" xml:space="preserve">
+<svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" x="0px" y="0px" width="1000px"
+    height="250px" viewBox="332.5 200 140 140" xml:space="preserve">
 <g>
 	<g>
 		<path fill="#45DFF4" d="M392.728,303.382l-32.384-24.241l-3.243,2.431c-5.523,4.131-5.523,10.837,0,14.97l35.627,26.672
--- a/coil-svg/src/main/java/coil/decode/SvgDecoder.kt
+++ b/coil-svg/src/main/java/coil/decode/SvgDecoder.kt
@@ -3,17 +3,16 @@
 package coil.decode
 
 import android.content.Context
-import android.graphics.Bitmap
 import android.graphics.Canvas
 import android.graphics.RectF
 import android.graphics.drawable.Drawable
-import android.os.Build.VERSION.SDK_INT
 import androidx.core.graphics.drawable.toDrawable
 import coil.bitmap.BitmapPool
 import coil.size.OriginalSize
 import coil.size.PixelSize
 import coil.size.Size
 import coil.util.indexOf
+import coil.util.toSoftware
 import com.caverock.androidsvg.SVG
 import okio.BufferedSource
 import okio.ByteString.Companion.encodeUtf8
@@ -23,12 +22,12 @@ import okio.buffer
  * A [Decoder] that uses [AndroidSVG](https://bigbadaboom.github.io/androidsvg/) to decode SVG files.
  *
  * @param context A [Context] used to create the [Drawable].
- * @param useViewBoundsAsSize If true, uses the SVG's view bounds as the intrinsic size for the SVG.
+ * @param useViewBoundsAsIntrinsicSize If true, uses the SVG's view bounds as the intrinsic size for the SVG.
  *  If false, uses the SVG's width/height as the intrinsic size for the SVG.
  */
 class SvgDecoder @JvmOverloads constructor(
     private val context: Context,
-    private val useViewBoundsAsSize: Boolean = true
+    private val useViewBoundsAsIntrinsicSize: Boolean = true
 ) : Decoder {
 
     override fun handles(source: BufferedSource, mimeType: String?): Boolean {
@@ -50,19 +49,19 @@ class SvgDecoder @JvmOverloads constructor(
 
         val svgWidth: Float
         val svgHeight: Float
-        val viewBox: RectF? = svg.documentViewBox
-        if (useViewBoundsAsSize && viewBox != null) {
-            svgWidth = viewBox.width()
-            svgHeight = viewBox.height()
-        } else {
-            svgWidth = svg.documentWidth
-            svgHeight = svg.documentHeight
-        }
-
         val bitmapWidth: Int
         val bitmapHeight: Int
+        val viewBox: RectF? = svg.documentViewBox
         when (size) {
             is PixelSize -> {
+                if (useViewBoundsAsIntrinsicSize && viewBox != null) {
+                    svgWidth = viewBox.width()
+                    svgHeight = viewBox.height()
+                } else {
+                    svgWidth = svg.documentWidth
+                    svgHeight = svg.documentHeight
+                }
+
                 if (svgWidth > 0 && svgHeight > 0) {
                     val multiplier = DecodeUtils.computeSizeMultiplier(
                         srcWidth = svgWidth,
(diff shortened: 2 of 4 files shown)


coil-svg/src/androidTest/assets/coil_logo.svg
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
-<svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" x="0px" y="0px" width="200px"
-    height="200px" viewBox="332.5 200 140 140" xml:space="preserve">
+<svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" x="0px" y="0px" width="1000px"
```
colinrtwhite, line 3: Changing this so the test cases fail if it uses the `width`/`height` for the aspect ratio.

coil-svg/src/main/java/coil/decode/SvgDecoder.kt
```diff
@@ -3,24 +3,32 @@
...
  * A [Decoder] that uses [AndroidSVG](https://bigbadaboom.github.io/androidsvg/) to decode SVG files.
+ *
+ * @param context A [Context] used to create the [Drawable].
+ * @param useViewBoundsAsIntrinsicSize If true, uses the SVG's view bounds as the intrinsic size for the SVG.
+ *  If false, uses the SVG's width/height as the intrinsic size for the SVG.
  */
-class SvgDecoder(private val context: Context) : Decoder {
+class SvgDecoder @JvmOverloads constructor(
+    private val context: Context,
+    private val useViewBoundsAsIntrinsicSize: Boolean = true
```
Jawnnypoo, line 30: Is it okay that this changes behavior for those who are using this currently? Maybe it should default to false until a 1.2 or a 2.0 so that folks don't update their library and end up with unexpected behavior? Otherwise, I'd just make sure its communicated clearly in the CHANGELOG.

Jawnnypoo (approved):

coil-svg/src/main/java/coil/decode/SvgDecoder.kt
colinrtwhite: @Jawnnypoo Good point. Will call this out at the top of the release notes.

I'm definitely for only enabling this in `1.2` (which will be the next release). `2.0` is still a ways away so I'd want to introduce the behaviour change before then. Also I'm hoping this change will be invisible for most users as for most SVGs `width`/`height` already has the same aspect ratio as `viewBox`. This change mostly improves the handling for weird SVGs that Coil wouldn't render properly.

Jawnnypoo: Gotcha! Thanks for clarifying 👍

zhanghai:
Could you clarify why should the intrinsic size of an SVG be its view box size (if present)? You mentioned this way it would "correctly follow the SVG specification", but I couldn't find relevant specification in [W3C SVG Specification](https://www.w3.org/TR/SVG11/coords.html#ViewBoxAttribute).

Meanwhile, I tried to create an SVG with width `24`, height `24` and a view box of `0 0 6 24`, so that the size and view box have different aspect ratio. Both Nautilus (GNOME Files) file properties and EOG (GNOME Image Viewer) reported a image size of `24x24`, and both EOG and Inkscape displayed the SVG in `1:1` aspect ratio instead of the `1:4` from view box.

```
<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 6 24" version="1.1">
    <path fill="#000000" d="M 10,5 V 19 H -4 V 5 H 10 M 10,3 H -4 c -1.1,0 -2,0.9 -2,2 v 14 c 0,1.1 0.9,2 2,2 h 14 c 1.1,0 2,-0.9 2,-2 V 5 C 12,3.9 11.1,3 10,3 Z m -4.86,8.86 -3,3.87 L 0,13.14 -3,17 H 9 Z" />
</svg>
```

colinrtwhite:
@zhanghai Sorry, the title of this PR isn't accurate. This PR changes the logic so the view box is used to compute the aspect ratio to fit into the target size. For example, if we have an SVG that is 100x200, but the view box is only 100x100, the image actually only has a 1x1 aspect ratio (everything outside of the view box isn't rendered).

If the target size is `OriginalSize`, the `width`/`height` elements of the SVG are used as expected.

zhanghai:
I see, thanks for the clarification about using view box for fitting the target size. However, it still seems to me the original behavior is correct.

> For example, if we have an SVG that is 100x200, but the view box is only 100x100, the image actually only has a 1x1 aspect ratio (everything outside of the view box isn't rendered).

This doesn't sound correct according to my understanding of SVG.

1. Although the view box may be smaller, it's still perfectly valid for content to be drawn outside the view box and be shown in the final SVG, because:

    1.  the `viewBox` attribute is essentially just a shorthand for a transformation matrix:

        >  The effect of the ‘viewBox’ attribute is that the user agent automatically supplies the appropriate transformation matrix to map the specified rectangle in user space to the bounds of a designated region (often, the viewport). 
        >
        > https://www.w3.org/TR/SVG11/coords.html#ViewBoxAttribute

    2. the default value for `preserveAspectRatio` (when `viewBox` is present) is `xMidYMid` instead of `none`, so that uniform scaling is performed and the view box is centered, leaving additional space for drawing.

        > - xMidYMid (the default) - Force uniform scaling.
        >     Align the midpoint X value of the element's ‘viewBox’ with the midpoint X value of the viewport.
        >     Align the midpoint Y value of the element's ‘viewBox’ with the midpoint Y value of the viewport.
        >
        > https://www.w3.org/TR/SVG11/coords.html#PreserveAspectRatioAttribute

        (A potential source of confusion is that for Android VectorDrawable XML files, there is no `preserveAspectRatio` equivalent and `android:viewBox` behaves as if `preserveAspectRatio` is `none`. But Android VectorDrawable XML isn't SVG.)

     3. The SVG file in my previous comment is an example of this (It's the Material Design icon for "image", and obviously draws outside its view box), and I've confirmed in EOG and Inkscape that they treat the SVG file as if the view box is just a transformation matrix, and nothing else (size, aspect ratio, zoom) is affected by the view box.

        ```svg
        <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 6 24" version="1.1">
            <path fill="#000000" d="M 10,5 V 19 H -4 V 5 H 10 M 10,3 H -4 c -1.1,0 -2,0.9 -2,2 v 14 c 0,1.1 0.9,2 2,2 h 14 c 1.1,0 2,-0.9 2,-2 V 5 C 12,3.9 11.1,3 10,3 Z m -4.86,8.86 -3,3.87 L 0,13.14 -3,17 H 9 Z" />
        </svg>
        ```

2. Even if no content is drawn outside of the view box, it is still perfectly valid for an SVG file to utilize the view box for transformation, and it should be handled as if the transformation were applied at a root group instead of as the view box.

    Take this SVG file for an ellipse as an example:

    ```svg
        <svg xmlns="http://www.w3.org/2000/svg" width="48" height="24" viewBox="0 0 24 24" preserveAspectRatio="none" version="1.1">
            <circle cx="12" cy="12" fill="#000000" r="12" />
        </svg>
    ```

    It should be handled exactly like the following equivalent SVG file:

    ```svg
        <svg xmlns="http://www.w3.org/2000/svg" width="48" height="24" version="1.1">
            <g transform="scale(2, 1)">
                <circle cx="12" cy="12" fill="#000000" r="12" />
            </g>
        </svg>
    ```

    This is actually also covered simiarly in the specification's discussion of how the user agent may implement `viewBox` with `transform`:

    > The effect is equivalent to having a viewport of size 300px by 200px and the following supplemental transformation in the document, as follows:
    >
    > ...
    >
    > https://www.w3.org/TR/SVG11/coords.html#ViewBoxAttribute

3. Even if some non-standard handling of `viewBox` for fitting the target size is desired for some reason, it should actually respect the value of `preserveAspectRatio` to be a correct implementation. The current implementation would return an incorrect result if `preserveAspectRatio` is set to `none`, in which case the aspect ratio from the `width` and `height` attributes should definitely be used instead of the one from `viewBox`. I still don't see why we need non-standard handling of `viewBox` for fitting the target size though.

PR #689, opened 2021-03-04
Add VideoFrameDecoder to support decoding video frames from any source.

### LINKED ISSUE #687: Support loading video frames from the network.
Currently, Coil only supports decoding video frames from local sources (files or local URIs). 

Coil should support decoding video frames from a URL.

**Additional context**
This is something that I'd really like to add! Unfortunately, the API Coil uses to decode video frames, [`MediaMetadataRetriever`](https://developer.android.com/reference/android/media/MediaMetadataRetriever), has limitations 
around using a network-backed `Source` and works much better with a `File`. A Coil `Decoder` intentionally has no knowledge of `File`s to allow the source of an image to be completely decoupled from decoding logic.

colinrtwhite:
Fixes #687.

This finally adds support for decoding video frame from any source (including the network!) with an important caveat. `VideoFrameDecoder` creates a short lived file on disk to buffer the source into. This is less performant than decoding the file in memory and it requires enough space on disk to copy the video which could be a problem for very large video files.

`VideoFrameFileFetcher` and `VideoFrameUriFetcher` don't have this caveat which is why they're not deprecated. Long term (possibly in the next major version) we'll hopefully be able to remove this limitation on `VideoFrameDecoder` and deprecate the fetchers.

### DIFF
--- a/coil-video/src/main/java/coil/decode/VideoFrameDecoder.kt
+++ b/coil-video/src/main/java/coil/decode/VideoFrameDecoder.kt
@@ -31,15 +31,18 @@ class VideoFrameDecoder(private val context: Context) : Decoder {
         options: Options
     ): DecodeResult {
         val tempFile = File.createTempFile("tmp", null, context.cacheDir.apply { mkdirs() })
-        source.use { tempFile.sink().use(it::readAll) }
-
-        val retriever = MediaMetadataRetriever()
         try {
-            retriever.setDataSource(tempFile.path)
-            return delegate.decode(pool, retriever, size, options)
+            source.use { tempFile.sink().use(it::readAll) }
+
+            val retriever = MediaMetadataRetriever()
+            try {
+                retriever.setDataSource(tempFile.path)
+                return delegate.decode(pool, retriever, size, options)
+            } finally {
+                retriever.release()
+            }
         } finally {
             tempFile.delete()
-            retriever.release()
         }
     }
 
--- a/coil-video/src/main/java/coil/decode/VideoFrameDecoder.kt
+++ b/coil-video/src/main/java/coil/decode/VideoFrameDecoder.kt
@@ -1,0 +1,50 @@
+@file:Suppress("unused")
+
+package coil.decode
+
+import android.content.Context
+import android.media.MediaMetadataRetriever
+import coil.bitmap.BitmapPool
+import coil.size.Size
+import okio.BufferedSource
+import okio.sink
+import java.io.File
+
+/**
+ * A [Decoder] that uses [MediaMetadataRetriever] to fetch and decode a frame from a video.
+ *
+ * NOTE: [VideoFrameDecoder] creates a temporary copy of the video on the file system. This may cause the decode
+ * process to fail if the video being decoded is very large and/or the device is very low on disk space.
+ */
+class VideoFrameDecoder(private val context: Context) : Decoder {
+
+    private val delegate = VideoFrameDecoderDelegate(context)
+
+    override fun handles(source: BufferedSource, mimeType: String?): Boolean {
+        return mimeType != null && mimeType.startsWith("video/")
+    }
+
+    override suspend fun decode(
+        pool: BitmapPool,
+        source: BufferedSource,
+        size: Size,
+        options: Options
+    ): DecodeResult {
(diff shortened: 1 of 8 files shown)


coil-video/src/main/java/coil/decode/VideoFrameDecoderDelegate.kt
```diff
@@ -0,0 +1,157 @@
...
+import androidx.core.graphics.drawable.toDrawable
+import coil.bitmap.BitmapPool
+import coil.request.videoFrameMicros
+import coil.request.videoFrameOption
+import coil.size.OriginalSize
+import coil.size.PixelSize
+import coil.size.Size
+import kotlin.math.roundToInt
+
+internal class VideoFrameDecoderDelegate(private val context: Context) {
```
colinrtwhite, line 22: This class is basically pulled out entirely from `VideoFrameFetcher` without changes.

coil-video/src/main/java/coil/decode/VideoFrameDecoder.kt
```diff
@@ -0,0 +1,54 @@
...
+        return mimeType != null && mimeType.startsWith("video/")
+    }
+
+    override suspend fun decode(
+        pool: BitmapPool,
+        source: BufferedSource,
+        size: Size,
+        options: Options
+    ): DecodeResult {
+        val tempFile = File.createTempFile("tmp", null, context.cacheDir.apply { mkdirs() })
```
Jawnnypoo, line 33: We don't need write permissions to do this, right?

Jawnnypoo (approved):
👍 Great work, I know this is much appreciated from those who need it.

coil-video/src/main/java/coil/decode/VideoFrameDecoder.kt
colinrtwhite: Yep! The `cacheDir` is free to write to.

ultraon:
It doesn't look like a workable solution in conjunction with streaming content urls (mpd - MPEG DASH, m3u8 - HLS), right?

colinrtwhite:
@ultraon It should work with [any format supported by Android](https://developer.android.com/guide/topics/media/media-formats#video-codecs).

yehia2030:
it still didn't work with me in compose and this is my example

            val context = LocalContext.current
            Image(
                modifier = Modifier
                    .fillMaxWidth()
                    .height(200.dp)
                    .padding(horizontal = 10.dp),
                painter = rememberAsyncImagePainter(ImageRequest.Builder(LocalContext.current)
                    .data(data = "http://commondatastorage.googleapis.com/gtv-videos-bucket/sample/BigBuckBunny.mp4")
                    .apply(block = {
                     decoderFactory(VideoFrameDecoder.Factory())
                    }).build()),
                contentDescription = "test",
            )

if there is any solution please let me know

yehia2030:
after  a lot of pain and i got a black image 

            val imageLoader = ImageLoader.Builder(context).build()
            val model = ImageRequest.Builder(LocalContext.current)
                .data(data = "http://commondatastorage.googleapis.com/gtv-videos-bucket/sample/BigBuckBunny.mp4")
                .apply(block = {
                    decoderFactory(VideoFrameDecoder.Factory())
                    videoFrameMillis(50)
                    listener(
                        onStart = {println("Start")},
                        onError = { _, result -> println("Error ${result.throwable.message}")},
                        onSuccess = { _, _ ->  println("Success")},
                        onCancel = { println("Cancel") }
                    )
                    size(100)
                }).build()
            Image(
                modifier = Modifier
                    .fillMaxWidth()
                    .height(200.dp)
                    .padding(horizontal = 10.dp),
                painter = rememberAsyncImagePainter(model, imageLoader = imageLoader,
                    onError = {
                    println("Error 2 ${it.result.throwable.message}")
                }, onLoading = {
                        println("Loading 2 ")
                    }, onSuccess = {
                        println("Success 2${it.result.isSampled} ")

                    }),
                contentDescription = "test"
            )
            LaunchedEffect(key1 = true) {
                imageLoader.execute(model)
            }

yehia2030:
could we make a buffer of something like that as if the video is long it takes too much time to load the image  ? as i think it loaded the fully video then get its thumbnail

PR #700, opened 2021-03-16
Move kotlinOptions into `setupBase()`

Goooler:


build.gradle.kts
```diff
@@ -50,21 +49,6 @@ allprojects {
         disabledRules by setOf("indent", "max-line-length")
     }
 
-    tasks.withType<KotlinCompile>().configureEach {
```
colinrtwhite, line 53: Currently this is in `allProjects` so it also configures the overall project and its subprojects to use Java 8. Moving it to `setupBaseModule` makes it so it's only configured for Android subprojects.

buildSrc/src/main/kotlin/Extensions.kt
```diff
@@ -104,6 +108,11 @@ fun DependencyHandler.addAndroidTestDependencies(kotlinVersion: String, includeT
     androidTestImplementation(Library.OKHTTP_MOCK_WEB_SERVER)
 }
 
+inline fun BaseExtension.kotlinOptions(block: KotlinJvmOptions.() -> Unit) {
+    (this as ExtensionAware).extensions.getByName<KotlinJvmOptions>("kotlinOptions").block()
```
colinrtwhite, line 112: Not sure if it also applies to extensions, but using `getByName` instead of `configure` forces eager evaluation instead of lazy.

build.gradle.kts
Goooler: `KotlinOptions` now should be used in subprojects only, root project use the default Java (currently it was `Java11` embedded in Android Studio)

colinrtwhite: @Goooler Do you have a source on that? I thought the JVM that runs the code is different from the language level target set by the `KotlinCompile` task.

Goooler: Not yet. 
You can change level to Java11 and compile for a test, it will be failed.
And I think arguments in `kotlinOptions` that you added are useless for root project.

buildSrc/src/main/kotlin/Extensions.kt
Goooler: Now with your [commit](https://github.com/coil-kt/coil/commit/18a3c12db6d3be86f7081b66bc0188443ce2b491), I think it's ok to write like this.

Goooler: Make `kotlinOptions` lazy now like what defined in gradle source
image

Goooler:
And I noticed test task config in root project, can we also move it into `setupBaseModule`?

image

PR #715, opened 2021-04-01
Remove redundant dependency

Goooler:


colinrtwhite:
@Goooler We [disable adding the standard library by default](https://github.com/coil-kt/coil/blob/master/gradle.properties#L10) so we need to add the dependency explicitly.

Goooler:
> @Goooler We [disable adding the standard library by default](https://github.com/coil-kt/coil/blob/master/gradle.properties#L10) so we need to add the dependency explicitly.

But you api the std library in base module，dependendcy should be implemented in sample.

Goooler:
I made a new project for test, just `implementation` the `coil`, it works fine.
image

Goooler:
And I think maybe we should add `<uses-permission android:name="android.permission.INTERNET" />` in coil base's manifest file.

colinrtwhite:
@Goooler Good point, we could technically remove this as `coil-base` declares it as an `api` dependency as the standard library is exposed as part of Coil's API. However, it's better to keep the explicit `implementation` dependency here as the `coil-sample` project depends on the standard library irrespective of its dependency on Coil.

Re adding the internet permission: we shouldn't do this automatically as there are users who only use Coil for local files and don't want the internet permission.

PR #718, opened 2021-04-06
Modify repositories & setupAppModule()

Goooler:
- Integrate repositories into `extra.gradle.kts`
- `mavenCentral` could be removed in buildScript now, as dokka have already been avaliable in [gradlePortal](https://plugins.gradle.org/plugin/org.jetbrains.dokka-android)
- Decouple `setupAppModule()`

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -10,7 +10,6 @@ buildscript {
     apply(from = "buildSrc/extra.gradle.kts")
     repositories {
         google()
-        mavenCentral()
         gradlePluginPortal()
     }
     dependencies {
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -30,21 +30,10 @@ extensions.configure<ApiValidationExtension> {
 }
 
 allprojects {
-    repositories {
-        google()
-        mavenCentral()
-
-        // https://github.com/Kotlin/dokka/issues/41
-        jcenter {
-            content {
-                includeModule("org.jetbrains.dokka", "dokka-fatjar")
-            }
-        }
-    }
-
     group = project.groupId
     version = project.versionName
 
+    apply(from = "${rootDir.path}/buildSrc/extra.gradle.kts")
     apply(plugin = "org.jlleitschuh.gradle.ktlint")
 
     extensions.configure<KtlintExtension>("ktlint") {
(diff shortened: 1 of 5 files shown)


Goooler:
image

Goooler:
It seems CI should be re-runed.

buildSrc/src/main/kotlin/Projects.kt
```diff
@@ -29,7 +29,17 @@ fun Project.setupAppModule(block: BaseAppModuleExtension.() -> Unit = {}) {
             versionName = project.versionName
             resConfigs("en")
             vectorDrawables.useSupportLibrary = true
+            multiDexEnabled = true
         }
+        buildTypes {
+            named("release") {
```
colinrtwhite, line 35: I think this is `coil-sample`-specific configuration that we don't want to pull out into `setupAppModule` - same with `multiDexEnabled`.

build.gradle.kts
```diff
@@ -30,21 +29,10 @@ extensions.configure<ApiValidationExtension> {
 }
 
 allprojects {
-    repositories {
-        google()
-        mavenCentral()
-
-        // https://github.com/Kotlin/dokka/issues/41
-        jcenter {
```
colinrtwhite, line 38: Dokka <1.0 has not been fully migrated to `mavenCentral` so this is needed.

buildSrc/extra.gradle.kts
```diff
@@ -1,3 +1,8 @@
+repositories {
```
colinrtwhite, line 1: I don't think we gain much by refactoring these into the `extras` script. I'd rather we specify `google` and `mavenCentral` explicitly for each project. It's one extra line as we don't need to apply `extra.gradle.kts`.

colinrtwhite:
Thanks for working on this, however I don't think these changes are necessary. If you're interested in contributing I'd check out the ["help wanted" issues](https://github.com/coil-kt/coil/issues?q=is%3Aissue+is%3Aopen+label%3A%22help+wanted%22). They're good jumping off points to add fixes/improvements to Coil.

build.gradle.kts
Goooler: Ok

PR #719, opened 2021-04-06
Add exported=true for MainActivity

Goooler:


Goooler:
Maybe we should reformat XMLs in whole projects

colinrtwhite:
Thanks, however I don't think this is needed for the sample app. If you're interested in contributing I'd check out the ["help wanted" issues](https://github.com/coil-kt/coil/issues?q=is%3Aissue+is%3Aopen+label%3A%22help+wanted%22). They're good jumping off points to add fixes/improvements to Coil.

PR #720, opened 2021-04-06
Replace build versionCode hardcode

Goooler:


colinrtwhite:
This is intentional; see [here](https://github.com/coil-kt/coil/issues/236) for the context.

Goooler:
> This is intentional; see [here](https://github.com/coil-kt/coil/issues/236) for the context.

Ok

PR #727, opened 2021-04-10
Remove sourceCompatibility & targetCompatibility declaration

Goooler:
After AGP 4.2, the default Java version in `compileOptions` is `JavaVersion.VERSION_1_8`, no need to declare `sourceCompatibility` & `targetCompatibility`.

image
image


colinrtwhite:
Thanks, however I'd prefer to specify this explicitly instead of relying on the defaults (which may change).

PR #729, opened 2021-04-13
Fix FakeImageLoader sample.

ychescale9:


docs/image_loaders.md
```diff
@@ -49,7 +49,7 @@ However, if you'd prefer a singleton the `io.coil-kt:coil` artifact provides a s
 For instance, you could inject a fake `ImageLoader` implementation which always returns the same `Drawable` synchronously:
 
 ```kotlin
-val fakeImageLoader = object : ImageLoader {
+fun createFakeImageLoader(context: Context) = object : ImageLoader {
```
colinrtwhite, line 52: Let's stick with `val` here to keep the callsite simple.

ychescale9: Changed to property extension.

colinrtwhite: I'd prefer to leave it as is since the code is meant to be used as an example to then call `Coil.setImageLoader` or inject the `fakeImageLoader` into your dependency graph.

colinrtwhite: Ah I see - for the context for `ImageLoader.Builder` we can just have `ImageLoader.Builder(context)` and assume anyone copying the code can provide a context (this example code will probably be copied into a `TestApplication` class). That way we can keep `val fakeImageLoader`.

ychescale9: Done 😄

colinrtwhite (approved):
Thanks!

Issue #747, opened 2021-05-06
Still used memory after the whole cache clear


patrykkubiakespeo:
I've loaded images with Coil which in my sample app uses the 70 MB in the graphics section in the Android Profiler. As you can see on the photo, after clearing the whole cache (memory, disk, and network) and clearing the bitmap pool there is still the 50 MB occupied memory. Also, the shutdown method didn't help to free the whole memory. 
![image](https://user-images.githubusercontent.com/74542228/117303364-e79dfa80-ae7c-11eb-969f-884a15877dba.png)

I am curious how can I free that memory. In comparison, Glide can free all the used memory (the memory usage is the same before and after the image loading) so I know that this should be possible.
![image](https://user-images.githubusercontent.com/74542228/117305291-e5d53680-ae7e-11eb-8170-a135ff487b0e.png)

colinrtwhite:
Hmm this looks related to the fact that we don't proactively call `Bitmap.recycle` if the [`WeakMemoryCache`](https://github.com/coil-kt/coil/blob/master/coil-base/src/main/java/coil/memory/WeakMemoryCache.kt#L137) is cleared. Those bitmaps will be cleared naturally by GC (assuming your code isn't holding a reference to those bitmaps), however we may be able Glide's behaviour here - will have to check.

colinrtwhite:
Going to close this out as expected behaviour and continue discussion in #749.

PR #749, opened 2021-05-07
Recycle bitmaps from WeakMemoryCache

kubiakdev:
This is the improvement pull request allowing us to recycle bitmaps from WeakMemoryCache and release still used memory after the current `WeakMemoryCache.clearMemory()` method. It is related to this issue: #747 

The solution was mentioned by @colinrtwhite . 

I tested the solution manually by changing the launching activity and then moving back from the MainActivity to the newly added activity together with clearing the memory cache and forcing the garbage collector. You can see on the screen that memory has returned to the level before the images loading.
image

### DIFF
--- a/coil-base/src/main/java/coil/memory/WeakMemoryCache.kt
+++ b/coil-base/src/main/java/coil/memory/WeakMemoryCache.kt
@@ -134,7 +134,16 @@ internal class RealWeakMemoryCache(private val logger: Logger?) : WeakMemoryCach
     override fun clearMemory() {
         logger?.log(TAG, Log.VERBOSE) { "clearMemory" }
         operationsSinceCleanUp = 0
+        recycleCachedBitmaps()
         cache.clear()
+    }
+
+    private fun recycleCachedBitmaps() {
+        cache.values.forEach { cacheValues ->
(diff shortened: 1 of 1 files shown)


kubiakdev:
After changes, the instrumented tests fail on trying to draw a recycled bitmap. The problem occurs only for enqueuing the vector image request. I didn't find the place where the Coil attempt to draw the recycled bitmap.

Also, I was wondering that should I add recycling bitmaps in the `WeakMemoryCache.cleanUp()` method.

colinrtwhite:
Thanks for the PR, however I don't think we can make this change (at least not by default). Coil doesn't own the bitmaps in the memory cache the same way Glide does; this makes the API much simpler for the user (compare Coil's base target with Glide's). The weak memory cache can hold references to bitmaps that are still in use by `ImageView`s in the user's app and clearing the bitmaps here could cause unexpected exceptions.

`MemoryCache.clear` instantly clears all references in the cache so they can be freed naturally by the GC; they will be garbage collected in the next garbage collection pass as long as nothing else is holding a reference to them (like one of the user's `ImageView`s). I'd be curious hear more about your use case - are you getting `OutOfMemoryException`s without this change?

PR #757, opened 2021-05-13
Modify org.gradle.jvmargs

Goooler:
[AGP 4.2 Optimizing Gradle builds for JDK 11](https://developer.android.com/studio/releases/gradle-plugin#kts)

image


colinrtwhite (approved):
Thanks!

PR #763, opened 2021-05-19
Fix fakeImageLoader in docs

chrisbanes:
Drawables are stateful, so it's not safe to re-use them across multiple requests which could be drawn at the same time. Fixed by returning new drawables for each request.

See: https://github.com/android/compose-samples/pull/538


colinrtwhite (approved):
Ah good catch, thanks!

PR #768, opened 2021-06-03
Update github actions version

Goooler:


.github/workflows/ci.yml
```diff
@@ -2,20 +2,21 @@ name: CI
 on:
     push:
         branches:
-            - main
+            - '*'
```
colinrtwhite, line 5: We don't want to run CI on non-PR branches since we have limited concurrent resources.

.github/workflows/ci.yml
```diff
@@ -2,20 +2,21 @@ name: CI
...
     pull_request:
 jobs:
     check:
         name: Check
         runs-on: ubuntu-latest
         timeout-minutes: 10
         steps:
-            - uses: actions/checkout@v2
+            - uses: actions/checkout@v2.3.4
```
colinrtwhite, line 13: I don't think we should pin to specific patch versions (same for the other non-major updates in this file). Instead we should keep `checkout@v2` so it automatically uses the latest non-breaking version.

.github/workflows/ci.yml
Goooler: Reverted.

.github/workflows/ci.yml
Goooler: Reverted.

colinrtwhite (approved):
Thanks!

PR #774, opened 2021-06-09
Fix Licensee not finding Coil's Licence url

leandrofavarin:
Hey folks,

It seems that there's a typo on the key for the licence url, according to to what [gradle-maven-publish-plugin](https://github.com/vanniktech/gradle-maven-publish-plugin/blob/fc8c5c2982d319c7debf3bc88ac2e1a6e64fcbad/src/main/kotlin/com/vanniktech/maven/publish/MavenPublishBaseExtension.kt#L183) expects.

This typo then causes [Licensee](https://github.com/cashapp/licensee) to not find Coil's Licence URL, returning null, blocking us from adding it to the allowlist.

Let me know if there's something else that needs changed, happy to contribute.


colinrtwhite (approved):
Good catch, thanks!

PR #775, opened 2021-06-10
Add new allowConversionToBitmap flag, fixes #773

### LINKED ISSUE #773: Automatically choose between animatedTransformation and transformations if both are set
I am trying to load images from user-provided URLs and I don't know ahead of time if the result will be a bitmap or an animated gif.  I want to circle-crop the result either way, but afaict this is not supported.

If there's an `animatedTransformation` on the request builder and the result is animated, don't apply `transformations`.  If the result is _not_ animated, apply `transitions` normally.

edenman:
This allows animated transformations and normal transformations to
coexist on the same request.  If this flag is set to false:

- A normal bitmap result will have normal transformations applied to it
- An animated result will be left intact which allows
  animatedTransformations to apply.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -12,4 +12,6 @@ If you would like to contribute code you can do so through GitHub by forking the
 
 When submitting code, please make every effort to follow existing conventions and style in order to keep the code as readable as possible. Please also make sure your code passes all tests by running `./test.sh`.
 
+If you are making an API change, run `./gradlew apiDump` and include any changed files in your pull request.
+
 *Modified from OkHttp's [Contributing](https://square.github.io/okhttp/contributing/) section.*
--- a/coil-base/src/main/java/coil/decode/Options.kt
+++ b/coil-base/src/main/java/coil/decode/Options.kt
@@ -166,7 +166,6 @@ class Options(
     ) = copy(context, config, colorSpace, scale, allowInexactSize, allowRgb565, premultipliedAlpha,
         allowConversionToBitmap, headers, parameters, memoryCachePolicy, diskCachePolicy, networkCachePolicy)
 
-
     @Deprecated(message = "Kept for binary compatibility.", level = DeprecationLevel.HIDDEN)
     fun copy(
         context: Context = this.context,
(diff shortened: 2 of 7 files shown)


edenman:
Wasn't sure how to re-generate `coil-base/api/coil-base.api`...tried assemble_docs.sh but that didn't do it

colinrtwhite:
Awesome, thanks for adding this! You'll need to run `./gradlew apiDump` to updated the API file. Also you'll need to add a new `@Deprecated(HIDDEN)`-annotated constructor and copy function for `Options` to maintain binary compatibility ([like this](https://github.com/coil-kt/coil/blob/main/coil-base/src/main/java/coil/decode/Options.kt#L111)).

coil-base/src/main/java/coil/decode/Options.kt
```diff
@@ -137,6 +144,24 @@ class Options(
...
         diskCachePolicy: CachePolicy = this.diskCachePolicy,
         networkCachePolicy: CachePolicy = this.networkCachePolicy
-    ) = copy(context, config, colorSpace, scale, allowInexactSize, allowRgb565, premultipliedAlpha, headers, parameters,
-        memoryCachePolicy, diskCachePolicy, networkCachePolicy)
+    ) = copy(context, config, colorSpace, scale, allowInexactSize, allowRgb565, premultipliedAlpha,
+        allowConversionToBitmap, headers, parameters, memoryCachePolicy, diskCachePolicy, networkCachePolicy)
+
+
+    @Deprecated(message = "Kept for binary compatibility.", level = DeprecationLevel.HIDDEN)
+    fun copy(
```
colinrtwhite, line 152: @edenman We need to add another  deprecated constructor [like this](https://github.com/coil-kt/coil/blob/main/coil-base/src/main/java/coil/decode/Options.kt#L111) as well.

CONTRIBUTING.md
```diff
@@ -12,4 +12,6 @@ If you would like to contribute code you can do so through GitHub by forking the
 
 When submitting code, please make every effort to follow existing conventions and style in order to keep the code as readable as possible. Please also make sure your code passes all tests by running `./test.sh`.
 
+If you are making an API change, run `./gradlew apiDump` and include any changed files in your pull request.
```
colinrtwhite, line 15: 🙏

colinrtwhite:
@edenman Looks like there are just a couple minor ktlint warnings left.

edenman:
Also had a test failure on api 19, fixed that and the ktlint stuff

colinrtwhite (approved):
@edenman Awesome, thanks for implementing this!

Issue #540, opened 2020-09-24
A gif is playing very fast


rupinderjeet:
![tumblr_ku2pvuJkJG1qz9qooo1_r1_400](https://user-images.githubusercontent.com/14011726/94132137-7d4fc100-fe7c-11ea-8512-69f90cb65e48.gif)

Plays really fast than how it is shown in Android Gallery and Chrome. Issue occurs while using `ImageDecoderDecoder`. If only `GifDecoder` is used, gif doesn't play at all.

How do I load this gif? Checkout #539 

This is how it looks when I load it:

[record.mp4.zip](https://github.com/coil-kt/coil/files/5275386/record.mp4.zip)

IljaKosynkin:
Hey @rupinderjeet, @colinrtwhite 
I poked around with this issue (have a little bit of free time on my hands and decided to see if there are OS projects that might need help).
From my research, it seems that this particular GIF has 0 delay time in-between frames @rupinderjeet. The reason it plays so differently on the web is that major browsers handle this case by replacing delay with a default value if it is lesser then cut-off. ImageDecoder plays it as received (really fast) and Movie just shows the first frame. 
When I replaced delay time with 15 it seems to be loading fine by Coil with both decoders.
![ezgif-6-53c89e37efaa](https://user-images.githubusercontent.com/6876788/96633009-d1818000-1318-11eb-9f1d-7f914f4ccb16.gif)
Now the question to the @colinrtwhite: I presume that pre-processing GIF or writing custom decoder, which allows replacing delay with minimal one, is way beyond the scope of that library, so what do we do with that issue?

colinrtwhite:
@IljaKosynkin Thanks for investigating this! Do you know how tough it would be to pre-process the GIF as it's being read? I'd be open to a solution that uses [`ForwardingSource`](https://square.github.io/okio/2.x/okio/okio/-forwarding-source/) to replace the GIF frame delay similar to web browsers depending the performance impact + how much code it would require.

IljaKosynkin:
@colinrtwhite I can't tell it from the top of my head regarding complexity and performance impact, to be honest, but I'm going to poke around to tell estimation, possibly with PR, if it is not too bad in my opinion.

IljaKosynkin:
Also, on a side note, from my investigation it seems that all files in `gifs.json` are actually WEBP files, maybe I should add some actual GIFs here?

colinrtwhite:
[...]

PR #783, opened 2021-06-21
Rewrite the GIF frame delay if it's below a threshold.

colinrtwhite:
Fixes: #540

Adds an `enforceMinimumFrameDelay` constructor arg to `GifDecoder` and `ImageDecoderDecoder` that enables rewriting the GIF's frame delay if it's below a threshold (we use the same values as major web browsers + Glide).

There's a very small chance that this fix introduces visual artifacts since it scans the source for a magic sequence instead of implementing a complete GIF parser so `enforceMinimumFrameDelay` is currently disabled by default.


asuka-mio:
No longer need that workaround on Android 14 above
https://android.googlesource.com/platform/frameworks/base/+/2be87bb707e2c6d75f668c4aff6697b85fbf5b15
@colinrtwhite

dexterpentlandmckinnon803-maker:
nice





good job gng

PR #789, opened 2021-07-01
Update dependencies

Goooler:


build.gradle.kts
```diff
@@ -11,7 +11,6 @@ buildscript {
     apply(from = "buildSrc/plugins.gradle.kts")
     repositories {
         google()
-        mavenCentral()
```
colinrtwhite, line 14: This is required for [Dokka](https://github.com/Kotlin/dokka#using-the-gradle-plugin).

Goooler: Dokka is avaliable in `gradlePluginPortal()`, I aslo execute `./gradlew clean dokkaHtmlMultiModule` for test

image

colinrtwhite: This is because `gradlePluginPortal` [implicitly mirrors Jcenter](https://github.com/gradle/gradle/issues/15406) which mirrors `mavenCentral`. This is an implementation detail that's likely to change soon which will make `gradlePluginPortal` stop resolving `mavenCentral` libraries.

Goooler: I get it. Reverted.

colinrtwhite (approved):

Issue #443, opened 2020-06-24
Add support for Jetpack Compose


Icyrockton:
Jetpack compose alpha is coming in summer.....
support jetpack compose, I think it will improve the competitiveness of coil
😎

colinrtwhite:
Totally agree! Chris Banes actually has a [library](https://github.com/chrisbanes/accompanist) that integrates Coil with Compose. Compose is still under development, but the eventual plan is to upstream Accompanist (just the Coil parts) into Coil.

andylamax:
I join the movement as well. Coil Compose is a necessity at the moment

luca992:
I made a [basic library](https://github.com/luca992/coil-composable) wrapping coil for jetpack compose. It even works with gifs, and animated drawables now: https://github.com/luca992/coil-composable

colinrtwhite:
Coil was mentioned [here](https://www.youtube.com/watch?v=PjQdFmiDgwk&t=1129s) and [here](https://www.youtube.com/watch?v=SMOhl9RK0BA&t=231) in the Jetpack Compose alpha videos!

liuhuiAndroid:
I find a library which use coil for jetpack compose, address is here: https://github.com/skydoves/Landscapist

liuhuiAndroid:
![image](https://user-images.githubusercontent.com/12246679/100335434-f51ea100-300f-11eb-8807-070f48085475.png)

colinrtwhite:
@chrisbanes and I are now working on upstreaming https://github.com/google/accompanist/tree/main/coil into a new `coil-compose` artifact in this repo. It'll be part of the next minor release!

PR #790, opened 2021-07-02
Add support for Jetpack Compose.

colinrtwhite:
Adds a `coil-compose` artifact heavily derived from [Accompanist](https://github.com/google/accompanist)'s Coil integration.

The API might change in this PR (looking for feedback!) but tentatively it'll look like this:

```kotlin
// Basic request
Image(
    painter = rememberImagePainter("https://www.example.com/image.jpg"),
    contentDescription = null
)

// Request with configuration
Image(
    painter = rememberImagePainter(
        data = "https://www.example.com/image.jpg",
        builder = {
            // A custom transition that only works with Compose.
            // Regular transitions (including crossfade) aren't compatible with Compose at the moment.
            fadeIn(true)
            transformation(CircleCropTransformation())
        }
    ),
    contentDescription = null
)
```

## Changes from Accompanist

This is built on Accompanist's integration but makes a number of API and behaviour changes summarized here (non-exhaustive):

- Creates two artifacts `coil-compose` (the default artifact) and `coil-compose-base` (doesn't include the singleton `ImageLoader`). `coil-compose` depends on `coil-compose-base` similar to how the `coil` artifact depends on `coil-base`.
- `rememberCoilPainter` -> `rememberImagePainter`
- `ShouldRefetchOnSizeChange` -> `ExecuteCallback`
    - `ExecuteCallback` has a superset of the functionality of `ShouldRefetchOnSizeChange`
- The `fadeIn` and `fadeInDurationMs` arguments for `rememberLoadPainter` have been moved to an `ImageRequest.Builder` extension function.
- `previewPlaceholder` is removed. Set `ImageRequest.Builder.placeholder` instead.
- `LocalImageLoader` is now not-null and defaults to the singleton `ImageLoader`.
- Notably, `rememberImagePainter` will no longer fall back to using the image's original size when used in a `LazyColumn` or other composable where the dimensions are not set.
  - This change was made to avoid using a timeout, which can be flaky.

## Open questions

- Is it possible to get the content scale for an image so it can be loaded at the correct size?
- What's the best way to handle `LazyColumn`?

cc @chrisbanes 

Resolves: https://github.com/coil-kt/coil/issues/443


coil-compose-base/src/main/java/coil/compose/Drawable.kt
```diff
@@ -0,0 +1,130 @@
...
+import androidx.compose.ui.graphics.painter.Painter
+import androidx.compose.ui.graphics.withSave
+import androidx.compose.ui.unit.LayoutDirection
+import kotlin.LazyThreadSafetyMode.NONE
+import kotlin.math.roundToInt
+
+/**
+ * Convert this [Drawable] into a [Painter] using Compose primitives if possible.
+ */
+internal fun Drawable.toPainter(): Painter {
```
chrisbanes, line 35: I'll likely keep this in Accompanist as a new small lib. I'll follow up once I've refactored.

chrisbanes (approved):
LGTM, I like the API changes.

I wonder if we could remove `fadeIn()` and instead hook-up `crossfade()` to do the same thing?

coil-compose-base/src/main/java/coil/compose/Drawable.kt
chrisbanes: https://github.com/google/accompanist/issues/540

colinrtwhite: Sounds good :+1:. It's definitely super useful, but I didn't want to make it public API in Coil.

colinrtwhite:
@chrisbanes I'd like to - having a transition that only works in Compose is a little strange. I'll do a quick exploration in a subsequent PR.

coil-compose-base/src/main/java/coil/compose/Drawable.kt
colinrtwhite: Sounds good, it's definitely super useful, but I didn't want to make it public API in Coil. We'll swap to using Accompanist's dependency once it's published.

PR #794, opened 2021-07-06
Add `setupCompose`

Goooler:


colinrtwhite:
Thanks, however I don't think we need to refactor this. The code is fairly small and only used in two places.

Goooler:
> Thanks, however I don't think we need to refactor this. The code is fairly small and only used in two places.

I think may use it in sample module

PR #798, opened 2021-07-09
Bump gradle-maven-publish-plugin to 0.17.0

Goooler:


buildSrc/src/main/kotlin/Library.kt
```diff
@@ -6,7 +6,7 @@ object Library {
 
     // CORE
 
-    private const val COROUTINES_VERSION = "1.5.0"
+    private const val COROUTINES_VERSION = "1.5.1"
```
colinrtwhite, line 9: We can't update to this until Jetpack Compose updates to Kotlin `1.5.20`.

Goooler: Reverted.

colinrtwhite (approved):

PR #802, opened 2021-07-11
Fix currently when image loading error it looks like the screen is em…

HuixingWong:
Fix currently when image loading error it looks like the item is disappeared， it's a little confusing


colinrtwhite (approved):
Thanks!

PR #818, opened 2021-07-22
[WIP] Merge 2.0 into main.

colinrtwhite:
This branch has the changes for the 2.0 release of Coil.


coil-base/src/androidTest/java/coil/EventListenerTest.kt
```diff
@@ -90,9 +87,9 @@ class EventListenerTest {
             imageLoader.testEnqueue {
                 data("$SCHEME_ANDROID_RESOURCE://${context.packageName}/${R.drawable.normal}")
                 transformations(object : Transformation {
-                    override fun key() = "test_transformation"
+                    override val cacheKey = "test_transformation"
```
Goooler, line 90: May should declare their types in override methods and properties ?

colinrtwhite: IMO it's probably ok to avoid specifying the type here since it's for a local anonymous class.

PR #827, opened 2021-07-27
Add .gitattributes

Goooler:
- Add `.gitattributes`, [document here](https://git-scm.com/docs/gitattributes).

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,4 @@
+* text eol=lf
+
+*.bat text eol=crlf
+*.jar binary
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,6 +1,8 @@
 root = true
 
 [*]
-indent_size=4
-insert_final_newline=true
-max_line_length=120
+charset = utf-8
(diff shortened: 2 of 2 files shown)


.gitattributes
```diff
@@ -0,0 +1,4 @@
+* text=auto eol=lf
+
+*.bat text eol=crlf
```
colinrtwhite, line 3: Thanks for the PR, but what do we gain by specifying these? Seems like extra config for limited gain.

Goooler: For example, if I execute `gradlew wrapper` in Windows, the Windows's default eol is crlf, will chnage the gradlew file, lead to it can't be executable.

Goooler:
[gradle's](https://github.com/gradle/gradle/blob/master/.gitattributes).

.gitattributes
colinrtwhite: Gotcha, sounds good 👍

colinrtwhite (approved):

PR #830, opened 2021-07-29
Add .editorconfig rules

Goooler:
Amended commit for #827.
Need add any other rules ?


colinrtwhite (approved):
Thanks, I don't think we need to add any more. Github will auto-detect these files are binaries. It's mostly the lf/crlf support that's useful.

PR #847, opened 2021-08-05
Use proper syntax

Goooler:


colinrtwhite:
Thanks, however I want to avoid merging PRs that could be considered [bikeshedding](https://en.wiktionary.org/wiki/bikeshedding). The syntax changes aren't definitively better or worse.

Issue #889, opened 2021-09-15
Disable crossfade transition if target view is not shown


sav007:
**Describe the bug**
If target view is not shown ([view.isShown](https://developer.android.com/reference/android/view/View#isShown())==false), executing image request with crossfade transition enabled, never completes if image fetched from disk / network. As a result passed `ImageRequest.Listener` is never called. If image is in the mem cache, everything works as expected because of `transition === Transition.NONE`.

Here is a possible use case, the main content (that includes image view) of the screen is invisible and placeholders are shown until the content and image are loaded. After content is loaded it shown to the user.

As workaround on client side, check if view is not shown then override image request with crossfade config disabled.

**Expected behavior**
My assumption if target view is not shown, then skip crossfade transition.

**Version**
1.3.2

colinrtwhite:
Thanks for the report - I agree it would be better to skip the transition in this case. `CrossfadeTransition` already skips suspending if [`View.isVisible`](https://github.com/coil-kt/coil/blob/main/coil-base/src/main/java/coil/transition/CrossfadeTransition.kt#L45) returns false, though I'm guessing a parent view is invisible in your case? We could replace `isVisible` with `isShown` - my only concern is if there is a potential performance impact.

sav007:
yes in our case the target image view is visible but the parent container view is not.

> with isShown - my only concern is if there is a potential performance impact.

You mean because it traverse the parent view hierarchy?

colinrtwhite:
@sav007 Yep, though I did some quick tests and the performance impact looks pretty negligible.

PR #898, opened 2021-09-20
Only execute CrossfadeTransition if view is shown.

colinrtwhite:
Fixes: #889

### DIFF
--- a/coil-base/src/main/java/coil/transition/CrossfadeTransition.kt
+++ b/coil-base/src/main/java/coil/transition/CrossfadeTransition.kt
@@ -4,7 +4,6 @@ package coil.transition
 
 import android.graphics.drawable.Drawable
 import android.widget.ImageView
-import androidx.core.view.isVisible
 import androidx.vectordrawable.graphics.drawable.Animatable2Compat
 import coil.annotation.ExperimentalCoilApi
 import coil.decode.DataSource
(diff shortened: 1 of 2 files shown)


coil-base/src/main/java/coil/transition/CrossfadeTransition.kt
```diff
@@ -42,7 +41,7 @@ class CrossfadeTransition(
 
         // Don't animate if the view is not visible as CrossfadeDrawable.onDraw
```
nuhkoca, line 42: You may want to update the comment line, as well.

```suggestion
        // Don't animate if the view is not shown as CrossfadeDrawable.onDraw
```

coil-base/src/main/java/coil/transition/CrossfadeTransition.kt
```diff
@@ -42,7 +41,7 @@ class CrossfadeTransition(
...
         // won't be called until the view becomes visible.
```
nuhkoca, line 43:
```suggestion
        // won't be called until the view is shown.
```

coil-base/src/main/java/coil/transition/CrossfadeTransition.kt
colinrtwhite: Ah good point.

Issue #817, opened 2021-07-21
Memory cache is always missed


gemiren:
I'm loading a large image into an small ImageView. ImageView size is 64 x 64.
```
<ImageView
        android:id="@+id/imageView_logo_small"
        android:layout_width="64dp"
        android:layout_height="64dp"
        android:scaleType="fitCenter"/>
```

The image is loaded fine. However, memory cache is always missed for that particular image when I scroll the recyclerview up and down to that image again.

Coil is complaining the cached image size doesn't match the target ImageView size.
```
Cached image's request size (175, 117) is smaller than the requested size (176, 176, FIT)
```
Based on the debug log it looks the 1px mismatch probably is a rounding error. Is possible to make Coil more tolerant for such a small mismatch?

Here is the image https://www.marketplace.org/wp-content/uploads/2021/07/GettyImages-1190028286.jpg?fit=6549%2C4366.

I also tried other the ImageView sizes. When I use the size 60 x 60, the size from memory cache matches target size.

colinrtwhite:
Coil actually already has [code to handle off by one sizes](https://github.com/coil-kt/coil/blob/main/coil-base/src/main/java/coil/intercept/EngineInterceptor.kt#L216). In your case it looks like you're making a previous image request with a different aspect ratio (175, 117) vs. (176, 176). Are you changing the `ImageView`'s size? Do you have a sample project that reproduces the issue that I could take a look at to help debug?

gemiren:
No, I didn't change the ImageView's size? It's always sets to 64dp. My guess is that the cached size is down sized to 175 due to rounding.

PR #899, opened 2021-09-20
Fix potential memory cache miss if scaling multiplier is slightly less than 1 due to rounding issue.

colinrtwhite:
Fixes: #817.


gemiren:
This pull request seems to have this 1px issue fixed. Thanks.

PR #937, opened 2021-10-10
Declare types

Goooler:
See square/okhttp/pull/6842.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -6,3 +6,7 @@ indent_size = 4
 insert_final_newline = true
 max_line_length = 100
 trim_trailing_whitespace = true
+
+[*.{kt, kts}]
+ij_kotlin_name_count_to_use_star_import = 999
+ij_kotlin_name_count_to_use_star_import_for_members = 999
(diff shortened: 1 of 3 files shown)


colinrtwhite:
I appreciate the work you put into this, but I don't think this PR should be merged (similar to [this PR](https://github.com/coil-kt/coil/pull/662)). If the goal is to ensure the public API is correct and doesn't change, we use the Kotlin binary compatibility plugin to track this. If the goal is to be more legible, then I think it makes sense to declare return types as needed. Declaring them for `Builder` classes, internal functions, and `Unit` return types isn't necessary.

Goooler:
Ok, let's make it more legible. I reverted some changes, should we reopen the PR?

colinrtwhite:
@Goooler Personally, I don't think it's a net improvement to declare the return types in most of these cases. Are there specific instances you think are improved by declaring the return type?

Goooler:
Yeah, I think it's more convenient for developers to read the coil's code and use APIs. Like this below one:
```
override suspend fun decode() = parallelismLock.withPermit {
    runInterruptible { BitmapFactory.Options().decode() }
}
```
If I don't go to the definition of `withPermit()` I can't understand the type of `decode()` immediately, instead it's easy to get it.

PR #939, opened 2021-10-10
Fix cancellation propagation

### LINKED ISSUE #933: Coil 2.0: Possible cancellation issue (at least with compose)
**Describe the bug**
When using coil 2.0 with compose in lazy column, during fast fling, the requests are no more cancelled and all previous requests needs to ends before loading the one currently on screen.

**Expected behavior**
Offscreen not yet loaded image requests should be cancelled.

**To Reproduce**
Fast scroll a lazycolumn and wait for image to appear (and see logs, no cancellation all success from network requests). Not yet sure 100% about repro code as I'm not yet 100% confident on my compose and coil 2.0 understanding.

So opening this early to get some pointers on where to look in existing coil tests for compose cancellation stuff to try build a repro test.

Tolriq:
Pass the proper parent coroutine context to the async job so that cancellation is properly passed down.

Fixes #933

Some tests are needed + Ensure if current propagation issue does not also touch enqueue()

### DIFF
--- a/coil-base/src/main/java/coil/RealImageLoader.kt
+++ b/coil-base/src/main/java/coil/RealImageLoader.kt
@@ -122,7 +122,8 @@ internal class RealImageLoader(
 
     override suspend fun execute(request: ImageRequest): ImageResult {
         // Start executing the request on the main thread.
-        val job = scope.async {
+        // We need to pass parent context for proper cancellation propagation.
+        val job = scope.async(coroutineContext) {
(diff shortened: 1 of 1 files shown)


Tolriq:
@colinrtwhite Thanks you're right about dispatcher order override sorry about that.

Your change have some side effects from before code as it no more use your declared scope in that class that have the logger on exception and the cancellation on shutdown. Not sure it really matters.

colinrtwhite:
Yep that's the same behaviour as in 1.x.

PR #957, opened 2021-10-18
Add ability to load music thumbnails on android 29.

### LINKED ISSUE #922: Coil is unable to load Audio album cover - "content://media/external/audio/albums/508"
**Describe the bug**
Coil is unable to load `content://media/external/audio/albums/508` urls. It fails with error `java.lang.IllegalArgumentException: Invalid column _data`

**Expected behavior**
Load the image using [imageDecoder](https://developer.android.com/reference/android/graphics/ImageDecoder#createSource(android.content.ContentResolver,%20android.net.Uri)) or [ContentResolver.loadThumbnail](https://developer.android.com/reference/android/content/ContentResolver#loadThumbnail(android.net.Uri,%20android.util.Size,%20android.os.CancellationSignal))

**How I obtain uris**

```
val cursor: Cursor = context.contentResolver.query(
            MediaStore.Audio.Albums.EXTERNAL_CONTENT_URI, projection,
            null, null, null
        )
 private fun Cursor?.parseAlbumUrls(): MutableList<String> {
        return parseCursor { cursor ->
            val id = cursor.getLong(MediaStore.Audio.Albums._ID)

            val imageUrl = if (Build.VERSION.SDK_INT >= 29) {
                ContentUris.withAppendedId(MediaStore.Audio.Albums.EXTERNAL_CONTENT_URI, id)
                    .toString()
            } else {
                cursor.getString(MediaStore.Audio.Albums.ALBUM_ART)
            }
        }
    }
```

**This is how I loaded this urls with Glide**

```

/**
 * Decodes [Audio MediaStore][MediaStore.Audio] [Uri]s into [InputStream]s.
 *
 * As of Android Q, album artworks could not longer be retrieved as a file path in [MediaStore.Audio.Albums.ALBUM_ART] ;
 * otherwise artworks are accessed from the content uri of its track or its album.
 *
 * @constructor
 * @param resolver The [ContentResolver] used to decode MediaStore uris.
 */
@RequiresApi(Build.VERSION_CODES.Q)
class MediaStoreAlbumArtworkLoader private constructor(
    private val resolver: ContentResolver
) : ModelLoader<Uri, InputStream> {

    override fun handles(model: Uri): Boolean = isMediaStoreAudioUri(model)

    private fun isMediaStoreAudioUri(uri: Uri): Boolean =
        ContentResolver.SCHEME_CONTENT == uri.scheme
                && MediaStore.AUTHORITY == uri.authority
                && "audio" in uri.pathSegments

    override fun buildLoadData(
        model: Uri,
        width: Int,
        height: Int,
        options: Options
    ): ModelLoader.LoadData<InputStream> {
        val diskCacheKey = ObjectKey(model)
        val fetcher = InputStreamFetcher(resolver, model, width, height)
        return ModelLoader.LoadData(diskCacheKey, fetcher)
    }

    /**
     * Factory for creating [MediaStoreAlbumArtworkLoader].
     *
     * @param resolver The [ContentResolver] used to decode [MediaStore] uris.
     */
    class InputStreamFactory(
        private val resolver: ContentResolver
    ) : ModelLoaderFactory<Uri, InputStream> {

        override fun build(
            multiFactory: MultiModelLoaderFactory
        ): ModelLoader<Uri, InputStream> = MediaStoreAlbumArtworkLoader(resolver)

        override fun teardown() {
            // Do nothing.
        }
    }

    /**
     * A [DataFetcher] that resolves a local [Uri] to load an album artwork for a [MediaStore] audio resource.
     * This internally uses [ContentResolver.openTypedAssetFile] to open a stream to the artwork,
     * specifying an optimal size in case multiple scaled versions of the artwork are available.
     *
     * While [ContentResolver.loadThumbnail] is the official way of loading those artworks,
     * we preferred [ContentResolver.openTypedAssetFile] (which is called internally by `loadThumbnail`)
     * as it produces an [InputStream], allowing Glide to decode it more efficiently.
     *
     * @param resolver ContentResolver used to decode [MediaStore] [Uri]s.
     * @param uri The uri pointing to a local track or album resource.
     * @param width The desired width of the loaded artwork.
     * @param height The desired height of the loaded artwork.
     */
    private class InputStreamFetcher(
        resolver: ContentResolver,
        uri: Uri,
        private val width: Int,
        private val height: Int
    ) : LocalUriFetcher<InputStream>(resolver, uri) {

        override fun getDataClass(): Class<InputStream> = InputStream::class.java

        @Throws(FileNotFoundException::class)
        override fun loadResource(uri: Uri, contentResolver: ContentResolver): InputStream {
            val optimalSizeOptions = Bundle(1)
            optimalSizeOptions.putParcelable(ContentResolver.EXTRA_SIZE, Point(width, height))

            return contentResolver.openTypedAssetFile(uri, "image/*", optimalSizeOptions, null)
                ?.createInputStream()
                ?: throw FileNotFoundException("FileDescriptor is null for: $uri")
        }

        @Throws(IOException::class)
        override fun close(data: InputStream) {
            data.close()
        }
    }
}
```

**Version**
1.3.2. The problem exists on all devices with android >= 28

krottv:
Fixes #922 
Not sure, whether we can use new method for contacts or not.
Btw, ContentUriFetcherTest are not passed on my device even without any modifications.

### DIFF
--- a/coil-base/src/androidTest/java/coil/fetch/ContentUriFetcherTest.kt
+++ b/coil-base/src/androidTest/java/coil/fetch/ContentUriFetcherTest.kt
@@ -15,6 +15,7 @@ import android.provider.ContactsContract.CommonDataKinds.StructuredName
 import android.provider.ContactsContract.Contacts.Photo.CONTENT_DIRECTORY
 import android.provider.ContactsContract.Contacts.Photo.DISPLAY_PHOTO
 import android.provider.ContactsContract.RawContacts
+import android.provider.MediaStore
 import androidx.core.net.toUri
 import androidx.test.core.app.ApplicationProvider
 import androidx.test.rule.GrantPermissionRule
@@ -44,6 +45,7 @@ class ContentUriFetcherTest {
 
     // Re-use the same contact across all tests. Must be created lazily.
     private val contactId by lazy(::createFakeContact)
+    private val albumId by lazy(::createFakeAlbum)
 
     @Before
     fun before() {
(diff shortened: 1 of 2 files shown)


colinrtwhite:
@krottv The test was failing so I removed the failing part and merged it in [here](https://github.com/coil-kt/coil/pull/967). I kept your commit history so you get credit - thanks again!

krottv:
@colinrtwhite awesome!

PR #972, opened 2021-11-05
Introduce AsyncImage.

colinrtwhite:
`AsyncImage` is a composable that executes an `ImageRequest` asynchronously and renders the result. It looks like this:

```kotlin
AsyncImage(
    model = "https://example.com/image.jpg",
    contentDescription = null
)
```

`AsyncImage` intentionally has a similar API to Compose's `Image` composable. The main difference is `AsyncImage` supports overriding the what's drawn for the loading/success/error states of the image request:

```kotlin
AsyncImage(
    model = "https://example.com/image.jpg",
    contentDescription = null,
    loading = {
         CircularProgressIndicator()
    },
    success = { state: State.Success ->
        Column {
            Image(
                painter = state.painter,
                contentDescription = null
            )
            Text("This image was loaded from: ${state.result.dataSource}.")
        }
    }
)
```

`AsyncImage` is intended to be a higher-level API than `rememberImagePainter`. It also fixes a number of issues/design challenges with `rememberImagePainter`:

- `AsyncImage` can correctly resolve the composable's size and doesn't need a fixed/bounded size. This makes it significantly easier to load images in a `LazyColumn` and other composables with unbounded width/height values.
- `AsyncImage` knows the correct `Scale` to load the image . If you use a custom `ContentScale`, you'll still have to set `Scale` manually.

Internally, `AsyncImage` uses `rememberImagePainter` to do the heavy lifting.

Fixes: https://github.com/coil-kt/coil/issues/973


Tolriq:
Thanks for this.

Can I ask again about handling of EmptyState? Loading state in case of contention or bounded dispatchers or many cases will happen too late and nothing is displayed during a fastscroll for example until the loading just jump in the middle of the screen.
Either having a placeholder param here or a param to tell to use the loading as placeholder for empty state too.

colinrtwhite:
@Tolriq It's fine to not overwrite the empty state as the `ImagePainter` will move to the `Loading` state synchronously. It won't have issues with dispatcher contention since Compose rendering occurs on the main thread and `Target.onStart` is executed in `Dispatchers.Main.immediate`.

Tolriq:
@colinrtwhite I do not see anything in this PR that change the order of the state.

Current implementation use the target.onStart, this is called from the async call that start a new coroutine / job this is not synchronous and can be delayed. It's probably a regression from moving to async in 2.0 and similar kind of unseen side effect of previous cancellation issue.

This is easily reproductible with fastscrolling with a 4 column grid for example.

colinrtwhite:
@Tolriq [This call](https://github.com/coil-kt/coil/blob/main/coil-base/src/main/java/coil/RealImageLoader.kt#L126) is synchronous because it is called from the main thread and it's launched using `Dispatchers.Main.immediate`.

Tolriq:
Well placeholder are delayed when using Coil placeholder or acting at LoadingState and works when applied at Empty state in my app.
I guess I'll have to build a repro or do a PR when this is merged.

hayahyts (approved):

Anton111111:
When new release planned with this cool things ?

PR #1006, opened 2021-11-26
Prepare 2.0.0-rc01.

colinrtwhite:


README.md
```diff
@@ -46,21 +45,40 @@ imageView.load("https://www.example.com/image.jpg") {
 }
 ```
 
+#### Jetpack Compose
+
+Import the [Jetpack Compose](https://developer.android.com/jetpack/compose) extension library:
+
+```kotlin
+implementation("io.coil-kt:coil-compose:1.4.0")
+```
```
voghDev, line 54: Hi!
I see this documentation update about the gradle dependency. Is coil `1.4.0` already available on `mavenCentral()`? 

I'm just trying to import it in my `app/build.gradle` file:

`implementation "io.coil-kt:coil-compose:1.4.0"` 

and it seems not available. If it's available on a different repository, maybe this documentation could be a good place to speak about it 🙂 
Thanks!

PR #1022, opened 2021-12-06
Remove inline from several suspend functions.

colinrtwhite:
This doesn't improve performance and only increases code sizes since the method is inlined, but also still exists in the class.


Goooler (commented):
👍

Issue #1025, opened 2021-12-07
Strange condition in AsyncImage


Anton111111:
@colinrtwhite , I confused by condition on this line:
https://github.com/coil-kt/coil/blob/0459229a69609bcf0f38967151b0eb96830b0556/coil-compose-base/src/main/java/coil/compose/AsyncImage.kt#L223

Looks like it must be:
`
if (loading = null && success = null && error = null)
`
Otherwise if we add to AsyncImage all 3 callbacks it will call DefaultContent.

colinrtwhite:
Good catch! Fixed here: https://github.com/coil-kt/coil/pull/1026

PR #1026, opened 2021-12-07
Fix using DefaultContent for AsyncImage if loading, success, and error are all set.

colinrtwhite:
Fixes: https://github.com/coil-kt/coil/issues/1025


Anton111111:
do you plan to publish alpha build with this fix in this year?

PR #1048, opened 2021-12-19
Improve AsyncImage performance.

colinrtwhite:
This PR focuses on improving `AsyncImage`'s performance through the following changes:

- `AsyncImage` now always uses `Layout` and never uses subcomposition. As a result, this PR moves `AsyncImage`'s `loading`/`success`/`error`/`content` composable arguments into a new composable, `SubcomposeAsyncImage`.
- Add painter arguments for `placeholder`/`error`/`fallback` and callbacks for `onLoading`/`onSuccess`/`onError`. Hopefully these arguments help users avoid using `loading`/`success`/`error`/`content`.
    - Also add overloads which accept `transform` and `onState` arguments. These are more flexible ways to override the `Painter` for a specific state and be notified about state changes.
    - These arguments don't restart the request if changed. I'm unsure if they should.
- Remove the `State` argument from the `content` composable function so users aren't automatically opted into recomposing when the painter's state changes.
- Remove code that was observing `AsyncImagePainter.state` inside `AsyncImagePainter.kt`.

There's a `coil-sample-compose` app in this repo that can be run to test out the improved scrolling performance.

EDIT: Updated to reflect the newest changes as of 26/01/2022.


coil-compose-base/src/androidTest/java/coil/compose/AsyncImagePainterTest.kt
```diff
@@ -612,6 +613,84 @@ class AsyncImagePainterTest {
...
+
+        waitForRequestComplete()
+
+        assertEquals(1, loadingCount.get())
+        assertEquals(0, successCount.get())
+        assertEquals(1, errorCount.get())
+    }
+
+    @Test
+    fun doesNotRecompose() {
```
colinrtwhite, line 675: This test currently fails, but a similar test for `AsyncImage` passes. If you know why, please let me know.

coil-compose-base/src/main/java/coil/compose/ContentPainterModifier.kt
```diff
@@ -0,0 +1,190 @@
...
+import androidx.compose.ui.unit.constrainHeight
+import androidx.compose.ui.unit.constrainWidth
+import kotlin.math.max
+import kotlin.math.roundToInt
+
+/**
+ * A custom [paint] modifier used by [AsyncImageContent] that fills the remaining space
+ * if one dimension of the incoming constraints is fixed.
+ */
+internal data class ContentPainterModifier(
```
colinrtwhite, line 33: I created a custom paint modifier since `AsyncImage` computes its size slightly differently than `Image`. We could just have a custom `LayoutModifier` and then set `Modifier.paint`, but I figured it was faster skip all the layout logic in `Modifier.paint` as it isn't used.

coil-compose-base/src/androidTest/java/coil/compose/AsyncImagePainterTest.kt
colinrtwhite: Called `AsyncImagePainter.onRemembered` synchronously was the cause.

Issue #1061, opened 2021-12-22
How add tag to http request.


xiazunyang:
My OkHttpClient can listen download progress by adding a DownloadProgressCallback as Tag to the Request.Builder. 
In Coil, I also need to add a Tag, but there is no such Api.

Here is my code for using OkHttpClient :
```
val request = Request.Builder()
    .tag(DlProgressCallback::class.java, DlProgressCallback { progress: Float ->
        //processing download progress
    })
    .tag(File::Class.java, file)
    ...
    .build()
val response = okHttpClient.newCall(request).execute()
```

I expect the Coil code:
```
    imageView.load(url) {
        error(error)
        placeholder(placeholder)
        //like add header
        addTag(DlProgressCallback { progress: Float ->  
             //processing download progress
        })
    }
```

colinrtwhite:
Coil 2.0 passes its `Parameters` as a tag. You can use this code to attach a tag:

```kotlin
imageView.load(url) {
    error(error)
    placeholder(placeholder)
    //like add header
    addParameter("progress_callback", DlProgressCallback { progress: Float ->  
         //processing download progress
    })
}

// Inside your interceptor:
val progressCallback = request.tag(Parameters::class.java)["progress_callback"] as DlProgressCallback
```

If we add a separate API for setting tags we'll have two APIs for attaching custom info to a request, which is confusing.

xiazunyang:
While looking at the latest source code, I also found Parameters, but they are part of the Coil, and I use DlProgressCallback in the okHttp3.Interceptor, just like DlProgressCallback.   It is another Java library that cannot dependence on Coil in the reverse direction, because Coil is an Android library.

Here is part of their code:
```
class ProgressInterceptor : Interceptor {

    override fun intercept(chain: Interceptor.Chain): Response {
        var request = chain.request()
        val upProgressCallback = request.getTag<UpProgressCallback>()
        val dlProgressCallback = request.getTag<DlProgressCallback>()
        if (upProgressCallback != null && request.body != null) {
            val progressRequestBody = ProgressRequestBody(request.body!!, upProgressCallback)
            request = request.newBuilder()
                .method(request.method, progressRequestBody)
                .build()
        }
        val response = chain.proceed(request)
        if (dlProgressCallback != null && response.body != null) {
            val progressResponseBody = ProgressResponseBody(response.body!!, dlProgressCallback)
            response = response.newBuilder()
                .body(progressResponseBody)
                .build()
        }
        return response
    }

}
[...]

PR #1062, opened 2021-12-22
Can add tag to okhttp.Request.Builder from ImageRequest.

xiazunyang:
See issues [#1061](https://github.com/coil-kt/coil/issues/1061)。


xiazunyang:
Lack of testing.

PR #1069, opened 2021-12-29
Update Coroutines (1.6.0)

chao2zhang:
- Most of the existing test code is already using Coroutine 1.6 API
- If we can remove usage of `runTestMain()` and `runTestAsync()`, the test code would be more idiomatic.
  - However, I have already spent quite some time refactoring even the production API.
  - I am posting this PR as the easy part for review.

## References
[1.6 guide](https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-test/README.md)
[1.6 migration](https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-test/MIGRATION.md)

### DIFF
--- a/buildSrc/src/main/kotlin/Library.kt
+++ b/buildSrc/src/main/kotlin/Library.kt
@@ -4,7 +4,7 @@ object Library {
 
     // CORE
 
-    private const val COROUTINES_VERSION = "1.6.0-RC3"
+    private const val COROUTINES_VERSION = "1.6.0"
     const val KOTLINX_COROUTINES_ANDROID = "org.jetbrains.kotlinx:kotlinx-coroutines-android:$COROUTINES_VERSION"
 
     const val ANDROIDX_ACTIVITY = "androidx.activity:activity-ktx:1.4.0"
--- a/coil-gif/src/androidTest/java/coil/transform/AnimationCallbacksTest.kt
+++ b/coil-gif/src/androidTest/java/coil/transform/AnimationCallbacksTest.kt
@@ -17,11 +17,10 @@ import coil.request.onAnimationStart
 import coil.request.repeatCount
 import coil.util.TestActivity
 import coil.util.activity
-import coil.util.runTestMain
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.first
-import kotlinx.coroutines.withTimeout
+import kotlinx.coroutines.test.runTest
 import org.junit.After
 import org.junit.Before
 import org.junit.Rule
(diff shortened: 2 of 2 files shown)


colinrtwhite (approved):
Thanks!

colinrtwhite:
@chao2zhang I don't think we can remove `runTestMain` and `runTestAsync` as certain tests need to be either run on the main thread or on a background thread and will fail if those conditions aren't met. Why do you want to remove them?

chao2zhang:
```
Calls to withContext(Dispatchers.IO), withContext(Dispatchers.Default) ,and withContext(Dispatchers.Main) are common in coroutines-based code bases. Unfortunately, just executing code in a test will not lead to these dispatchers using the virtual time source, so delays will not be skipped in them.
```
This is from the [coroutine guide](https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-test/README.md#virtual-time-support-with-other-dispatchers). `RealImageLoader.scope`, for instance, is against this design.

Test code is just yet another app user/client. Therefore library author is recommended to fully expose configurability of coroutineContext for its callers.

In the short term though, leaving the test code as is should be fine. We will need to revisit this one day if we were to support Kotlin multiplatform.

colinrtwhite:
@chao2zhang I don't think we can allow configuration of `RealImageLoader.scope` since its jobs rely on executing on the main thread. `executeMain` must be run on the main thread as it interacts with `Lifecycle`s and `View`s, which will throw exceptions if accessed off the main thread. If it's necessary to overwrite the main thread for tests, Coroutines has `Dispatchers.setMain` to set a custom dispatcher.

Also unfortunately I don't think it's likely that Coil supports KMM. We'd have to abstract away all AndroidX/Android platform/JVM dependencies first, which Coil has many for core functionality (e.g. `BitmapFactory`, `Drawable`, `View`, etc.).

chao2zhang:
Probably I didn't articulate it sufficiently. 
- All of the supervisor job, the main dispatcher, and exception handler are necessary for coil runtime.
- The problem based on my understanding for testability is to create a `CoroutineScope` in `RealImageLoader`, which is not visible in testing.
  -  Even `Dispatchers.setMain()` only intercepts the main dispatcher but not the scope.
  -  [Here](https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-test/README.md#accessing-the-test-coroutine-scope) exemplifies how to configure a `TestScope`. This looks like the way to go but I spent a few hours yesterday trying to get instrumentation tests to pass.

I will need to study more on the testing design in coroutine 1.6. It feels like test API in 1.6 is more restricted than the old API.

colinrtwhite:
@chao2zhang Ah I see - maybe we can use `execute` instead of `enqueue` if we need to configure a `TestScope`? It avoids using `RealImageLoader.scope` and only uses `withContext(Dispatchers.Main.immediate)` to execute the image load.

PR #1072, opened 2021-12-30
Fix scaling for Bitmap resources with original size

### LINKED ISSUE #1053: Png image size differs between different screen densities
Hello,

**Describe the bug**
The images are provided as pngs for the different screen densities (dpi, dpi, xhip, xxhdpi, xxxhdpi).  Before the images were loaded via `BitmapFactory.decodeResource(context.resources, id)` and had the same relative size on devices with different screen densities, also non standard one (e.g. 420dpi). 
After migrating to coil the loaded bitmaps differ in the displayed size and appear bigger on devices which do not have a standard screen density.
I would expect to have the same image size for different screen densities as it was with BitmapFactory.

**To Reproduce**
The image has to be saved as pngs in the folders for the different screen densities (as described here: https://developer.android.com/training/multiscreen/screendensities#TaskProvideAltBmp). Then load this image with coil setting the size to OriginalSize. Do not set it directly inside an image view, instead convert the drawable to a bitmap. Compare the size with the bitmap size loaded with BitmapFactory. 
This only applies if the used device does not have a standard screen density, e.g. 420 dpi. 
See for my implementation https://gist.github.com/CordulaMW/495a5c313198e04671cedf41e8373fbb

**Logs/Screenshots**

**Version**
io.coil-kt:coil-base:1.4.0
Does only appear on devices with non standard screen density, e.g. Pixel 3a (440dpi) or Pixel 2 (420dpi)

chao2zhang:
Resolves #1053

When the requested size is `Original`, previously it would fall into `(outWidth > 0 && outHeight > 0)`, where `val dstWidth = width.pxOrElse { srcWidth }` would always be evaluated to `srcWidth`, and `srcWidth` ignores the screen density. 

This PR addresses the case of android resources specifically by applying the same logic handled inside `BitmapFactory.decodeResourceStream()`.

## Testing
https://github.com/chao2zhang/coil/tree/chao/fixresourcescale-test
Tested on 440dpi and 560dpi devices using the code snippet in the original issue. Coil loaded image and android native BitmapFactory displays the image of the same size.

### DIFF
--- a/coil-base/src/main/java/coil/decode/BitmapFactoryDecoder.kt
+++ b/coil-base/src/main/java/coil/decode/BitmapFactoryDecoder.kt
@@ -2,14 +2,11 @@ package coil.decode
 
 import android.graphics.Bitmap
 import android.graphics.BitmapFactory
-import android.graphics.Matrix
-import android.graphics.Paint
-import android.graphics.RectF
 import android.os.Build.VERSION.SDK_INT
-import androidx.core.graphics.applyCanvas
-import androidx.core.graphics.createBitmap
 import androidx.exifinterface.media.ExifInterface
 import coil.ImageLoader
+import coil.decode.Exif.Data.Companion.toExifData
+import coil.decode.Exif.applyExifTransformations
 import coil.fetch.SourceResult
 import coil.request.Options
 import coil.size.isOriginal
@@ -23,7 +20,6 @@ import okio.Buffer
 import okio.ForwardingSource
 import okio.Source
 import okio.buffer
-import java.io.InputStream
 import kotlin.math.roundToInt
 
 /** The base [Decoder] that uses [BitmapFactory] to decode a given [ImageSource]. */
@@ -32,8 +28,6 @@ class BitmapFactoryDecoder @JvmOverloads constructor(
     private val options: Options,
     private val parallelismLock: Semaphore = Semaphore(Int.MAX_VALUE)
 ) : Decoder {
-
-    private val paint = Paint(Paint.ANTI_ALIAS_FLAG or Paint.FILTER_BITMAP_FLAG)
 
     override suspend fun decode() = parallelismLock.withPermit {
         runInterruptible { BitmapFactory.Options().decode() }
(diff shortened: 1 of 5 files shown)


coil-base/src/main/java/coil/decode/BitmapFactoryDecoder.kt
```diff
@@ -78,7 +79,15 @@ class BitmapFactoryDecoder @JvmOverloads constructor(
         // Always create immutable bitmaps as they have performance benefits.
         inMutable = false
 
-        if (outWidth > 0 && outHeight > 0) {
+        if (options.size.isOriginal && source.metadata is ResourceMetadata) {
```
colinrtwhite, line 82: This won't account for the case when only one dimension is `Dimension.Original`; also I'm not keen on adding a special case in `BitmapFactoryDecoder` only for resources. Maybe we only need to set the density after the bitmap is decoded and we can avoid creating a special case in `BitmapFactoryDecoder`?

colinrtwhite (commented):
Thanks for the fix! You'll need to run `./gradlew apiDump` to update the public API file. Also do we need to make similar changes in the other decoders? Is it possible to avoid adding special cases to each decoder?

coil-base/src/main/java/coil/decode/BitmapFactoryDecoder.kt
chao2zhang: _This won't account for the case when only one dimension is Dimension.Original_
Good callout when one dimension is `Dimension.Original`. Do you think we need to support mixed `Dimension.Original` with `Dimension.Pixel`? Is this something we can regard as invalid?

_Also do we need to make similar changes in the other decoders_
I would need to check the behavior of `ImageDecoderDecoder`. Other decoders are probably not density-aware.

_Is it possible to avoid adding special cases to each decoder? Maybe we can set the density after the bitmap is decoded and avoid creating a special case in BitmapFactoryDecoder?_
1. The problem with resetting the density after `BitmapFactory.decode()` is that, we probably need another bitmap transformation to scale the previous image to meet the dimension need.
2. For additional special case, it might be necessary because Android `BitmapFactory` and `ImageDecoder` both expose specific API to decode native `Resource`. `Resource` is unique from other sources in that it is density-aware.

chao2zhang: I played around with `ImageDecoderDecoder` and `GifDecoder`. They seem to be very different from the default `BitmapFactory` decoder - Therefore this PR might be better to scoped to only handle `Bitmap`. I have updated the PR title accordingly.

carstenhag: >Do you think we need to support mixed Dimension.Original with Dimension.Pixel?

My colleague @CordulaMW created the issue #1053 for this PR, so I can comment on at least what we need: We don't need mixed Dimension values. We would either use `Dimension.Original` or `Dimension.Pixel`, never both at the same time.  

I generally don't think it would make sense to support this, as it's afaik only for scaling. Why would someone want the height dpi-scaled but the width not?   
As a counter argument, the android layout system does support it. Not sure.

carstenhag:
Hello there, I have tried out the PR and it works great for me at least :)

colinrtwhite (approved):
Sorry for the delay on this. I agree let's tackle this only for the case where both dimensions are `Dimension.Original`. We can "fix" the behaviour for single dimension later if someone has a strong argument that it should be handled differently than it currently is. Thanks again!

PR #1100, opened 2022-01-21
Convert AsyncImage.contentDescription to be an optional argument.

colinrtwhite:
Creating this PR for consideration before the 2.x API is finalized. Compose's built in `Image` composable has `contentDescription` as a required parameter to encourage users to set it to a not-null value, which is important for users who use screen readers. To be clear, accessibility is important and users should set a content description (we do at Instacart)! That said, I think there are some differences between `Image` and `AsyncImage` that might make us want to choose a different default:

- `Image` typically displays static content from a local source. `AsyncImage` typically displays dynamic content from a remote source where it's tough to know what the content description should be without a separate endpoint.
- `AsyncImage` is a more complicated composable than `Image`. Having an extra required argument can crowd the callsite especially if it's often set to `null`.
- Coil's `AsyncImage` takes notes from [SwiftUI's `AsyncImage`](https://developer.apple.com/documentation/swiftui/asyncimage) which does not have `contentDescription` as a required parameter.


colinrtwhite:
After thinking about this, I think it makes sense to keep `contentDescription` as required for many of the same reasons that `Image` requires it. Making it a required param signals that it's important to set it! You have to consciously set it to `null` if it's not possible to provide alt text (or if the image is decorative and shouldn't be read by a screen reader).

If you're getting the `model` (e.g. a URL) from a dynamic source (e.g. a backend API), I'd recommend having that dynamic source also return a content description for the model if possible. If you're hardcoding a model it should be easier to provide the content description using a `stringResource`.

Long term it might make sense for Coil to support a `Content-Description` header, though that will take some extra work to support and can be added in later. It's also unclear how frequently that header is set.

PR #1115, opened 2022-01-28
Use Okio in Cache

yschimke:
Following up on https://github.com/coil-kt/coil/discussions/1105

Questions:
- How far should this change go?
- should the fileSystem be on interface Snapshot? OR implied by context of the Cache always?

ImageSource will break for a non SYSTEM FileSystem

`ImageSource(file = data.toFile()`


coil-base/src/main/java/coil/disk/DiskCache.kt
```diff
@@ -115,8 +125,18 @@ interface DiskCache {
...
+        fun directory(directory: File) =
+            directory(directory = directory.toOkioPath(), fileSystem = FileSystem.SYSTEM)
+
+        /**
+         * Set the [directory] where the cache stores its data.
+         *
+         * IMPORTANT: It is an error to have two [DiskCache] instances active in the same
+         * directory at the same time as this can corrupt the disk cache.
+         */
+        fun directory(directory: Path, fileSystem: FileSystem = FileSystem.SYSTEM) = apply {
```
colinrtwhite, line 137: I'd vote to have a separate `DiskCache.Builder.fileSystem` function instead of a default argument.

coil-base/src/main/java/coil/disk/DiskCache.kt
```diff
@@ -102,7 +111,8 @@ interface DiskCache {
 
     class Builder {
 
-        private var directory: File? = null
+        private var directory: Path? = null
+        private var fileSystem: FileSystem? = null
```
colinrtwhite, line 115: I think we should default to `FileSystem.SYSTEM`.

coil-base/src/main/java/coil/fetch/HttpUriFetcher.kt
```diff
@@ -221,15 +228,18 @@ internal class HttpUriFetcher(
...
         } catch (_: IOException) {
             // If we can't parse the metadata, ignore this entry.
             return null
         }
     }
 
     private fun DiskCache.Snapshot.toImageSource(): ImageSource {
-        return ImageSource(file = data, diskCacheKey = diskCacheKey, closeable = this)
+        // TODO discuss supporting Path in ImageSource as well
+        return ImageSource(file = data.toFile(), diskCacheKey = diskCacheKey, closeable = this)
```
colinrtwhite, line 242: I hadn't thought about it, but it probably makes sense to have `ImageSource` use Okio's filesystem as well. Though, internally in one of the implementations we use `File.createTempFile` which only works on the system filesystem so we'll have to figure out what to do that. Let's leave it as is in this PR for now. What do you think?

Fortunately, `ImageSource` is a sealed class so we can change the implementations to use Okio's filesystem and expose `Path`s without breaking binary compatibility later if needed.

coil-base/src/main/java/coil/disk/DiskCache.kt
```diff
@@ -59,12 +65,13 @@ interface DiskCache {
      * disk cache. To modify the contents of those files, use [edit].
      */
     interface Snapshot : Closeable {
+        val fileSystem: FileSystem
```
colinrtwhite, line 68: I'm not certain, but I think the `FileSystem` should be implied by `DiskCache.fileSystem` otherwise there could be confusion (i.e. can each individual `Snapshot`/`Editor` have a different `FileSystem` from `DiskCache.fileSystem`?). What do you think?

colinrtwhite (commented):
Thanks for doing this! Left a few comments.

yschimke:
Took another iteration on this.  Went deeper with Okio.

It gets a bit awkward in the Android layer, when it expects files.  Introduced MediaDataSource in one case, but the other assumes files.

But should provoke further discussion.

coil-base/src/main/java/coil/decode/ImageSource.kt
```diff
@@ -133,14 +141,18 @@ sealed class ImageSource : Closeable {
...
+    abstract fun file(): Path
 
     /**
      * Return a [File] containing this [ImageSource]'s data if one has already been created.
      * Else, return 'null'.
      */
-    abstract fun fileOrNull(): File?
+    abstract fun fileOrNull(): Path?
+
+    abstract fun mediaDataSource(): MediaDataSource
```
colinrtwhite, line 155: I don't think we should expose a custom `MediaDataSource` here. Originally `ImageSource` was only a `BufferedSource` to keep `Decoder`s agnostic to the data source. Unfortunately, Android has several decoders that are coupled with `File`s which is why `file` and `fileOrNull` were added. `ImageSource.Metadata` was added recently, however it's optional.

coil-video/src/main/java/coil/decode/VideoFrameDecoder.kt
```diff
@@ -172,7 +173,12 @@ class VideoFrameDecoder(
                 setDataSource("android.resource://${metadata.packageName}/${metadata.resId}")
             }
             else -> {
-                setDataSource(source.file().path)
+                // TODO check how to handle this cleanly
+                if (SDK_INT >= Build.VERSION_CODES.M) {
+                    setDataSource(source.mediaDataSource())
+                } else {
+                    setDataSource(source.file().toFile().path)
```
colinrtwhite, line 180: Let's always use `setDataSource(source.file().toFile().path)` and avoid creating a custom media source (even if it's unsafe).

coil-base/src/main/java/coil/fetch/FileFetcher.kt
```diff
@@ -5,13 +5,18 @@ import coil.ImageLoader
...
 import java.io.File
 
-internal class FileFetcher(private val data: File) : Fetcher {
+internal class FileFetcher(private val data: Path, private val fileSystem: FileSystem) : Fetcher {
+    constructor(data: File) : this(data.toOkioPath(), FileSystem.SYSTEM)
 
     override suspend fun fetch(): FetchResult {
         return SourceResult(
-            source = ImageSource(file = data),
+            source = ImageSource(file = data, fileSystem = fileSystem),
```
colinrtwhite, line 19: Let's just do `data.toOkioPath()` here instead.

coil-base/src/main/java/coil/fetch/FileFetcher.kt
```diff
@@ -5,13 +5,18 @@ import coil.ImageLoader
...
 import coil.decode.ImageSource
 import coil.request.Options
+import coil.util.extension
+import okio.FileSystem
+import okio.Path
+import okio.Path.Companion.toOkioPath
 import java.io.File
 
-internal class FileFetcher(private val data: File) : Fetcher {
+internal class FileFetcher(private val data: Path, private val fileSystem: FileSystem) : Fetcher {
```
colinrtwhite, line 14: Let's keep the `data: File` constructor only; it's convention for `Fetcher`s to perform any data mapping inside `fetch`.

coil-base/src/main/java/coil/decode/ImageSource.kt
```diff
@@ -133,14 +141,18 @@ sealed class ImageSource : Closeable {
...
      */
-    abstract fun file(): File
+    abstract fun file(): Path
 
     /**
      * Return a [File] containing this [ImageSource]'s data if one has already been created.
      * Else, return 'null'.
      */
-    abstract fun fileOrNull(): File?
+    abstract fun fileOrNull(): Path?
```
colinrtwhite, line 153: I think we also need a `val fileSystem: FileSystem` for `ImageSource`, right? We can always return `FileSystem.SYSTEM` for image sources not backed by a file.

coil-base/src/main/java/coil/decode/ImageSource.kt
```diff
@@ -133,14 +141,18 @@ sealed class ImageSource : Closeable {
      * Return a [File] containing this [ImageSource]'s data.
      * If this image source is backed by a [BufferedSource], a temporary file containing this
      * [ImageSource]'s data will be created.
+     *
+     * This method is only safe when used on FileSystem.SYSTEM.
```
colinrtwhite, line 145: I think this is always safe, right? We only need to create a temp file if the `ImageSource` is backed by a `BufferedSource` in which case we should always use `FileSystem.SYSTEM` to create the temp file.

yschimke:
Thanks for the multiple reviews on this. I'll iterate in next few weeks.

colinrtwhite (approved):
@yschimke Is this good to merge or do you want to keep iterating on this? Thanks again for doing this!

yschimke:
> @yschimke Is this good to merge or are you still working on it (LGTM)? Thanks again for doing this!

All good for me, let me know if there are other related changes to make.

PR #1129, opened 2022-02-09
Migrate to Gradle version catalogs.

colinrtwhite:

PR #1131, opened 2022-02-09
Enable Gradle's type-safe project accessors

Goooler:
Follow up #1129.

https://docs.gradle.org/7.0/userguide/declaring_dependencies.html#sec:type-safe-project-accessors


colinrtwhite (approved):
Nice, catch! Thanks

PR #1140, opened 2022-02-14
Use Java 17 on CI.

colinrtwhite:


.github/workflows/ci.yml
```diff
@@ -17,7 +17,8 @@ jobs:
             - uses: actions/setup-java@v2
               with:
                   distribution: 'adopt'
-                  java-version: 11
+                  java-version: 17
```
Goooler, line 20: Suggesting use `zulu` instead of `adopt`,
https://api.adoptopenjdk.net/

colinrtwhite: @Goooler What's the benefit? I thought `zulu` was best for M1 Macs, though I don't think the Github Mac runners are M1.

Goooler: As jake recommending this.

colinrtwhite: Sounds good to me :)

PR #1146, opened 2022-02-15
Move plugins into libs.versions.toml

Goooler:
Seems we can use version catalogs by adding `buildSrc/settings.gradle.kts`, now all deps are declared in `gradle/libs.versions.toml`.

### DIFF
--- a/buildSrc/settings.gradle.kts
+++ b/buildSrc/settings.gradle.kts
@@ -1,0 +1,7 @@
+dependencyResolutionManagement {
+    versionCatalogs {
+        create("libs") {
+            from(files("../gradle/libs.versions.toml"))
+        }
+    }
+}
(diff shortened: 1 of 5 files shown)


buildSrc/settings.gradle.kts
```diff
@@ -0,0 +1,7 @@
+dependencyResolutionManagement {
+    versionCatalogs {
+        create("libs") {
+            from(files("../gradle/libs.versions.toml"))
+        }
+    }
+}
```
Goooler, line 7: Inspired by https://github.com/detekt/detekt/blob/2aad546d74bb8da669bddee48752209e3ec7ebfc/build-logic/settings.gradle.kts#L3-L9

Goooler: https://docs.gradle.org/current/userguide/platforms.html#sec:sharing-catalogs

colinrtwhite (approved):
Nice!

PR #1149, opened 2022-02-16
Bump ktlint to 0.44.0

Goooler:


build.gradle.kts
```diff
@@ -46,6 +45,11 @@ tasks.withType<DokkaMultiModuleTask>().configureEach {
     ))
 }
 
+ktlint {
+    version by libs.versions.ktlint.get()
+    disabledRules by setOf("indent", "max-line-length", "parameter-list-wrapping")
+}
+
```
Goooler, line 52: It should run `ktlintCheck` & `ktlintFormat` on all projects parallelly.

colinrtwhite: @Goooler Are you sure about this? When I tested this change locally it only runs `ktlint` on the root project; `./gradlew coil-base:ktlintCheck` doesn't work.

Goooler: You are right, sorry for miss-testing. 
I'll close this PR, as finding a bug in the current configuration, it's not working to check or format KTS files in submodules, I'm investigating that.

Goooler: I know why, as you disable the `indent` rule, I've added a extra blank to test like this:
```kotlin
plugins {
     id("com.android.library")
}
```
Seems like all `indent-size` checked don't work now, if we need re-enable this rule?

PR #1276, opened 2022-05-15
Clean up documentation after deploying.

colinrtwhite:


coil-base/src/main/java/coil/fetch/Fetcher.kt
```diff
@@ -10,8 +10,8 @@ import coil.request.Options
  *
  * To accomplish this, fetchers fit into one of two types:
  *
- * - Uses the data as a key to fetch bytes from a remote source (e.g. network or disk)
- *   and exposes it as an [ImageSource]. e.g. [HttpUriFetcher]
+ * - Uses the data as a key to fetch bytes from a remote source (e.g. network, disk)
+ *   and exposes it as an [ImageSource] (e.g. [HttpUriFetcher]).
```
Jawnnypoo, line 14: This leads to a 404 when you click on it, see https://coil-kt.github.io/coil/api/coil-base/coil-base/coil.fetch/-fetcher/

Maybe because the class is `internal`?

Jawnnypoo: @colinrtwhite

Issue #1296, opened 2022-05-30
Allow specifying a relative frame location for VideoFrameDecoder


zhanghai:
**Is your feature request related to a problem? Please describe.**

The current `videoFrameMicros()` option allows specifying an absolute frame location for retrieving the video frame. However, I assume that might be rarely useful for generic video files because a video file can always have a shorter duration than the one specified, unless it stays the default `0`.

In contrast, GNOME Nautilus uses totem-video-thumbnailer by default which tries multiple relative frame locations (`1/3`, `2/3`, `0.1`, `0.9`, `0.5`) - see also https://gitlab.gnome.org/GNOME/totem/-/blob/master/src/totem-video-thumbnailer.c#L543 . Meanwhile, ffmpegthumbnailer using `0.1` which can be found at https://github.com/dirkvdb/ffmpegthumbnailer/blob/2858726b318ec1c95e9edd3f0863f7d843dd63d2/libffmpegthumbnailer/videothumbnailer.cpp#L49 .

So it would be nicer if `coil-video` also allows specifying a relative frame location.

**Describe the solution you'd like**

```
fun ImageRequest.Builder.videoFrameFraction(frameFraction: Double): ImageRequest.Builder {
    require(frameFraction in 0.0..1.0) { "frameFraction must be >= 0 and <= 1." }
    return setParameter(VideoFrameDecoder.VIDEO_FRAME_FRACTION_KEY, frameFraction)
}

fun Parameters.videoFrameFraction(): Double? =
    value(VideoFrameDecoder.VIDEO_FRAME_FRACTION_KEY) as Double?
...
class VideoFrameDecoder(
...
    override suspend fun decode() = MediaMetadataRetriever().use { retriever ->
...
            val durationMillis =
                retriever.extractMetadata(MediaMetadataRetriever.METADATA_KEY_DURATION)
                    ?.toLongOrNull() ?: 0L
            val frameFraction = options.parameters.videoFrameFraction() ?: 0.0
            val frameMicros = TimeUnit.MICROSECONDS.convert(
                (frameFraction * durationMillis).roundToLong(), TimeUnit.MILLISECONDS
            )
...
    companion object {
...
        const val VIDEO_FRAME_FRACTION_KEY = "coil#video_frame_fraction"
...
    }
...
```

**Additional context**

The code above is take from my modified copy of `VideoFrameDecoder` and I removed the absolute frame location support there. If we need to support both, maybe make the relative one override the absolute one?

[...]

PR #1318, opened 2022-06-11
Add videoFramePercent() for VideoFrameDecoder.

zhanghai:
Fixes: #1296

### DIFF
--- a/coil-video/src/main/java/coil/decode/VideoFrameDecoder.kt
+++ b/coil-video/src/main/java/coil/decode/VideoFrameDecoder.kt
@@ -3,6 +3,7 @@ package coil.decode
 import android.graphics.Bitmap
 import android.graphics.Paint
 import android.media.MediaMetadataRetriever
+import android.media.MediaMetadataRetriever.METADATA_KEY_DURATION
 import android.media.MediaMetadataRetriever.METADATA_KEY_VIDEO_HEIGHT
 import android.media.MediaMetadataRetriever.METADATA_KEY_VIDEO_ROTATION
 import android.media.MediaMetadataRetriever.METADATA_KEY_VIDEO_WIDTH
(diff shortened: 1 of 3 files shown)


colinrtwhite (approved):
Great, thanks!

coil-video/src/main/java/coil/decode/VideoFrameDecoder.kt
```diff
@@ -35,7 +40,13 @@ class VideoFrameDecoder(
     override suspend fun decode() = MediaMetadataRetriever().use { retriever ->
         retriever.setDataSource(source)
         val option = options.parameters.videoFrameOption() ?: OPTION_CLOSEST_SYNC
-        val frameMicros = options.parameters.videoFrameMicros() ?: 0L
+
+        val frameMicros = options.parameters.videoFrameMicros()
+            ?: options.parameters.videoFramePercent()?.let { framePercent ->
+                val durationMillis = retriever.extractMetadata(METADATA_KEY_DURATION)
+                    ?.toLongOrNull() ?: 0L
+                MICROSECONDS.convert((framePercent * durationMillis).roundToLong(), MILLISECONDS)
```
colinrtwhite, line 48: Oops - looks like this should be the other way around. I'll correct it on `main` + add a test.

zhanghai: Why? [`TimeUnit.convert()`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/TimeUnit.html#convert-long-java.util.concurrent.TimeUnit-) takes `long sourceDuration, TimeUnit sourceUnit`, and we are indeed converting from milliseconds to microseconds here.

colinrtwhite: Ah sorry my mistake! Either way I've added a regression test on `main` to verify the behaviour.

PR #1365, opened 2022-07-22
Update doc/recipies/okhttp section to align 2 sample code

kuramu1108:
## Cause

The Receipes-Using a custom OkHttpClient-Header sections provides 2 approaches for add header to image request
while the sample code in the 2 approaches is actually doing the different thing
- first one setting header for request, while
- second one [overwrite header of the response](https://square.github.io/okhttp/features/interceptors/#rewriting-responses)

## Changes
Algin 2 sample code to accomplish the same outcome

### DIFF
--- a/docs/recipes.md
+++ b/docs/recipes.md
@@ -50,7 +50,7 @@ Headers can be added to your image requests in one of two ways. You can set head
 ```kotlin
 val request = ImageRequest.Builder(context)
     .data("https://www.example.com/image.jpg")
-    .setHeader("Cache-Control", "max-age=31536000")
+    .setHeader("Cache-Control", "no-cache")
     .target(imageView)
     .build()
 imageLoader.execute(request)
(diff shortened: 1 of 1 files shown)


docs/recipes.md
```diff
@@ -50,7 +50,7 @@ Headers can be added to your image requests in one of two ways. You can set head
 ```kotlin
-    .setHeader("Cache-Control", "max-age=31536000")
```
colinrtwhite, line 53: I think we should probably leave this as `max-age=31536000`. It's more typical to overwrite the `cache-control` header to cache for longer instead of disabling caching completely.

docs/recipes.md
```diff
@@ -59,22 +59,24 @@ imageLoader.execute(request)
 Or you can create an OkHttp [`Interceptor`](https://square.github.io/okhttp/interceptors/) that sets headers for every request executed by your `ImageLoader`:
 
 ```kotlin
-class ResponseHeaderInterceptor(
```
colinrtwhite, line 62: Good catch, thanks!

docs/recipes.md
```diff
@@ -59,22 +59,24 @@ imageLoader.execute(request)
...
+            .build()
+        return chain.proceed(request)
     }
 }
 
 val imageLoader = ImageLoader.Builder(context)
     .okHttpClient {
         OkHttpClient.Builder()
             // This header will be added to every image request.
-            .addNetworkInterceptor(ResponseHeaderInterceptor("Cache-Control", "max-age=31536000,public"))
```
colinrtwhite, line 77: Same here re: leaving this as `max-age=31536000` (we can probably remove the `public`).

colinrtwhite (approved):
Looks good, thanks!

PR #1375, opened 2022-07-29
Update androidx libraries.

colinrtwhite:


colinrtwhite:
Punting on these updates since they're likely to cause compatibility issues (see the build failure) for users and there's no strong reason update the dependency in Coil. Will re-evaluate this if `androidx.compose.ui:ui-test-junit4` updates its `androidx.activity` dependency to `1.5+` since that fixes the compatibility issue.

Issue #1349, opened 2022-07-07
Load very big image throw OOM


XuQK:
**Describe the bug**
Load a 107MB png image, sometimes throw OOM.

**Logs/Screenshots**
attached.
[2022-07-07 14:36:25.093 4541-4541.txt](https://github.com/coil-kt/coil/files/9061113/2022-07-07.14.36.25.093.4541-4541.txt)

**Version**
coil-2.1.0
What library version are you using? Does this occur on a specific API level or Android device?

colinrtwhite:
There isn’t enough information in this report. What code are you using to load the image. How can I reproduce this?

XuQK:
I success reproduce it in a empty project just now.

1. Download images attached, copy them 3 times, now we have 15 images.
2. Put 15 images to /sdcard/Pictures/bigpic.
3. Install the demo, give it storage permission manual in Settings.
4. Then Run the demo, you should see there are image list loaded, keep sliding up and down for a while, then the OOM occured.
5. If not reproduce, copy the picture into 30 pieces, repeat.

Wait for your news.

This is demo:
[CoilTest.zip](https://github.com/coil-kt/coil/files/9068647/CoilTest.zip)

This is images:
https://drive.google.com/file/d/1-CNSN7AxCfsG7TX4PKU2CUvLeDLlu6rF/view?usp=sharing

colinrtwhite:
Thanks for the repro project. I’ll take a look soon.

spacecowboy:
Same issue in Coil 2.2.0 when trying to load this image (on an Android Emulator):

http://www.corsix.org/images/495f98eb0abe80037a8bfb0b3dfb943df9e04495.svg

Note that the image is only 14kb but coil is attempting to decode a 142mb file?

```
2022-08-31 00:39:32.474 6398-6398/com.nononsenseapps.feeder.debug E/AndroidRuntime: FATAL EXCEPTION: main
    Process: com.nononsenseapps.feeder.debug, PID: 6398
    java.lang.RuntimeException: Canvas: trying to draw too large(142657200bytes) bitmap.
        at android.graphics.RecordingCanvas.throwIfCannotDraw(RecordingCanvas.java:266)
        at android.graphics.BaseRecordingCanvas.drawBitmap(BaseRecordingCanvas.java:94)
        at androidx.compose.ui.graphics.AndroidCanvas.drawImageRect-HPBpro0(AndroidCanvas.android.kt:271)
[...]

PR #1421, opened 2022-09-01
Added try-catch for RuntimeException when image is too large

spacecowboy:
Related to #1349


colinrtwhite:
Thanks for the submission, but I think adding a `try/catch` when drawing only masks the issue. Additionally this change creates different behaviour for Compose and Views.

Ideally we should fail after decoding before returning the result if the `Drawable` is going to be too large. Though, we've held off doing that so far since it's not easy to know what's too large as it varies by device/manufacturer/API level.

PR #1424, opened 2022-09-05
Don't use kotlin-parcelize plugin

luca020400:
The kotlin-parcelize plugin isn't supported in AOSP
( read, Parcelable methods aren't generated )

Since there's only one usage in the library
it's rather easy to re-implement it manually
instead of using code-generation

Signed-off-by: Luca Stefani <luca.stefani.ge1@gmail.com>

### DIFF
--- a/coil-base/api/coil-base.api
+++ b/coil-base/api/coil-base.api
@@ -413,6 +413,7 @@ public final class coil/memory/MemoryCache$Builder {
 
 public final class coil/memory/MemoryCache$Key : android/os/Parcelable {
 	public static final field CREATOR Landroid/os/Parcelable$Creator;
+	public static final field Companion Lcoil/memory/MemoryCache$Key$Companion;
 	public fun <init> (Ljava/lang/String;Ljava/util/Map;)V
 	public synthetic fun <init> (Ljava/lang/String;Ljava/util/Map;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
 	public final fun copy (Ljava/lang/String;Ljava/util/Map;)Lcoil/memory/MemoryCache$Key;
@@ -424,6 +425,9 @@ public final class coil/memory/MemoryCache$Key : android/os/Parcelable {
 	public fun hashCode ()I
 	public fun toString ()Ljava/lang/String;
 	public fun writeToParcel (Landroid/os/Parcel;I)V
+}
+
+public final class coil/memory/MemoryCache$Key$Companion {
 }
 
 public final class coil/memory/MemoryCache$Key$Creator : android/os/Parcelable$Creator {
--- a/coil-base/src/main/java/coil/memory/MemoryCache.kt
+++ b/coil-base/src/main/java/coil/memory/MemoryCache.kt
@@ -3,12 +3,12 @@ package coil.memory
 import android.content.ComponentCallbacks2
 import android.content.Context
 import android.graphics.Bitmap
+import android.os.Parcel
 import android.os.Parcelable
 import androidx.annotation.FloatRange
 import coil.key.Keyer
 import coil.util.calculateMemoryCacheSize
 import coil.util.defaultMemoryCacheSizePercent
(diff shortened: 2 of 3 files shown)


colinrtwhite:
Can you explain a bit more about your use case? Are you importing Coil's source into AOSP?

coil-base/src/main/java/coil/memory/MemoryCache.kt
```diff
@@ -78,6 +77,38 @@ interface MemoryCache {
...
+                parcel.writeString(it.key)
+                parcel.writeString(it.value)
+            }
+        }
+
+        override fun describeContents(): Int {
+            return 0
+        }
+
+        companion object CREATOR : Parcelable.Creator<Key> {
```
colinrtwhite, line 94: Let's make this just an `object` instead of `companion object` since we don't want to support all the syntax that comes with a `companion object`.

colinrtwhite:
Looks like the [binary compatibility check failed](https://github.com/coil-kt/coil/runs/8198724983?check_suite_focus=true). You'll need to ensure that the `CREATOR`'s existing public methods aren't changed/removed.

luca020400:
> Can you explain a bit more about your use case? Are you importing Coil's source into AOSP?

Ideally yes, building in AOSP would be the final use case, and aside plain kotlin stdlib that was the only "extension" that was used to generate code
AOSP build system doesn't really like gradle plugins :)

luca020400:
> Looks like the [binary compatibility check failed](https://github.com/coil-kt/coil/runs/8198724983?check_suite_focus=true). You'll need to ensure that the `CREATOR`'s existing public methods aren't changed/removed.

My bad, I didn't run API check to verify I didn't break public APIs, I'll be doing that and make it compatible

coil-base/api/coil-base.api
```diff
@@ -426,6 +427,9 @@ public final class coil/memory/MemoryCache$Key : android/os/Parcelable {
 	public fun writeToParcel (Landroid/os/Parcel;I)V
 }
 
+public final class coil/memory/MemoryCache$Key$Companion {
```
luca020400, line 430: I'd rather not have to expose the Companion object, even if empty, but it gets generated regardless

coil-base/src/main/java/coil/memory/MemoryCache.kt
```diff
@@ -78,6 +77,43 @@ interface MemoryCache {
...
+            return 0
+        }
+
+        companion object {
+            @JvmField
+            val CREATOR: Parcelable.Creator<*> = Creator()
+        }
+
+        class Creator : Parcelable.Creator<Key> {
+            final override fun createFromParcel(parcel: Parcel): Key {
```
luca020400, line 100: the final keyword isn't needed per-se, and kotlin wants me to remove it, bit I couldn't find a way to make the API see "final" in there, and simply adding it works...

luca020400 (commented):
I don't usually deal with public facing APIs so there may be some magic tricks to do what I commented about.

coil-base/api/coil-base.api
colinrtwhite: Ah yep looks like there's no way around this unfortunately (at least until Kotlin adds namespaces).

coil-base/src/main/java/coil/memory/MemoryCache.kt
colinrtwhite: Interesting - I think making a function non-final doesn't break binary compatibility so it's probably safe to remove. The class is also final by default.

colinrtwhite (approved):
Thanks!

coil-base/src/main/java/coil/memory/MemoryCache.kt
luca020400: Feel free to remove them in a follow-up change if you're sure it's not breaking ( I'm leaning on that )
I tried to have a 1:1 match with current exposed API for simplicity.

PR #1427, opened 2022-09-06
Hide the companion object in MemoryCache. Minor clean up.

colinrtwhite:


coil-base/api/coil-base.api
```diff
@@ -427,14 +426,11 @@ public final class coil/memory/MemoryCache$Key : android/os/Parcelable {
 	public fun writeToParcel (Landroid/os/Parcel;I)V
 }
 
-public final class coil/memory/MemoryCache$Key$Companion {
-}
-
 public final class coil/memory/MemoryCache$Key$Creator : android/os/Parcelable$Creator {
 	public fun <init> ()V
-	public final fun createFromParcel (Landroid/os/Parcel;)Lcoil/memory/MemoryCache$Key;
+	public fun createFromParcel (Landroid/os/Parcel;)Lcoil/memory/MemoryCache$Key;
```
colinrtwhite, line 431: https://stackoverflow.com/questions/10969607/does-removing-the-final-keyword-affect-binary-compatibility

PR #1429, opened 2022-09-06
Add test for MemoryCache.Key parcel implementation.

colinrtwhite:


coil-base/api/coil-base.api
```diff
@@ -426,14 +426,6 @@ public final class coil/memory/MemoryCache$Key : android/os/Parcelable {
 	public fun writeToParcel (Landroid/os/Parcel;I)V
 }
 
-public final class coil/memory/MemoryCache$Key$Creator : android/os/Parcelable$Creator {
```
colinrtwhite, line 429: After searching, this class was actually never part of the public API and isn't possible to reference when compiling.

Issue #1327, opened 2022-06-16
Support a FakeImageLoader for typical Preview use cases


yschimke:
The https://coil-kt.github.io/coil/image_loaders/#testing docs suggest implementing a custom ImageLoader.

But there could be some useful defaults, like loading DrawableRes, or mapping a url to a test image.

https://github.com/google/horologist/blob/fdc4fa6b4bb4387c2ec7804a9c5eb9c67f909764/tiles/src/androidTest/java/com/google/android/horologist/tiles/FakeImageLoader.kt

It would be nice to have a Test or Fake ImageLoader that implemented sensible defaults for situations like Previews.

Would this be accepted as a PR?

colinrtwhite:
I agree, having a `coil-fakeimageloader` artifact (or similar) would be useful. I've been a bit apprehensive to add a 1st party artifact for it since it would require ongoing support vs. providing a recipe that can be copied + modified.

I'm going on holiday until the end of July so let's circle back on this then!

colinrtwhite:
@yschimke Took an initial stab at this [here](https://github.com/coil-kt/coil/pull/1451). It's not finished, but let me know if you have any feedback!

PR #1451, opened 2022-09-12
Add support for new testing artifact.

colinrtwhite:
This PR adds support for a new `coil-test` artifact which has a `FakeImageLoaderEngine` class that can be used to hardcode responses for your `ImageLoader` in screenshot tests:

```kotlin
val engine = FakeImageLoaderEngine.Builder()
    .intercept("https://www.example.com/image.jpg", ColorDrawable(Color.RED))
    .intercept({ it is String && it.endsWith("test.png") }, ColorDrawable(Color.GREEN))
    .default(ColorDrawable(Color.BLUE))
    .build()
    .components { add(engine) }
    .build()
Coil.setImageLoader(imageLoader)
```

Fixes: https://github.com/coil-kt/coil/issues/1327


coil-fakeimageloader/src/main/java/coil/FakeImageLoader.kt
```diff
@@ -0,0 +1,232 @@
...
+    return FakeImageLoader.Builder().build()
+}
+
+/** Convenience function to get [FakeImageLoader.memoryCache] as a [FakeMemoryCache]. */
+val FakeImageLoader.fakeMemoryCache: FakeMemoryCache get() = memoryCache as FakeMemoryCache
+
+/** Convenience function to get [FakeImageLoader.diskCache] as a [FakeDiskCache]. */
+val FakeImageLoader.fakeDiskCache: FakeDiskCache get() = diskCache as FakeDiskCache
+
+fun FakeImageLoader.enqueue(
```
yschimke, line 212: enqueue is an awkward name given the ImageLoader.enqueue also. one is a push, one is a pull.

coil-fakeimageloader/src/main/java/coil/FakeImageLoader.kt
```diff
@@ -0,0 +1,232 @@
...
+            }
+        )
+
+        @Synchronized
+        fun enqueue(data: Any, engine: Engine) = apply {
+            results.getOrPut(data) { LinkedList() }.offer(engine)
+        }
+
+        @Synchronized
+        fun default(default: Engine) = apply {
```
yschimke, line 126: This design with Engine, is beautiful and probably deserves to be called out in the PR.

Seems great for implementing a FakeImageLoader in a real app that uses local images for screenshot testing without predicting all requests up front.

yschimke:
LGTM

coil-fakeimageloader/src/main/java/coil/FakeImageLoader.kt
colinrtwhite: +1, pulled the name from `MockWebServer`, but there's definitely a better name for this case. Maybe `offer` or `push`?

coil-fakeimageloader/src/main/java/coil/FakeImageLoader.kt
colinrtwhite: Thanks! 🙏

coil-fakeimageloader/src/main/java/coil/FakeImageLoader.kt
```diff
@@ -0,0 +1,232 @@
...
+        fun enqueue(
+            data: Any,
+            drawable: Drawable,
+            dataSource: DataSource = DataSource.DISK,
+            memoryCacheKey: MemoryCache.Key? = null,
+            diskCacheKey: String? = null,
+        ) = enqueue(
+            data = data,
+            engine = { request ->
+                SuccessResult(
```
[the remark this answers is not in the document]
colinrtwhite: Good point, I'll create a convenience method for failed results too.

coil-fakeimageloader/src/main/java/coil/FakeImageLoader.kt
yschimke: I suck at naming things. Those work, or enqueueResponse.

yschimke:
Tangential: Is Coil.reset meant to be private?

coil-test-roborazzi/src/test/java/coil/test/RoborazziComposeTest.kt
```diff
@@ -0,0 +1,62 @@
...
+            override fun getIntrinsicHeight() = 100
+        }
+        val engine = FakeImageLoaderEngine.Builder()
+            .intercept(url, drawable)
+            .build()
+        val imageLoader = ImageLoader.Builder(composeTestRule.activity)
+            .components { add(engine) }
+            .build()
+
+        composeTestRule.setContent {
```
alexjlockwood, line 52: Does the unit test fail when the comparison image is different from the recording image?

In this comment it sounded like it was intentional (for now) for the unit tests to not fail: https://github.com/takahirom/roborazzi/issues/23#issuecomment-1510523504

Just figured I would ask because I was surprised by the behavior as well and was just wondering if you were aware...

Issue #1344, opened 2022-06-30
File name contain '#' load error.


XuQK:
**Describe the bug**
If parse a local path contain '#' like this: `/sdcard/fi#le.jpeg` to this method:
```kotlin
ImageView.load(
    data: Any?,
    imageLoader: ImageLoader = context.imageLoader,
    builder: ImageRequest.Builder.() -> Unit = {}
)
```
the path will be cut to `/sdcard/fi`, lost path after '#'.
This happend after `ComponentRegistry.map(data: Any, options: Options): Any` be called, after being processed by `FileUriMapper`.

I can call `Uri.encode()` to solve it every time use `ImageView.load()`, but it looks stupid. can you make it smater? Thanks.

**To Reproduce**
Parse any local path contain '#'.

**Version**
coil-2.1.0

PR #1466, opened 2022-09-18
Fix parsing paths containing # in FileUriMapper.

colinrtwhite:
Fixes: https://github.com/coil-kt/coil/issues/1344


coil-base/src/main/java/coil/map/FileUriMapper.kt
```diff
@@ -11,7 +11,8 @@ internal class FileUriMapper : Mapper<Uri, File> {
 
     override fun map(data: Uri, options: Options): File? {
         if (!isApplicable(data)) return null
-        return File(data.path!!)
+        val uri = if (data.scheme == null) data else data.buildUpon().scheme(null).build()
+        return File(uri.toString())
```
oguzozcan, line 15: Well that was a breaking change for us as we were previously sending local image's uri, now uri.path works, you should have document this

PR #1472, opened 2022-09-22
Update to Compose 1.3.

colinrtwhite:
Opening this to show a behaviour change in Compose `1.3.0-beta03`. Do not merge.


colinrtwhite:
This will be fixed in the next Compose UI release: https://issuetracker.google.com/issues/248100419#comment8

Issue #1482, opened 2022-09-30
`VideoFrameDecoder` doesn't support videos in subfolder of Android `/assets`


mhelder:
**Describe the bug**
When attempting to decode video frames from a file in an Android assets **subfolder**, the image load fails.

Stack trace:
```
java.io.FileNotFoundException: video.mp4
	at android.content.res.AssetManager.nativeOpenAssetFd(Native Method)
	at android.content.res.AssetManager.openFd(AssetManager.java:963)
	at coil.decode.VideoFrameDecoder.setDataSource(VideoFrameDecoder.kt:185)
	at coil.decode.VideoFrameDecoder.decode(VideoFrameDecoder.kt:39)
	at coil.intercept.EngineInterceptor.decode(EngineInterceptor.kt:199)
	at coil.intercept.EngineInterceptor.access$decode(EngineInterceptor.kt:41)
	at coil.intercept.EngineInterceptor$execute$executeResult$1.invokeSuspend(EngineInterceptor.kt:127)
	at coil.intercept.EngineInterceptor$execute$executeResult$1.invoke(Unknown Source:8)
	at coil.intercept.EngineInterceptor$execute$executeResult$1.invoke(Unknown Source:4)
	at kotlinx.coroutines.intrinsics.UndispatchedKt.startUndispatchedOrReturn(Undispatched.kt:89)
	at kotlinx.coroutines.BuildersKt__Builders_commonKt.withContext(Builders.common.kt:169)
	at kotlinx.coroutines.BuildersKt.withContext(Unknown Source:1)
	at coil.intercept.EngineInterceptor.execute(EngineInterceptor.kt:126)
	at coil.intercept.EngineInterceptor.access$execute(EngineInterceptor.kt:41)
	at coil.intercept.EngineInterceptor$intercept$2.invokeSuspend(EngineInterceptor.kt:75)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
	at kotlinx.coroutines.internal.LimitedDispatcher.run(LimitedDispatcher.kt:42)
	at kotlinx.coroutines.scheduling.TaskImpl.run(Tasks.kt:95)
	at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:570)
[...]

PR #1489, opened 2022-10-01
Fix decoding videos inside asset subfolders.

colinrtwhite:
Fixes: https://github.com/coil-kt/coil/issues/1482

### DIFF
--- a/coil-base/src/main/java/coil/decode/ImageSource.kt
+++ b/coil-base/src/main/java/coil/decode/ImageSource.kt
@@ -172,10 +172,17 @@ sealed class ImageSource : Closeable {
 }
 
 /**
- * Metadata containing the [fileName] of an Android asset.
+ * Metadata containing the [filePath] of an Android asset.
  */
 @ExperimentalCoilApi
-class AssetMetadata(val fileName: String) : ImageSource.Metadata()
+class AssetMetadata(val filePath: String) : ImageSource.Metadata() {
+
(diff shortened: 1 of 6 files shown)


coil-base/src/main/java/coil/decode/ImageSource.kt
```diff
@@ -172,10 +172,17 @@ sealed class ImageSource : Closeable {
...
  */
 @ExperimentalCoilApi
-class AssetMetadata(val fileName: String) : ImageSource.Metadata()
+class AssetMetadata(val filePath: String) : ImageSource.Metadata() {
+
+    @Deprecated(
+        message = "Migrate to filePath as it supports assets inside subfolders.",
+        level = DeprecationLevel.ERROR
+    )
+    val fileName: String get() = filePath.substringAfter('/')
```
mhelder, line 184: Shouldn't this be `substringAfterLast`? If `filePath` has multiple subfolders, then `substringAfter` will only 'skip' the first.

Practically speaking, I guess it may not matter too much. Anyone who was using `filePath` before had their file located in the root of the `assets` folder. And for those files `substringAfter` and `substringAfterLast` will yield the same result. 🤷‍♂️

colinrtwhite: Great point! Fixed here: https://github.com/coil-kt/coil/pull/1493

Issue #1423, opened 2022-09-01
Duplicate class androidx.lifecycle.ViewModel in coil 2.2.0


lUjek-Sh:
**Describe the bug**
After update coil to 2.2.0, i have error "Duplicate class androidx.lifecycle.ViewModel". But if i use coil 2.1.0 all is well.

**To Reproduce**
Update to coil version 2.2.0

**Logs/Screenshots**
```
Duplicate class androidx.lifecycle.ViewModelLazy found in modules lifecycle-viewmodel-2.5.0-runtime (androidx.lifecycle:lifecycle-viewmodel:2.5.0) and lifecycle-viewmodel-ktx-2.3.1-runtime (androidx.lifecycle:lifecycle-viewmodel-ktx:2.3.1)
Duplicate class androidx.lifecycle.ViewTreeViewModelKt found in modules lifecycle-viewmodel-2.5.0-runtime (androidx.lifecycle:lifecycle-viewmodel:2.5.0) and lifecycle-viewmodel-ktx-2.3.1-runtime (androidx.lifecycle:lifecycle-viewmodel-ktx:2.3.1)
```

**Version**
Kotlin 1.7.10
compileSdkVersion 32

**My Dependencies**
```
dependencies {
    implementation 'androidx.core:core-ktx:1.8.0'
    implementation 'androidx.exifinterface:exifinterface:1.3.3'
    implementation 'org.jetbrains:annotations:23.0.0'

    implementation "androidx.recyclerview:recyclerview:1.2.1"
    implementation 'androidx.appcompat:appcompat:1.4.2'
    implementation 'com.google.android.material:material:1.6.1'
    implementation "androidx.constraintlayout:constraintlayout:2.1.4"
    implementation 'com.github.faruktoptas:FancyShowCaseView:1.3.9'
    implementation "me.zhanghai.android.materialratingbar:library:1.4.0"

    implementation "com.google.firebase:firebase-analytics-ktx:21.1.0"
    implementation "com.google.firebase:firebase-crashlytics-ktx:18.2.12"
    implementation "com.google.firebase:firebase-messaging:23.0.7"

    implementation(platform("com.squareup.okhttp3:okhttp-bom:4.10.0"))
    implementation("com.squareup.okhttp3:okhttp")
    implementation("com.squareup.okhttp3:logging-interceptor")

    implementation "androidx.room:room-runtime:2.5.0-alpha02"
    implementation "androidx.room:room-ktx:2.5.0-alpha02"
[...]

PR #1507, opened 2022-10-12
Update Activity and Lifecycle libraries

sitamshrijal:


colinrtwhite:
Thanks, but I think we should hold off updating these libraries until Lifecycle 2.6.0 is stable as updating previously caused [issues](https://github.com/coil-kt/coil/issues/1423). Users can add a dependency on these libraries to manually update the version if they need the newer releases.

PR #1541, opened 2022-11-18
Update README.md

benju69:
Update compose sample for how we set the placeholder and error drawables


colinrtwhite:
Thanks! However, I think we actually want to encourage users to use the `placeholder` argument to `AsyncImage` since it allows using Compose's `Painter` as a placeholder.

Issue #1461, opened 2022-09-16
Avoid multiple parallel network requests for the same URL


Kazachenko1998:
I am developing an application where there can be many identical pictures on one screen (SVG from the server).
When I open the application for the first time, I see multiple parallel requests behind the same picture.

I think it's worth adding a pool of downloads, and if this URL is already being processed, do not make this request again, but wait for the first one to complete and use its result.

```
override fun onCreate(savedInstanceState: Bundle?) {
        CODE
        Coil.setImageLoader(
            ImageLoader
                .Builder(this)
                .components { add(SvgDecoder.Factory()) }
                .crossfade(true)
                .build()
        )
        CODE
}

```

```
AsyncImage(
    placeholder = rememberAsyncImagePainter(R.drawable.ic_plchldr),
    error = rememberAsyncImagePainter(R.drawable.ic_plchldr),
    model = URL,
    contentDescription = URL,
    contentScale = ContentScale.Crop,
    modifier = Modifier.fillMaxSize()
)

```
<img width="1464" alt="image" src="https://user-images.githubusercontent.com/22072948/190662362-1714d6be-4fee-4e13-b0d6-48d24a72cc86.png">

colinrtwhite:
I agree this would be nice to have. Ideally, `HttpUriFetcher` should reference count the number of responses waiting for its network request. We probably can't merge requests that only share the same URL - we'll need request headers to match as well (at least by default).

colinrtwhite:
Thinking about this more we should also provide a configurable "keep alive" duration for requests. That way the user can configure how long to wait before cancelling a network request if there's no more consumer for the result.

This would be useful for `RecyclerView`s and other cases where a request can be restarted multiple times in quick succession.

gargVader:
What is the status on this?

Airsaid:
How is the progress on this matter?

Airsaid:
> How is the progress on this matter?

I tried using an interceptor as a temporary workaround for the issue while waiting for an official fix:
```kotlin
/**
 * Merge the same requests to avoid multiple parallel network requests for the same URL.
 *
 * You can set the [mergeFilter] to determine which requests need to be merged,
 * the default is to merge the network requests.
 *
 * You can set the [keyGenerator] to generate the key for the request,
 * the default is to use the request data as the key.
 *
 * Fix the [issue](https://github.com/coil-kt/coil/issues/1461).
 *
 * @param mergeFilter A function that determines which requests need to be merged.
[...]

PR #1598, opened 2023-01-20
Merge image requests with the same memory cache key, maybe a solution for #1461 (If user use url to generate memory cache key)

asuka-mio:
Merge image requests with the same memory cache key, maybe a solution for #1461 (If user use url to generate memory cache key)


asuka-mio:
Sorry for that PR, I have achieved same effect by a Interceptor https://github.com/Ehviewer-Overhauled/Ehviewer/blob/1.8.x.x/app/src/main/java/com/hippo/ehviewer/coil/MergeInterceptor.kt

Issue #1513, opened 2022-10-20
Version 2.2.2 uri load bug


JYLEE1988:
**Describe the bug**
uri starting with `file:///` is not loaded.
A change occurred in the file creation method in version `2.2.2`, which is expected to be related to this.
#1466
```kotlin
override fun map(data: Uri, options: Options): File? {
        if (!isApplicable(data)) return null
        val uri = if (data.scheme == null) data else data.buildUpon().scheme(null).build()
        return File(uri.toString())
    }
```

In version `2.2.1`, the uri image loads well.

**To Reproduce**
Try to load a uri starting with `file:///`.
For example `file:///storage/emulated/0/android/data/{package-name}/cache/images/temp.jpg`

**Version**
2.2.2

colinrtwhite:
Thanks for the report - I'll take a look. As a work-around looks like removing the `file://` prefix should work?

JYLEE1988:
Yes, but in version 2.2.2, the File path is created like this
`///storage/emulated/0/android/data/{package-name}/cache/images/temp.jpg`

colinrtwhite:
@JYLEE1988 Where are you getting this file URI from? There's already a regression test for `file://` URIs, which is passing.

JYLEE1988:
@colinrtwhite 
The file name contained a `+`, which seems to be the cause.
When I remove the `+`, it works fine.

DjakaTechnology:
I encountered this problem and ended up downgrading the version to 2.2.1. Here's the URI:
`file:///storage/emulated/0/Download/Foo%20Bar/732234b189214a33a84d77609f60012e.jpg`

PR #1601, opened 2023-01-28
Fix file:// URI parsing

### LINKED ISSUE #1344: File name contain '#' load error.

TheKeeperOfPie:
A previous change to fix #1344 to handle # characters in literal no-scheme paths broke file:// scheme URI parsing by including the entire literal URI in the file path, which breaks encoded paths (like those with spaces) and URIs with non-path data (such as query parameters).

This partially reverts the prior fix, so that # is not supported in file:// URIs, requiring the user to either use a no-scheme URI or encode the path, as # is reserved in URIs for fragment declarations.

Fixes: #1513

Test: FileUriMapperTest#ignoresAfterPathCorrectly
Test: FileUriMapperTest#decodesEncodedPath


colinrtwhite (approved):
Thanks for fixing this and adding regression tests! It's too bad we can't support `#` in file URIs, but makes sense given the spec (as you mentioned).

Issue #1619, opened 2023-02-16
DiskCache.maxSize calculates incorrectly


Samuel-Unknown:
**Describe the bug**
I found a problem with `DiskCache`. It calculates `maxSize` incorrectly. 

**To Reproduce**
If you create a `DiskCache` by setting `maxSizePercent` and a directory that doesn't exist (e.g. the first run before putting anything in the cache), `maxSize` will be equal to `minimumMaxSizeBytes`.

You can't catch this with Robolectric tests, but you can with InstrumentedTests:

```
@Test
fun `checkInitSize`() {
val appContext = InstrumentationRegistry.getInstrumentation().targetContext
val minimumMaxSizeBytes = 100L
val maximumMaxSizeBytes = 200L
val maxSizePercent = 0.5

val diskCache = DiskCache.Builder().directory(appContext.filesDir.resolve("test_dir"))
   .maxSizePercent(maxSizePercent)
   .minimumMaxSizeBytes(minimumMaxSizeBytes)
   .maximumMaxSizeBytes(maximumMaxSizeBytes)
   .build()

    assertNotEquals(minimumMaxSizeBytes, diskCache.maxSize)
}
```

**Version**
2.2.2

**P.S. Workaround:**

```
DiskCache.Builder()
   .directory(context.filesDir.resolve(CACHE_DIR_NAME))
   .maxSizePercent(0.5)
   .also {
      if (context.filesDir.resolve(CACHE_DIR_NAME).exists().not()) {
         context.filesDir.resolve(CACHE_DIR_NAME).mkdir()
      }
   }
   .build()
```

colinrtwhite:
Fixed by https://github.com/coil-kt/coil/pull/1620.

PR #1620, opened 2023-02-16
Fix DiskCache maxSize calculation (#1619)

Samuel-Unknown:
Fix for #1619


colinrtwhite:
Awesome, thanks! Could you add that test in the issue to this PR? Feel free to create `DiskCacheAndroidTest` and add it there.

colinrtwhite (approved):
Thanks for the contribution!

PR #1675, opened 2023-03-23
Minimize compose state of AsyncImagePainter

zmunm:
# Context

Hi. I have a LazyList what state changes frequently, and I'm experiencing a leak where the forgotten state persists for a long time because the snapshots don't go away in time.
I know this isn't a coil problem, but it's hard to fix the compose leak right away, so I've reduced the state to a minimum so that it's okay if the AsyncImagePainter survives longer than expected.

# Change
- Changed the field holding the drawable in AsyncImagePainter to be non-state and indirectly notify to an int field.
- In clear , explicitly change the state to Empty to clear the painter and the drawable held by the state.

### DIFF
--- a/coil-compose-base/src/main/java/coil/compose/AsyncImagePainter.kt
+++ b/coil-compose-base/src/main/java/coil/compose/AsyncImagePainter.kt
@@ -164,7 +164,18 @@ class AsyncImagePainter internal constructor(
     private var rememberScope: CoroutineScope? = null
     private val drawSize = MutableStateFlow(Size.Zero)
 
-    private var painter: Painter? by mutableStateOf(null)
+    private var painterNotifier: Int by mutableStateOf(0)
+    private var painter: Painter? = null
+        get() {
+            painterNotifier
+            return _painter
+        }
+        set(value) {
+            if (field != value) {
+                field = value
+                painterNotifier = value.hashCode()
+            }
+        }
     private var alpha: Float by mutableStateOf(DefaultAlpha)
     private var colorFilter: ColorFilter? by mutableStateOf(null)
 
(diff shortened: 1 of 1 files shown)


colinrtwhite:
Thanks! Do you know how long Compose is keeping the `AsyncImagePainter` alive after it's no longer used? Also can you link me to the associated Compose bug ticket for the issue?

zmunm:
Thanks.

It persists until the activity is destroyed and triggers OutOfMemory.
Even within my team, it is difficult to provide the minimum reproduction code because the reproduction situation is different for each device. I'll try again soon.
For reference, it reproduces convincingly on my device and I use a Galaxy A50 (SM-A505N). I can provide a heap dump if needed.

I'm suspecting the same issue as [this](https://issuetracker.google.com/issues/240975572#comment23), but I'm not sure.

phucynwa:
Thank you. It is also the issue I got.

colinrtwhite:
@zmunm A heap dump would be great, thanks! I'm a little apprehensive to merge this change without a way to reproduce the issue and verify that it's an issue in Compose itself. Maybe a safer change would be to expose the `clear` method so `AsyncImagePainter`'s state can be cleared manually if desired?

zmunm:
@colinrtwhite 

https://github.com/zmunm/compose-leak-sample/

Finally, I succeeded in writing the minimal reproduction code. Hope this helps you too. @phucynwa 
I've included a profile sample as well, but hope it reproduces well on your device

Snapshot of SubActivity does not disappear when state change occurs in MainActivity.
As mentioned above, this is also correct that I wrote it wrong, but it seems to be a compose bug.
However, regular app developers find these things hard to spot. Most states don't take up that much memory. I believe this will make the coil safer.

I understand your concerns. I think it's good to apply this gradually rather than replacing the AsyncImagePainter right away.
But, it seems necessary to make the painter and state non-mutableState.
Beacuse I tried to change the mutableState to null, but the old state is still in memory because the recomposer is holding it.
So only putting a call to update(Empty) in clear might be an option.

zmunm:
Is it a possible approach to make AsyncImagePainter an interface or abstract and add another impl that applies this change as OptIn?

colinrtwhite:
@zmunm Thanks for the sample - I'll take a look! Unfortunately we can't convert `AsyncImagePainter` to an interface/abstract class as that would break binary compatibility.

colinrtwhite:
@zmunm Sorry for the delay on this. I was able to reproduce the issue with your sample (thanks again) and as a result I was also able to verify that upgrading to `androidx.compose.ui:ui:1.5.0-alpha03` fixes the issue. Since this is now fixed in the Compose runtime I'd prefer to avoid adding a work-around to Coil as it changes the external behaviour for other users.

zmunm:
I agree. I will close this issue. Thank you for thinking about it together. :smiley:

PR #1697, opened 2023-04-01
Build with Java 20 on CI.

colinrtwhite:


colinrtwhite:
Blocked by https://youtrack.jetbrains.com/issue/KT-57495/Add-JVM-target-bytecode-version-20.

PR #1698, opened 2023-04-01
[nits] Update CHANGELOG.md with correct info

ntsk:
It's a minor issue, but the version number of androidx.lifecycle is incorrect.
https://github.com/coil-kt/coil/blob/8694abaf0574f909caef533b73f1d818cdd96021/gradle/libs.versions.toml#L3


colinrtwhite (approved):

colinrtwhite:
Thanks!

Issue #1754, opened 2023-05-12
Coil tries to use Context#cacheDir even when a custom disk cache directory is set


saket:
**Describe the bug**
When Coil is used to load an image with disk caching disabled, it tries to access `Context#cacheDir` even when a custom cache directory is set. I'm using Coil's read image loader in a paparazzi test where `cacheDir` returns null, causing an NPE.

**To Reproduce**
```kotlin
val fakeCacheSystem = FakeFileSystem()
val loader = ImageLoader.Builder(context)
  .diskCache(
    DiskCache.Builder()
      .fileSystem(fakeCacheSystem)
      .directory(fakeCacheSystem.workingDirectory)
      .build()
  )
  .build()

loader.execute(
  ImageRequest.Builder(context)
    .data("foo")
    .memoryCachePolicy(CachePolicy.ENABLED)
    .diskCachePolicy(CachePolicy.DISABLED)
    .build()
)
```

Setting `diskCachePolicy` to `CachePolicy.ENABLED` fixes this problem.

**Stacktrace**
```kotlin
java.lang.NullPointerException
	at coil.util.-Utils.getSafeCacheDir(Utils.kt:174)
	at coil.decode.ImageSources.create(ImageSource.kt:68)
	at coil.fetch.HttpUriFetcher.toImageSource(HttpUriFetcher.kt:268)
	at coil.fetch.HttpUriFetcher.fetch(HttpUriFetcher.kt:97)
	at coil.fetch.HttpUriFetcher$fetch$1.invokeSuspend(HttpUriFetcher.kt)
	at _COROUTINE._BOUNDARY._(CoroutineDebugging.kt:46)
	at coil.intercept.EngineInterceptor.intercept(EngineInterceptor.kt:73)
	at coil.intercept.RealInterceptorChain.proceed(RealInterceptorChain.kt:25)
	at coil.RealImageLoader$executeMain$result$1.invokeSuspend(RealImageLoader.kt:191)
	at coil.RealImageLoader.executeMain(RealImageLoader.kt:182)
	at coil.RealImageLoader$execute$2$job$1.invokeSuspend(RealImageLoader.kt:133)
```

**Version**
2.3.0

colinrtwhite:
Good catch, thanks!

colinrtwhite:
[...]

PR #1758, opened 2023-05-14
Guard against Context.cacheDir being null.

colinrtwhite:
Related to https://github.com/coil-kt/coil/issues/1754 (does not resolve).

### DIFF
--- a/coil-base/src/main/java/coil/util/Utils.kt
+++ b/coil-base/src/main/java/coil/util/Utils.kt
@@ -49,6 +49,8 @@ import java.io.File
 import kotlinx.coroutines.Deferred
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import okhttp3.Headers
+import okhttp3.Response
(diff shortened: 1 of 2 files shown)


coil-base/src/main/java/coil/util/Utils.kt
```diff
@@ -171,7 +173,11 @@ internal fun <T> Deferred<T>.getCompletedOrNull(): T? {
 internal inline operator fun MemoryCache.get(key: MemoryCache.Key?) = key?.let(::get)
 
 /** https://github.com/coil-kt/coil/issues/675 */
```
saket, line 175: do you want to also move this comment?

```kotlin
internal val Context.safeCacheDir: File
    get() {
        val cacheDir = checkNotNull(cacheDir) { "cacheDir == null" }
        return cacheDir.apply { mkdirs() }  // https://github.com/coil-kt/coil/issues/675
    }
```

PR #1803, opened 2023-07-03
Fix incorrect link to OkHttpClient.newBuilder

aouerf:
It might also be worth linking to [some more detailed docs](https://square.github.io/okhttp/4.x/okhttp/okhttp3/-ok-http-client/#customize-your-client-with-newbuilder) instead.

### DIFF
--- a/docs/recipes.md
+++ b/docs/recipes.md
@@ -41,7 +41,7 @@ val imageLoader = ImageLoader.Builder(context)
 ```
 
 !!! Note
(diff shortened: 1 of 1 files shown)


colinrtwhite:
Thanks! I think you're right we should link [here](https://square.github.io/okhttp/4.x/okhttp/okhttp3/-ok-http-client/#customize-your-client-with-newbuilder) instead.

colinrtwhite (approved):

PR #1942, opened 2023-12-08
Merge 3.x into main.

colinrtwhite:


coil-network/src/commonMain/kotlin/coil3/fetch/NetworkFetcher.kt
```diff
@@ -0,0 +1,289 @@
...
+            file = data,
+            fileSystem = fileSystem,
+            diskCacheKey = diskCacheKey,
+            closeable = this,
+        )
+    }
+
+    private suspend fun ByteReadChannel.toImageSource(): ImageSource {
+        return ImageSource(
+            source = Buffer().apply { writeTo(this) },
```
VDostoyevskiy, line 254: This will squash the whole http body to art heap if no diskcache available
May cause OOM if image blob too large

coil-network/src/commonMain/kotlin/coil3/network/utils.common.kt
```diff
@@ -0,0 +1,73 @@
...
+/** Write a [ByteReadChannel] to [sink] using streaming. */
+internal suspend fun ByteReadChannel.writeTo(sink: BufferedSink) {
+    val buffer = ByteArray(OKIO_BUFFER_SIZE)
+
+    while (!isClosedForRead) {
+        val packet = readRemaining(buffer.size.toLong())
+        if (packet.isEmpty) break
+
+        // TODO: Figure out how to remove 'buffer' and read directly into 'sink'.
+        val bytesRead = packet.remaining.toInt()
```
VDostoyevskiy, line 49: On JVM/Android, we can use ByteReadChannel.copyTo
https://github.com/ktorio/ktor/blob/main/ktor-io/jvm/src/io/ktor/utils/io/jvm/nio/Writing.kt
But I'm not sure if this API is subject to change

coil-network/src/commonMain/kotlin/coil3/fetch/NetworkFetcher.kt
colinrtwhite: Yep, this is only used as a fallback if we can't write to disk. Skiko also requires an image to be fully loaded into memory before decoding: https://github.com/JetBrains/skiko/issues/741

coil-network/src/commonMain/kotlin/coil3/network/utils.common.kt
colinrtwhite: Nice! Updated

colinrtwhite: I wonder if it's possible to copy that extension's implementation and modify it to write to a `BufferedSink` directly. I took a quick look, but didn't look too closely

VDostoyevskiy: > Nice! Updated

Aha, I suggest using FileChannel not BufferedSink
Since ByteReadChannel use DirectByteBuffer as inner buffer for now(current behaviour, not sure after ktor migrated to kotlinx-io)

VDostoyevskiy: > I wonder if it's possible to copy that extension's implementation and modify it to write to a `BufferedSink` directly. I took a quick look, but didn't look too closely

Not possible, it rely on ByteReadChannel's platform implementation

coil-network/src/commonMain/kotlin/coil3/fetch/NetworkFetcher.kt
VDostoyevskiy: It seems like this is the only way to do it.
Impossible to wrap a ByteReadChannel with okio Source(or InputStream), unless we pay for a thread blocking on it, as okio is sync and ktor is async,

coil-network/src/commonMain/kotlin/coil3/network/utils.common.kt
VDostoyevskiy: Another things to clarify, the ByteReadChannel. not always provide DirectByteBuffer, various by engine.

https://github.com/FooIbar/EhViewer/blob/main/app/src/oss/kotlin/com/hippo/ehviewer/ktor/CronetEngine.kt
Here I wrapped Android 14 HttpEngine(actually it's cronet)as a ktor engine, it always provide DirectByteBuffer as Cronet use DirectByteBuffer.
Anyway, engine provides HeapByteBuffer, we won't lose performance using FileChannel, engine provides DirectByteBuffer, we gain performance for zero-copying

colinrtwhite: Gotcha - reverted the previous commit. Feel free to send a PR that uses `FileChannel`!

VDostoyevskiy: > Gotcha - reverted the previous commit. Feel free to send a PR that uses `FileChannel`!

With nonJvm platforms, we can use UnsafeCursor(https://square.github.io/okio/3.x/okio/okio/okio/-buffer/-unsafe-cursor/index.html) though, I don't suggest to do this

coil-network/src/commonMain/kotlin/coil3/fetch/NetworkFetcher.kt
```diff
@@ -0,0 +1,287 @@
...
+
+    private suspend fun <T> executeNetworkRequest(
+        request: HttpRequestBuilder,
+        block: suspend (HttpResponse) -> T,
+    ): T {
+        // Prevent executing requests on the main thread that could block due to a
+        // networking operation.
+        if (options.networkCachePolicy.readEnabled) {
+            assertNotOnMainThread()
+        }
```
VDostoyevskiy, line 207: I don't think this is still needed, ktor is not okhttp, it's fully async, no way to do network on MainThread unless engine author is a fool who use Dispatcher.Main to do socket select..
But we still need take care of disk cache writing to ensure it not happen on MainThread

colinrtwhite: @VDostoyevskiy Some users previously requested the ability to force loading the image from the disk cache on the main thread. It's definitely not recommended, but can be supported by requiring the user to set the `networkCachePolicy = CachePolicy.DISABLED` if they want to use the main thread.

Developer354 (approved):

PR #1962, opened 2023-12-19
Set the stability of coil-core classes for the Compose compiler.

colinrtwhite:
This can't be used until Multiplatform Compose compiler [version](https://developer.android.com/jetpack/androidx/releases/compose-compiler#1.5.5) `1.5.5` is released.


rustamsmax:
["1.5.7"](https://mvnrepository.com/artifact/org.jetbrains.compose.compiler/compiler/1.5.7) released

PR #1965, opened 2023-12-21
Add multiplatform support to coil-svg

MohamedRejeb:
- [x] Migrate `coil-svg` module to multiplatform
- [x] Add `SvgDecoder` for non android platforms based on skia svg
- [x] Add a public API to create an `SvgDecoder` from `commonMain`
- [x] Add tests to jvm target


coil-svg/src/commonMain/kotlin/coil3/decode/SvgDecoder.kt
```diff
@@ -0,0 +1,11 @@
+package coil3.decode
+
+object SvgDecoder {
```
colinrtwhite, line 3: Let's do an expect/actual class instead:

```kotlin
expect class SvgDecoder : Decoder {
    class Factory : Decoder.Factory
}
```

colinrtwhite, line 3: Let's do an expect/actual class instead of factory functions:

```kotlin
expect class SvgDecoder @JvmOverloads constructor(
    source: ImageSource,
    options: Options,
    useViewBoundsAsIntrinsicSize: Boolean = true,
) : Decoder {
    class Factory : Decoder.Factory
}
```

MohamedRejeb: Done

colinrtwhite:
Btw you can run the CI checks locally using `./test.sh`. Unfortunately I can't set it to run on each push

MohamedRejeb:
It should run successfully now.

colinrtwhite:
Great! This looks pretty good. I think we should be able to merge it once the tests are updated to run on Android and the JVM

coil-svg/src/androidMain/kotlin/coil3/decode/SvgDecoder.android.kt
```diff
@@ -20,16 +20,17 @@ import kotlin.math.roundToInt
 import kotlinx.coroutines.runInterruptible
 
 /**
- * A [Decoder] that uses [AndroidSVG](https://bigbadaboom.github.io/androidsvg/) to decode SVG
+ * A [Decoder] that uses [AndroidSVG](https://bigbadaboom.github.io/androidsvg/)
+ * and [SVGDOM](https://api.skia.org/classSkSVGDOM.html/) to decode SVG
```
colinrtwhite, line 24: Seems like this should be the docs for the `expect` declaration

coil-svg/src/commonMain/kotlin/coil3/decode/SvgDecoder.kt
```diff
@@ -0,0 +1,24 @@
+package coil3.decode
+
+import coil3.request.Options
+
+/**
+ * A [Decoder] that uses [AndroidSVG](https://bigbadaboom.github.io/androidsvg/) and  to decode SVG
```
colinrtwhite, line 6: This should be swapped with the Android implementation docs

coil-svg/src/jvmTest/kotlin/coil3/decode/SvgDecoderTest.jvm.kt
```diff
@@ -0,0 +1,149 @@
...
+import okio.FileSystem
+import okio.Path.Companion.toPath
+import okio.buffer
+
+class SvgDecoderTestJvm {
+
+    private val decoderFactory = SvgDecoder.Factory()
+
+    @Test
+    fun handlesSvgMimeType() {
```
colinrtwhite, line 31: Is it possible to implement these tests in a common way? If you can't put a common test class in `jvmCommonTest`, you could create `AbstractSvgDecoderTest` in `jvmCommonTest` which defines all the tests then have `SvgDecoderTestAndroid` and `SvgDecoderTestJvm` extend that class.

internal/test-utils/src/jvmMain/kotlin/coil3/test/utils/bitmaps.jvm.kt
```diff
@@ -0,0 +1,76 @@
+package coil3.test.utils
+
+import coil3.size.Size
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.async
+import kotlinx.coroutines.runBlocking
+import org.jetbrains.skia.Bitmap
+
+val Bitmap.size: Size
```
colinrtwhite, line 9: Thanks for converting these! Could we create a common `Bitmap` type (only for our tests)? Looks like something like this should be enough:

```kotlin
expect class Bitmap {
    val width: Int
    val height: Int

    fun getPixels(): ByteArray
}
```

This way we can put the rest of these functions in `commonMain`, which will be useful for migrating more of Coil's tests to `commonMain`.

coil-svg/src/androidMain/kotlin/coil3/decode/SvgDecoder.android.kt
MohamedRejeb: Done

coil-svg/src/commonMain/kotlin/coil3/decode/SvgDecoder.kt
MohamedRejeb: Done

coil-svg/src/jvmTest/kotlin/coil3/decode/SvgDecoderTest.jvm.kt
MohamedRejeb: I was trying to do this, but I can't access `jvmCommonTest` or even `commonTest` from `androidInstrumentedTest`.

MohamedRejeb:
I created `AbstractSvgDecoderTest` in the internal test utils module, and now the svg decoder test is shared between jvm and android. Also, I created `CoilBitmap` for testing and the bitmap comparison functions are shared between jvm and android. I didn't name it Bitmap to avoid confusion.
The only thing that prevents us from moving these tests to common is the ability to read resources from common test.
Now some functions are still duplicated in android for Bitmap and in common for `CoilBitmap` just to keep the old android tests working, But after migrating those tests, these functions can be removed from android.

MohamedRejeb:
A little mistake in the dependencies caused this failure. It should be running successfully now.

coil-svg/src/jvmTest/kotlin/coil3/decode/SvgDecoderTest.jvm.kt
```diff
@@ -0,0 +1,35 @@
+package coil3.decode
+
+import coil3.test.utils.AbstractSvgDecoderTest
+import kotlin.test.Test
+
+class SvgDecoderTestJvm : AbstractSvgDecoderTest(SvgDecoder.Factory()) {
+    @Test
+    override fun handlesSvgMimeType() =
```
colinrtwhite, line 8: Same for here.

internal/test-utils/src/jvmCommonMain/kotlin/coil3/test/utils/AbstractSvgDecoderTest.kt
```diff
@@ -1,91 +1,69 @@
...
-import okio.Buffer
 import okio.BufferedSource
 import okio.FileSystem
+import okio.Path.Companion.toPath
 import okio.buffer
-import okio.source
-import org.junit.Test
 
-class SvgDecoderTest {
+abstract class AbstractSvgDecoderTest(
```
colinrtwhite, line 25: Can this class live in `coil-svg/jvmCommonTest` instead of moving it to `test-utils`?

internal/test-utils/src/androidMain/kotlin/coil3/test/utils/bitmaps.android.kt
```diff
@@ -100,7 +129,7 @@ fun Bitmap.isSimilarTo(
  */
 fun Bitmap.assertIsSimilarTo(
     expected: Bitmap,
-    threshold: Double = 0.99
+    threshold: Double = 0.987
```
colinrtwhite, line 132: Let's use `0.98`

internal/test-utils/src/jvmCommonMain/kotlin/coil3/test/utils/AbstractSvgDecoderTest.kt
MohamedRejeb: When it was in `jvmCommonTest` it was not visible to android instrumented test.

internal/test-utils/src/androidMain/kotlin/coil3/test/utils/bitmaps.android.kt
MohamedRejeb: I changed it back to 0.99 in android and 0.98 in common.

coil-svg/src/androidInstrumentedTest/kotlin/coil3/decode/SvgDecoderTest.android.kt
```diff
@@ -0,0 +1,35 @@
+package coil3.decode
+
+import coil3.test.utils.AbstractSvgDecoderTest
+import kotlin.test.Test
+
+class SvgDecoderTestAndroid : AbstractSvgDecoderTest(SvgDecoder.Factory()) {
+    @Test
+    override fun handlesSvgMimeType() =
```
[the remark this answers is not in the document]
MohamedRejeb: Done

colinrtwhite (approved):
Thanks!

KapilYadav-dev:
@MohamedRejeb awesome to see your contribution in coil also < 3

PR #1974, opened 2023-12-28
Optimize ByteBufferFetcher, Prototype

### LINKED ISSUE #1971: [Coil 3] Introduce ByteBufferMetadata
Now that we prefer ImageDecoderDecoder for API above 28
https://github.com/coil-kt/coil/pull/1969
ImageDecoder can efficiently decoding a ByteBuffer
After https://github.com/coil-kt/coil/issues/1970 we lazily create a BufferedSource (Write the ByteBuffer totally to art heap)
When ImageDecoder is available, we no longer pay for the overhead, as ByteBuffer maybe a DirectByteBuffer allocated in native heap or file mapping
See https://github.com/coil-kt/coil/issues/1943#issuecomment-1848350924

VDostoyevskiy:
Close #1971 

We will do following things in this CL

* Make ByteBuffer lazily extracted to okio Buffer, to avoid heap thrashing (Actually we should let okio Buffer take these byte[]'s ownership if it is a HeapByteBuffer, but seems it is not possible)
* Introduce ByteBufferMetadata , to allow ImageDecoder take it directly

### DIFF
--- a/coil-core/src/androidMain/kotlin/coil3/decode/StaticImageDecoderDecoder.kt
+++ b/coil-core/src/androidMain/kotlin/coil3/decode/StaticImageDecoderDecoder.kt
@@ -7,6 +7,7 @@ import androidx.core.util.component1
 import androidx.core.util.component2
 import coil3.ImageLoader
 import coil3.asCoilImage
+import coil3.fetch.ByteBufferFetcher.ByteBufferMetadata
 import coil3.fetch.SourceFetchResult
 import coil3.request.Options
 import coil3.request.allowRgb565
@@ -125,6 +126,9 @@ private fun ImageSource.fastImageDecoderSourceOrNull(options: Options): ImageDec
     }
     if (metadata is ResourceMetadata && metadata.packageName == options.context.packageName) {
         return ImageDecoder.createSource(options.context.resources, metadata.resId)
+    }
+    if (metadata is ByteBufferMetadata) {
+        return ImageDecoder.createSource(metadata.byteBuffer)
     }
 
     return null
--- a/coil-core/src/androidMain/kotlin/coil3/decode/StaticImageDecoderDecoder.kt
+++ b/coil-core/src/androidMain/kotlin/coil3/decode/StaticImageDecoderDecoder.kt
@@ -7,7 +7,6 @@ import androidx.core.util.component1
 import androidx.core.util.component2
 import coil3.ImageLoader
 import coil3.asCoilImage
-import coil3.fetch.ByteBufferFetcher.ByteBufferMetadata
 import coil3.fetch.SourceFetchResult
 import coil3.request.Options
 import coil3.request.allowRgb565
--- a/coil-core/src/jvmCommonMain/kotlin/coil3/decode/ImageSource.kt
+++ b/coil-core/src/jvmCommonMain/kotlin/coil3/decode/ImageSource.kt
@@ -1,0 +1,12 @@
+package coil3.decode
+
+import coil3.annotation.ExperimentalCoilApi
+import java.nio.ByteBuffer
+
+/**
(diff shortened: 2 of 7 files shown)


VDostoyevskiy:
@colinrtwhite 
Done, PTAL
Didn't let AnimatedImageDecoderDecoder use ByteBufferMetadata in case one pass a HeapByteBuffer

coil-core/src/androidMain/kotlin/coil3/decode/StaticImageDecoderDecoder.kt
```diff
@@ -126,6 +127,9 @@ private fun ImageSource.fastImageDecoderSourceOrNull(options: Options): ImageDec
     if (metadata is ResourceMetadata && metadata.packageName == options.context.packageName) {
         return ImageDecoder.createSource(options.context.resources, metadata.resId)
     }
+    if (metadata is ByteBufferMetadata) {
```
colinrtwhite, line 130: I think we can add this same code to `AnimatedImageDecoderDecoder`.

coil-core/src/jvmCommonMain/kotlin/coil3/fetch/ByteBufferFetcher.kt
```diff
@@ -36,4 +37,23 @@ internal class ByteBufferFetcher(
             return ByteBufferFetcher(data, options)
         }
     }
+
+    class ByteBufferMetadata(val byteBuffer: ByteBuffer) : ImageSource.Metadata()
```
colinrtwhite, line 41: Let's move this to the same file as `AssetMetadata` and mark it with `@ExperimentalCoilApi`.

VDostoyevskiy: AssetMetadata is in androidMain, but ByteBufferFetcher is in jvmCommonMain

VDostoyevskiy: Done

coil-core/src/androidMain/kotlin/coil3/decode/StaticImageDecoderDecoder.kt
VDostoyevskiy: Done

colinrtwhite (approved):

coil-core/src/jvmCommonMain/kotlin/coil3/decode/ImageSource.kt
```diff
@@ -0,0 +1,12 @@
+package coil3.decode
+
+import coil3.annotation.ExperimentalCoilApi
+import java.nio.ByteBuffer
+
+/**
+ * Metadata containing the [bytebuffer] of a ByteBuffer, maybe direct
+ */
+@ExperimentalCoilApi
+class ByteBufferMetadata(
```
colinrtwhite, line 10: I think this file might have to be renamed to `ImageSource.jvm.kt` to avoid class name conflicts.

VDostoyevskiy: Well, it doesn't have top-level functions/field, so will not have clash.

VDostoyevskiy:
@colinrtwhite 
Aha, Btw ByteArrayFetcher could be optimized with the same method(lazy extraction)
https://github.com/coil-kt/coil/blob/98541dbaba103f1e1aafa825ad3830e6287716e7/coil-core/src/commonMain/kotlin/coil3/fetch/ByteArrayFetcher.kt#L17
But I decided not to optimize it as I think one use ByteArray to decode would not care performance

PR #1975, opened 2023-12-29
fixup! Optimize with FileChannel on JVM/Android 

VDostoyevskiy:
Sorry for that

### DIFF
--- a/coil-network/src/jvmCommonMain/kotlin/coil3/network/utils.jvm.kt
+++ b/coil-network/src/jvmCommonMain/kotlin/coil3/network/utils.jvm.kt
@@ -14,8 +14,9 @@ internal actual suspend fun ByteReadChannel.writeTo(sink: BufferedSink) {
 internal actual suspend fun ByteReadChannel.writeTo(fileSystem: FileSystem, path: Path) {
     if (fileSystem === FileSystem.SYSTEM) {
         // Fast path: normal jvm File, write to FileChannel directly.
-        RandomAccessFile(path.toFile(), "rw").use {
-            copyTo(it.channel)
+        RandomAccessFile(path.toFile(), "rw").use { file ->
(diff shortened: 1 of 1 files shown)


coil-network/src/jvmCommonMain/kotlin/coil3/network/utils.jvm.kt
```diff
@@ -14,8 +14,9 @@ internal actual suspend fun ByteReadChannel.writeTo(sink: BufferedSink) {
 internal actual suspend fun ByteReadChannel.writeTo(fileSystem: FileSystem, path: Path) {
     if (fileSystem === FileSystem.SYSTEM) {
         // Fast path: normal jvm File, write to FileChannel directly.
-        RandomAccessFile(path.toFile(), "rw").use {
-            copyTo(it.channel)
+        RandomAccessFile(path.toFile(), "rw").use { file ->
+            val copied = copyTo(file.channel)
+            file.setLength(copied) // truncate tail that could remain from the previously written data
```
colinrtwhite, line 19: What happens if we throw during `copyTo`? Should we set the length to 0?

VDostoyevskiy: DiskCache should abort edit and delete that entry
https://github.com/coil-kt/coil/blob/8389ba8b95a41b49d2d347aaf6ea5abb7f3eff90/coil-network/src/commonMain/kotlin/coil3/fetch/NetworkFetcher.kt#L166

VDostoyevskiy:
Sorry for that, after read DiskLruCache implementation, seems it always provides us a new file to write, so this should not be a problem

colinrtwhite:
👍🏻 No problem, yep I agree `DiskLruCache` will always provide a fresh file and aborts the edit if an exception occurs.

PR #1977, opened 2023-12-31
Fix link to OkHttp Proguard file

teobaranga:
OkHttp recently updated the location of their Proguard file so this should now point to the correct location.


colinrtwhite (approved):
Thanks!

PR #1982, opened 2024-01-01
Improve bitmapFactoryMaxParallelism docs.

colinrtwhite:


coil-core/src/androidMain/kotlin/coil3/RealImageLoader.kt
```diff
@@ -83,20 +83,18 @@ internal actual fun ComponentRegistry.Builder.addAndroidComponents(
...
-            // ImageDecoder cannot properly decode a hardware bitmap on API 28
+            // Require API 29 for ImageDecoder support as API 28 has framework bugs:
             // https://github.com/element-hq/element-android/pull/7184
-            if (SDK_INT >= 29) {
-                add(
-                    StaticImageDecoderDecoder.Factory(
-                        maxParallelism = options.bitmapFactoryMaxParallelism,
-                    )
+            val decoderFactory = if (SDK_INT >= 29) {
+                StaticImageDecoderDecoder.Factory(
```
VDostoyevskiy, line 89: StaticImageDecoderDecoder cannot decode all source, need BitmapFactoryDecoder as fallback

VDostoyevskiy: If you want StaticImageDecoder process all source anyway, you can add back
https://github.com/coil-kt/coil/blob/98202efd5f7060784215ef004cf722a31a2f79a8/coil-gif/src/main/java/coil3/decode/AnimatedImageDecoderDecoder.kt#L130

But it means we will squash the stream totally to art heap, not recommended

colinrtwhite: Hmm I thought we could use `ImageDecoder` as a full replacement for `BitmapFactory`...

Using different decoders based on different types of data opens us up to potential inconsistencies between the two decoders and adds maintenance burden.

VDostoyevskiy: Maybe... we can?
Fallback to a memfd and map it to a DirectByteBuffer, then it does not affect art heap.

colinrtwhite: Let's leave it as is for now

PR #1986, opened 2024-01-01
Use DirectByteBuffer as AnimatedImageDecoderDecoder fallback source

### LINKED ISSUE #1980: [Coil 3] Unify AnimatedImageDecoderDecoder fallback source to DirectByteBuffer
Following up https://github.com/coil-kt/coil/pull/1969
With StaticImageDecoderDecoder, if there is no appropriate source, we can fallback to BitmapFactory.
But with AnimatedImageDecoderDecoder, we must decode it anyway, considering AnimatedImageDecoderDecoder decodes **animated** images
It means, "source" **cannot** be released until the image recycled, ImageDecoder will continuously use that source,
Now that the source will exist on heap for a long time, a DirectByteBuffer must be more appropriate as its content **will not be copied** during GC
@colinrtwhite 

And about https://github.com/coil-kt/coil/pull/111/ , have you tested **Direct**ByteBuffer? I guess this is something decode during GC heap moving, though I haven't dig into framework source code.

Addition: https://github.com/DerpFest-AOSP/frameworks_base/commit/a7ec12f7efc032211e54aa1ce28c698413f02b8d
Seems this is a HeapByteBuffer only thing

VDostoyevskiy:
Close #1980

### DIFF
--- a/coil-gif/src/main/java/coil3/decode/AnimatedImageDecoderDecoder.kt
+++ b/coil-gif/src/main/java/coil3/decode/AnimatedImageDecoderDecoder.kt
@@ -191,10 +191,11 @@ class AnimatedImageDecoderDecoder @JvmOverloads constructor(
 }
 
 internal fun BufferedSource.squashToDirectByteBuffer(): ByteBuffer {
-    // TODO: How can we know BufferedSource's length, avoid allocating that bytearray?
-    val bytes = readByteArray()
-    return ByteBuffer.allocateDirect(bytes.size).apply {
-        put(bytes)
+    // Squash bytes to BufferedSource inner buffer then we know total byteCount
+    request(Long.MAX_VALUE)
+    return ByteBuffer.allocateDirect(buffer.size.toInt()).apply {
+        // Squash to DirectByteBuffer
+        buffer.read(this)
         flip()
     }
 }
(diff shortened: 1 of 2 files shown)


coil-gif/src/main/java/coil3/decode/AnimatedImageDecoderDecoder.kt
```diff
@@ -195,5 +195,14 @@ class AnimatedImageDecoderDecoder @JvmOverloads constructor(
     }
 }
 
+internal fun BufferedSource.squashToDirectByteBuffer(): ByteBuffer {
+    // TODO: How can we know BufferedSource's length, avoid allocating that bytearray?
+    val bytes = readByteArray()
```
VDostoyevskiy, line 200: @colinrtwhite 
Is it acceptable to add a length field for [ImageSource] and let fetcher initialize it
To avoid allocating this ByteArray, we use https://docs.oracle.com/javase/8/docs/api/java/nio/channels/ReadableByteChannel.html#read-java.nio.ByteBuffer-
write to DirectByteBuffer directly?

colinrtwhite: We can't add a `length` field to `ImageSource` as it's not possible to support consistently. Adding it would mean supporting it inconsistently across our fetchers. I'd also want to avoid adding `length` just to support this use case.

Unfortunately this might prevent us from adding this optimization (since we have to allocate the byte array anyways).

VDostoyevskiy: No matter, there is another way to achieve this

VDostoyevskiy:
@colinrtwhite 
Rebased, PTAL

colinrtwhite (approved):

PR #1990, opened 2024-01-02
Share AssetFileDescriptor with ImageDecoder

### LINKED ISSUE #1970: [Coil 3] Lazy allocated BufferedSource for ImageSource
Now that we prefer ImageDecoderDecoder for API above 28
https://github.com/coil-kt/coil/pull/1969
ImageDecoder prefer a (File, Uri) not inputstream, we have no reason to preallocate a BufferedSource(this opens a native file descriptor), save this overhead since ImageDecoder will not use that fd
See https://github.com/VDostoyevskiy/coil/blob/068e566ce04e9e1d77e08b4c80c386870d876860/coil-core/src/androidMain/kotlin/coil3/fetch/AssetUriFetcher.kt#L25
https://github.com/VDostoyevskiy/coil/blob/068e566ce04e9e1d77e08b4c80c386870d876860/coil-core/src/androidMain/kotlin/coil3/fetch/ContentUriFetcher.kt#L33

VDostoyevskiy:
Close #1970

### DIFF
--- a/coil-core/src/androidMain/kotlin/coil3/decode/ImageSource.kt
+++ b/coil-core/src/androidMain/kotlin/coil3/decode/ImageSource.kt
@@ -1,5 +1,6 @@
 package coil3.decode
 
+import android.content.res.AssetFileDescriptor
 import androidx.annotation.DrawableRes
 import coil3.Uri
 
@@ -15,6 +16,7 @@ class AssetMetadata(
  */
 class ContentMetadata(
     val uri: Uri,
+    val afd: AssetFileDescriptor
 ) : ImageSource.Metadata()
 
 /**
--- a/coil-core/src/androidMain/kotlin/coil3/decode/ImageSource.kt
+++ b/coil-core/src/androidMain/kotlin/coil3/decode/ImageSource.kt
@@ -16,7 +16,7 @@ class AssetMetadata(
  */
 class ContentMetadata(
     val uri: Uri,
-    val afd: AssetFileDescriptor
+    val assetFileDescriptor: AssetFileDescriptor
 ) : ImageSource.Metadata()
 
 /**
--- a/coil-core/src/androidMain/kotlin/coil3/fetch/ContentUriFetcher.kt
+++ b/coil-core/src/androidMain/kotlin/coil3/fetch/ContentUriFetcher.kt
@@ -30,30 +30,28 @@ internal class ContentUriFetcher(
     override suspend fun fetch(): FetchResult {
         val androidUri = data.toAndroidUri()
         val contentResolver = options.context.contentResolver
-        val inputStream = if (isContactPhotoUri(data)) {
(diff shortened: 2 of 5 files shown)


coil-core/src/androidMain/kotlin/coil3/decode/ImageSource.kt
```diff
@@ -15,6 +16,7 @@ class AssetMetadata(
  */
 class ContentMetadata(
     val uri: Uri,
+    val afd: AssetFileDescriptor
```
colinrtwhite, line 19:
```suggestion
    val assetFileDescriptor: AssetFileDescriptor,
```

colinrtwhite: Should this be optional? Are there situations were an `AssetFileDescriptor` is not available for a URI?

coil-core/src/androidMain/kotlin/coil3/fetch/ContentUriFetcher.kt
```diff
@@ -30,30 +30,28 @@ internal class ContentUriFetcher(
     override suspend fun fetch(): FetchResult {
         val androidUri = data.toAndroidUri()
         val contentResolver = options.context.contentResolver
-        val inputStream = if (isContactPhotoUri(data)) {
+        val afd = if (isContactPhotoUri(data)) {
```
colinrtwhite, line 33: Separate to this PR, but will `ImageDecoder.createSource(options.context.contentResolver, metadata.uri.toAndroidUri())` correctly handle this special case for a contacts image URI?

One of the issues with passing this data to `StaticImageDecoderDecoder` is we're coupling the source with the decoder.

coil-core/src/androidMain/kotlin/coil3/decode/ImageSource.kt
VDostoyevskiy: Nope, If we can open a InputStream for a ContentUri, there must have a afd as that I/OStream use afd internally

VDostoyevskiy: Done

coil-core/src/androidMain/kotlin/coil3/fetch/ContentUriFetcher.kt
VDostoyevskiy: > Separate to this PR, but will `ImageDecoder.createSource(options.context.contentResolver, metadata.uri.toAndroidUri())` correctly handle this special case for a contacts image URI?
> 
> One of the issues with passing this data to `StaticImageDecoderDecoder` is we're coupling the source with the decoder.

Haven't test it, but now ImageDecoder use a afd that InputStream use for this case so it doesn't matter

colinrtwhite (approved):

PR #2004, opened 2024-01-03
Don't share afd in AnimatedImageDecoderDecoder

VDostoyevskiy:
ImageDecoder require its source still available if returned drawable is Animatable
But we will close ImageSource after decode, associated afd will be closed as well
https://developer.android.com/reference/android/graphics/ImageDecoder#createSource(java.nio.ByteBuffer)
My bad


VDostoyevskiy:
Don't mind me
This limit is only for ByteBuffer.
ImageDecoder will internally duplicate that fd
https://github.com/DerpFest-AOSP/frameworks_base/blob/0f805759a96b2c83bfc10aefcf966cce9eed6af4/libs/hwui/jni/ImageDecoder.cpp#L176

PR #2013, opened 2024-01-04
Update Image API on non android targets to support drawing on Canvas

MohamedRejeb:
The design is not final.
I tried to completely replace `Bitmap` with `Canvas.onDraw()`, but for `SkiaImageDecoder` we will need to do some extra processing by converting the `Bitmap` to `Image` again so we can draw it to `Canvas`.
That's why I decided to create `CoilPainter` sealed interface to support both options.
- `CoilPainter.BitmapPainter`: a painter based on `Bitmap` that will be converted to Compose `BitmapPainter`.
- `CoilPainter.VectorPainter`: a painter based on `Canvas.onDraw()` that it's going to be called from Compose `Painter` `DrawScope.onDraw()` using the native canvas.


coil-core/src/nonAndroidMain/kotlin/coil3/Image.kt
```diff
@@ -44,7 +68,24 @@ class BitmapImage internal constructor(
...
+
+@ExperimentalCoilApi
+@Data
+class SvgImage internal constructor(
+    val svg: SVGDOM,
+    override val width: Int,
+    override val height: Int,
+): Image {
+    override val size: Long
+        get() = 4L * width * height
```
MohamedRejeb, line 84: The size of `SVGDOM` should be way smaller, but I didn't find a way to calculate it. I was thinking about setting the same size of the data coming from the `ImageSource`

coil-core/src/nonAndroidMain/kotlin/coil3/Image.kt
```diff
@@ -4,21 +4,45 @@ import coil3.annotation.Data
...
+import org.jetbrains.skia.svg.SVGDOM
 
 @ExperimentalCoilApi
 @JvmOverloads
 fun Bitmap.asCoilImage(
     shareable: Boolean = isImmutable,
 ): Image = BitmapImage(this, shareable)
 
+@ExperimentalCoilApi
+fun SVGDOM.asCoilImage(
```
colinrtwhite, line 17: We should keep this private. Maintaining this public API going forward could be tough.

coil-core/src/nonAndroidMain/kotlin/coil3/Image.kt
colinrtwhite: This should live in `coil-svg` as we can't have any SVG classes in the main artifact.

coil-core/src/nonAndroidMain/kotlin/coil3/CoilPainter.kt
```diff
@@ -0,0 +1,18 @@
+package coil3
+
+import coil3.annotation.ExperimentalCoilApi
+import org.jetbrains.skia.Bitmap
+import org.jetbrains.skia.Canvas
+
+@ExperimentalCoilApi
+sealed interface CoilPainter {
```
colinrtwhite, line 8: I think this API isn't the right abstraction as it locks us into a fixed set of supported use cases via the `sealed interface`. For example, how would we add future support for rendering an animated gif without adding another sealed type?

It also feels like we're duplicating functionality from the `Image` class here. I think we should remove this type and add `fun Canvas.onDraw()` to the common `Image` type.

MohamedRejeb: But then how to decide how the `Image` is going to be converted to a Compose `Painter`.

colinrtwhite: I think ideally we can create an `ImagePainter` which wraps a generic `Image` and draws it

coil-core/src/nonAndroidMain/kotlin/coil3/Image.kt
MohamedRejeb: Done

coil-core/src/nonAndroidMain/kotlin/coil3/CoilPainter.kt
MohamedRejeb: The idea of `ImagePainter` is not clear to me.
Check the new changes, I removed the `CoilPainter` and added `fun Canvas.onDraw()` to `Image`.

coil-core/src/nonAndroidMain/kotlin/coil3/Image.kt
```diff
@@ -18,33 +19,49 @@ actual interface Image {
     actual val height: Int
     actual val shareable: Boolean
 
-    fun asBitmap(): Bitmap
-}
+    fun Canvas.onDraw()
```
colinrtwhite, line 22: Would it be possible for us to make this a common function for `Image`? I'm thinking we could `typealias Canvas = android.graphics.Canvas` on Android and `typealias Canvas = org.jetbrains.skia.Canvas` on non-Android.

coil-core/src/nonAndroidMain/kotlin/coil3/Image.kt
```diff
@@ -18,33 +19,49 @@ actual interface Image {
...
 @ExperimentalCoilApi
 @Data
 class BitmapImage internal constructor(
-    val bitmap: Bitmap,
+    bitmap: Bitmap,
     override val shareable: Boolean,
 ) : Image {
+    private val image =
+        try {
+            org.jetbrains.skia.Image.makeFromBitmap(bitmap)
```
colinrtwhite, line 39: Hmm doesn't seem like we can avoid this conversion unfortunately since `Bitmap` can't draw directly to `Canvas`. Let's rename the `asBitmap` functions to `toBitmap` since they no longer wrap the underlying data and instead do a conversion.

coil-core/src/nonAndroidMain/kotlin/coil3/Image.kt
```diff
@@ -18,33 +19,49 @@ actual interface Image {
...
+        } finally {
+            bitmap.close()
```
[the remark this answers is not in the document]
MohamedRejeb: Yes, but we can't make sure that the bitmap is shared, and also if it is shared then closing the bitmap will cause a problem, but it's not the case. I'm thinking of removing this line `bitmap.setImmutable()` so we can make sure that it's not going to be shared and we can close it safely.

coil-svg/src/nonAndroidMain/kotlin/coil3/SvgImage.kt
```diff
@@ -0,0 +1,31 @@
...
+)
+
+@ExperimentalCoilApi
+private class SvgImage(
+    val svg: SVGDOM,
+    override val width: Int,
+    override val height: Int,
+) : Image {
+    override val size: Long
+        get() = 4L * width * height
```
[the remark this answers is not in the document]
MohamedRejeb: I was thinking of the byte size of the SVG file, since the SVG is not going to be resized unlike the Image.

coil-compose-core/src/nonAndroidMain/kotlin/coil3/compose/AsyncImagePainter.nonAndroid.kt
```diff
@@ -15,10 +18,23 @@ import coil3.request.crossfadeMillis
 internal actual fun Image.toPainter(
     context: PlatformContext,
     filterQuality: FilterQuality,
-): Painter = BitmapPainter(
-    image = asBitmap().asComposeImageBitmap(),
-    filterQuality = filterQuality,
-)
+): Painter = object : Painter() {
```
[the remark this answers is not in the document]
MohamedRejeb: Done, it's called `DrawablePainter` now since now it's only used for `DrawableImage`

coil-core/src/nonAndroidMain/kotlin/coil3/Image.kt
MohamedRejeb: I removed this and I am keeping the bitmap so we can use the `BitmapPainter` and use `FilterQuality` with bitmaps

colinrtwhite (approved):
Thanks for working on this! LGTM

coil-core/src/nonAndroidMain/kotlin/coil3/Image.kt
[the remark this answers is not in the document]
colinrtwhite: @MohamedRejeb We want to strongly prefer immutable bitmaps as they have performance benefits when being rendered

coil-svg/src/nonAndroidMain/kotlin/coil3/SvgImage.kt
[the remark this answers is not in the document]
colinrtwhite: @MohamedRejeb That's a good idea!

Issue #2031, opened 2024-01-08
[Coil 3] Uri `pathSegment` generates unexpected empty segments


SimonMarquis:
**Describe the bug**

`coil3.Uri.pathSegments` extension generates unexpected empty segments (when compared with Android's Uri).

**To Reproduce**

```kotlin
"https://example.org".toUri().pathSegments // Doesn't work at the moment because of #2029
"https://example.org/".toUri().pathSegments // [""]
"https://example.org/foo".toUri().pathSegments // ["foo"]
"https://example.org/foo/".toUri().pathSegments // ["foo", ""]
```

Android's `Uri.pathSegments` will never return empty segments, is this coil behavior on purpose?

https://github.com/coil-kt/coil/blob/9dd4ad7e14e9dd596dd7c60dad4c5b5d5ef52736/coil-core/src/commonMain/kotlin/coil3/Uri.common.kt#L29-L33

colinrtwhite:
I don't think this is expected - we should shoot for compatibility with Android's Uri where it makes sense.

PR #2032, opened 2024-01-09
Strip empty uri path segments.

colinrtwhite:
Fixes: https://github.com/coil-kt/coil/issues/2031

### DIFF
--- a/coil-core/src/commonMain/kotlin/coil3/Uri.common.kt
+++ b/coil-core/src/commonMain/kotlin/coil3/Uri.common.kt
@@ -44,7 +44,6 @@ val Uri.pathSegments: List<String>
                 segments += segment
             }
         }
-
         return segments
     }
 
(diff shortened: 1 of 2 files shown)


coil-core/src/commonMain/kotlin/coil3/Uri.common.kt
```diff
@@ -29,7 +29,22 @@ class Uri internal constructor(
...
+            if (index == -1) {
+                index = path.length
+            }
+
+            val segment = path.substring(startIndex, index)
+            if (segment.isNotEmpty()) {
+                segments += segment
+            }
+        }
+        return segments
```
connyduck, line 47: why not
```suggestion
        return  path.substring(1)
            .split('/')
            .filterNot { it.isEmpty() }
```
?

colinrtwhite: @connyduck Using `split('/')` and `filterNot { it.isEmpty() }` iterates over the data twice and creates two lists. The other solution does one iteration and creates one list.

Issue #2048, opened 2024-01-16
Inconsistencies in SVG Image Sizing and Blurriness When Loaded from URLs Using Coil in Jetpack Compose


NatashaIJ:
**Description:**
In a Jetpack Compose project utilizing Coil (version 2.4.0), SVG images loaded from URLs fail to retain their intrinsic dimensions, scaling up to the maximum size set with .sizeIn(maxWidth = 16.dp, maxHeight = 16.dp). This issue does not occur with SVGs loaded from drawable resources, which correctly maintain their intrinsic sizes. To ensure accurate sizing, Size.ORIGINAL is utilized in the image request, which leads to another problem: the images become blurry due to scaling from pixel-based sizes to dp in the UI.

**Steps to Reproduce:**
1. Initiate a Jetpack Compose project with Coil for image loading.
2. Load an SVG from a URL, which has smaller intrinsic dimensions than the maximum size, using .sizeIn(maxWidth = 16.dp, maxHeight = 16.dp) in the Image modifier
3. Observe the SVG incorrectly scaling to the maximum set size.
4. Apply Size.ORIGINAL to retrieve correct intrinsic dimensions for sizing in the Image modifier.
5. Notice the correct sizing but with resultant blurriness.

**Expected Behavior:**
1. SVG images from URLs should reflect their intrinsic dimensions without scaling to a set maximum, similar to drawable resource behavior.
2. When employing Size.ORIGINAL, the images should remain clear, without the blurriness introduced by upscaling for the UI's dp dimensions.

Additional Context:
A custom SvgDecoder was introduced to tackle the blurriness, altering the getDstSize function to translate SVG dimensions into dp for UI consistency. While this resolved the blurriness, modifying library components is risky, potentially leading to future compatibility issues. Ideally, default support for this scenario in the SVGDecoder would avoid the need for custom implementations.

**Version:**
Coil version: 2.4.0
The problem has been noted on various API levels and Android devices.

**Related Bug and Discussions:**
1. [Issue #566 with a similar problem and discussion](https://github.com/coil-kt/coil/issues/566#issuecomment-717244394)
2. [A related discussion on handling SVGs with Coil](https://github.com/coil-kt/coil/discussions/1455)

**Note:**
In my view, such intrinsic size handling should be natively supported by Coil's SVGDecoder to avoid the need for custom implementations that may introduce future maintenance challenges.

colinrtwhite:
[...]

PR #2090, opened 2024-01-30
Update SvgDecoder.android.kt to support scaling

seanpont:
Scales SVGs correctly

Issue: https://github.com/coil-kt/coil/issues/2048

### DIFF
--- a/coil-svg/src/androidMain/kotlin/coil3/svg/SvgDecoder.android.kt
+++ b/coil-svg/src/androidMain/kotlin/coil3/svg/SvgDecoder.android.kt
@@ -65,18 +65,20 @@ actual class SvgDecoder @JvmOverloads actual constructor(
             )
             bitmapWidth = (multiplier * svgWidth).toInt()
             bitmapHeight = (multiplier * svgHeight).toInt()
+            svg.setDocumentWidth("${svgWidth}px")
+            svg.setDocumentHeight("${svgHeight}px")
+            svg.scale(multiplier)
         } else {
             bitmapWidth = dstWidth
             bitmapHeight = dstHeight
+            svg.setDocumentWidth("100%")
+            svg.setDocumentHeight("100%")
         }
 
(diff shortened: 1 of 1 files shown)


colinrtwhite:
Thanks! I think we can ship this with a few changes:

- Let's put these changes behind a flag in `SvgDecoder.Factory`'s constructor. We need to preserve the existing behaviour for existing users.
- Do we need to make the same changes to `SvgDecoder.nonAndroid`? Ideally we want the Android and non-Android implementations to behave as similarly as possible.
- Please add a test in `AbstractSvgDecoderTest`. This'll make sure that we don't break the behaviour in the future.

seanpont:
Hey Colin,

The changes that you suggest sound reasonable, but I'm not able to make
those modifications right now. We (Google) are using the prior version, and
upgrading to 3.x so that I can make the change to head is going to take a
little longer.

colinrtwhite:
@seanpont Ah gotcha! If you'd like to make the change to the 2.x releases you'll need to target your PR to the `2.x` branch. `main` currently has all the 3.0 alpha code which supports multiplatform.

I think we should still add a flag to `SvgDecoder`'s constructor to avoid changing the behaviour for existing users. Maybe `densityScalingEnabled: Boolean = false`?

I can work on releasing `2.6.0` with those changes once they're merged.

davidcorradowell:
Looking at the commits does svg.scale even exist?  I attempted to copy this code and it did not work.  Is there something I am missing to get svg.scale be available?

Is there any obvious downsides of doing an implementation like this by default eventually?

colinrtwhite:
Hi folks, I added a `scaleToDensity` construction param to `SvgDecoder` [here](https://github.com/coil-kt/coil/pull/2502) that should produce similar behaviour to this PR (multiplying the source dimensions by the device's density). It'll be available in the next Coil 3.x release.

PR #2116, opened 2024-02-11
Remember AsyncImageState.

colinrtwhite:
I think this will work with strong skipping enabled.


colinrtwhite:
Unfortunately it doesn't seem like this will work even with strong skipping.

PR #2183, opened 2024-03-26
Support extracting video image by frame index

### LINKED ISSUE #2180: Support retrieving video image by frame index on VideoFrameDecoder
There is no way to decode an image by its frame index using coil-video.

MediaMetadataRetriever has [getFrameAtIndex](https://developer.android.com/reference/android/media/MediaMetadataRetriever#getFrameAtIndex(int,%20android.media.MediaMetadataRetriever.BitmapParams)) that is added on API 28, but it is not implemented on `VideoFrameDecoder` yet.

Add a video frame index option to `VideoFrameDecoder` in addition to the current `videoFrameMillis`, `videoFrameMicros` and `videoFramePercent`

**Additional context**
I've implemented this for my project https://github.com/coil-kt/coil/compare/main...tnj:coil:video-index-support

tnj:
This fixes #2180.

This changeset adds the `videoFrameIndex` option to the `VideoFrameDecoder`, which allows extracting an image from videos by its index (frame number).

### DIFF
--- a/coil-video/src/main/java/coil3/video/imageRequests.kt
+++ b/coil-video/src/main/java/coil3/video/imageRequests.kt
@@ -10,6 +10,32 @@ import coil3.getExtra
 import coil3.request.ImageRequest
 import coil3.request.Options
 
+// region videoFrameIndex
+
+/**
+ * Set the the frame index to extract from a video.
+ *
+ * When both [videoFrameIndex] and other videoFrame-prefixed properties are set,
+ * [videoFrameIndex] will take precedence.
+ */
+fun ImageRequest.Builder.videoFrameIndex(frameIndex: Int) = apply {
+    require(frameIndex >= 0) { "frameIndex must be >= 0." }
+    memoryCacheKeyExtra("coil#videoFrameIndex", frameIndex.toString())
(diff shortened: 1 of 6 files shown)


coil-video/src/main/java/coil3/video/imageRequests.kt
```diff
@@ -10,6 +10,32 @@ import coil3.getExtra
...
 
+// region videoFrameIndex
+
+/**
+ * Set the the frame index to extract from a video.
+ *
+ * When both [videoFrameIndex] and other videoFrame-prefixed properties are set,
+ * [videoFrameIndex] will take precedence.
+ */
+fun ImageRequest.Builder.videoFrameIndex(frameIndex: Int) = apply {
```
colinrtwhite, line 21: Let's annotate these methods with `@RequiresApi(28)` since they need API 28 to work.

colinrtwhite (approved):
Nice! Thanks for implementing this. Left one comment.

coil-video/src/main/java/coil3/video/imageRequests.kt
tnj: Sure thing. I've added it to the README too.

Issue #1866, opened 2023-09-19
Performance issue in lazy grid


PauliusVal:
**Description**
Compose performance issue in super simple app which just [displays network images list using Coil](https://github.com/PauliusVal/ComposeGridPerformance/blob/main/app/src/main/java/net/composegridperformance/feature/images/ImagesGridScreen.kt).

We see noticeable jank while scrolling even though:
- app runs in R8 release mode
- contains baseline profile for scrolling screen
- stable Compose parameters used
- unique and stable item keys applied for grid items

Jank is visible by human eye and also in profiler:
https://github.com/PauliusVal/ComposeGridPerformance/blob/main/images/images_grid_janky_frames_1.png?raw=true

Full app code and more screenshots: https://github.com/PauliusVal/ComposeGridPerformance/tree/main

**Version**
Coil version: `io.coil-kt:coil-compose:2.4.0`

colinrtwhite:
Thanks for the repro project! Which device did you run the tests on?

PauliusVal:
Samsung S21 Ultra, Android 13

PauliusVal:
Not sure if that helps, but updated the app with identical grid just with Glide image loading. It has noticeably fewer janky frames in profiler and looks/feels much smoother.

Mkohm:
Same problem here, seems to use a lot of memory! Up to 2gigs of memory in my app when scrolling in lazy column with images.

lkhore:
Same issue janky frames with grid

colinrtwhite:
@Mkohm That sounds like it could be a separate memory leak. Please file an issue (ideally with a way to reproduce)!

@lkhore Did you test in release mode? Compose is often much slower in debug mode.

Kuki93:
Is there any progress on this issue

matejdro:
I've added benchmark and tracing here to make diagnosing this easier: https://github.com/matejdro/ComposeGridPerformance/commit/beb902c3ad6d1348c169fc4565524fcaf0da5c5


Here is an example trace file: [ScrollBenchmark_scrollImageList_iter003_2023-12-21-08-27-41.perfetto-trace.zip](https://github.com/coil-kt/coil/files/13739015/ScrollBenchmark_scrollImageList_iter003_2023-12-21-08-27-41.perfetto-trace.zip)

Not sure why the trace is so weird (it shows one big compose event spanning multiple seconds), but you can still drill down and it appears that `rememberAsyncPainter` is taking considerable time:

[...]

PR #2205, opened 2024-04-16
Improve RealImageLoader.execute to properly call async request

mlykotom:
I was investigating the performance of `rememberAsyncImagePainter` and found out the `AsyncImagePainter.onRemembered` is the longest call inside ~0.25ms for each image.
image

When I was investigating more, I think the reason of this is that the `RealImageLoader.execute` request uses `async` call with dispatcher `Main.immediate`, but the coroutine is already called with `Main.immediate`. If I understand things correclty, this can't be called in parallel and therefore is called sequentially, hence the duration.

This should improve #1866 .

before
```
ScrollBenchmark_fullCompilation
coil.compose.rememberAsyncImagePainter%AverageMs   min     0,5,   median     0,6,   max     0,7
coil.compose.rememberAsyncImagePainter%Count       min   156,0,   median   173,0,   max   190,0
coil.compose.rememberAsyncImagePainter%Ms          min    81,0,   median    99,9,   max   128,2
timeToInitialDisplayMs                             min   525,2,   median   547,1,   max   579,6
frameDurationCpuMs                                 P50     10,5,   P90     31,0,   P95     42,1,   P99    315,3
frameOverrunMs                                     P50     -7,5,   P90     17,1,   P95     63,4,   P99  1 170,9
Traces: Iteration 0 1 2 3 4 5 6 7 8 9
```

after
```
ScrollBenchmark_fullCompilation
coil.compose.rememberAsyncImagePainter%AverageMs   min     0,3,   median     0,4,   max     0,5
coil.compose.rememberAsyncImagePainter%Count       min   154,0,   median   174,0,   max   184,0
coil.compose.rememberAsyncImagePainter%Ms          min    51,4,   median    70,6,   max    83,1
timeToInitialDisplayMs                             min   521,4,   median   563,9,   max   605,9
frameDurationCpuMs                                 P50      9,6,   P90     30,7,   P95     45,1,   P99    301,4
frameOverrunMs                                     P50     -8,1,   P90     22,8,   P95     52,3,   P99  1 172,8
Traces: Iteration 0 1 2 3 4 5 6 7 8 9
```

From the benchmarks, I can see that the frame duration was improved slightly for the scroll.
```
before : frameDurationCpuMs  P50     10,5,   P90     31,0,   P95     42,1,   P99    315,3
after    : frameDurationCpuMs  P50      9,6, (-9%)   P90     30,7, (-1%)   P95     45,1, (+6%)  P99    301,4 (-5%)
```

This solution is probably not sufficient as the dispatcher probably needs to be passed, or called somewhere else, or even this might be done differently all together.

I have experimented with this in 2.x branch, but I see a similar code is in 3.x.

### DIFF
--- a/coil-compose-base/src/main/java/coil/compose/AsyncImagePainter.kt
+++ b/coil-compose-base/src/main/java/coil/compose/AsyncImagePainter.kt
@@ -1,6 +1,5 @@
 package coil.compose
 
-import android.annotation.SuppressLint
 import android.graphics.drawable.BitmapDrawable
 import android.graphics.drawable.Drawable
 import androidx.compose.foundation.Image
@@ -10,12 +9,12 @@ import androidx.compose.runtime.Composable
 import androidx.compose.runtime.NonRestartableComposable
 import androidx.compose.runtime.RememberObserver
 import androidx.compose.runtime.Stable
-import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableFloatStateOf
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.setValue
+import androidx.compose.runtime.snapshotFlow
 import androidx.compose.ui.geometry.Size
 import androidx.compose.ui.geometry.isUnspecified
 import androidx.compose.ui.graphics.ColorFilter
@@ -210,9 +209,7 @@ private fun rememberAsyncImagePainter(
     painter.filterQuality = filterQuality
     painter.isPreview = LocalInspectionMode.current
     painter.imageLoader = state.imageLoader
-    @SuppressLint("StateFlowValueCalledInComposition")
-    painter.requestInternal.value =
-        request // Update request last so all other properties are up to date.
+    painter.request = request // Update request last so all other properties are up to date.
     painter.onRemembered() // Invoke this manually so `painter.state` is set to `Loading` immediately.
     return painter
 }
@@ -251,19 +248,17 @@ class AsyncImagePainter internal constructor(
     internal var contentScale = ContentScale.Fit
     internal var filterQuality = DefaultFilterQuality
     internal var isPreview = false
-    internal val requestInternal = MutableStateFlow(request)
 
     /** The current [AsyncImagePainter.State]. */
     var state: State by mutableStateOf(State.Empty)
         private set
 
     /** The current [ImageRequest]. */
-    val request
-        @Composable
-        get() = requestInternal.collectAsState().value
+    var request by mutableStateOf(request)
+        internal set
 
     /** The current [ImageLoader]. */
-    var imageLoader: ImageLoader = imageLoader
+    var imageLoader: ImageLoader by mutableStateOf(imageLoader)
         internal set
 
     override val intrinsicSize: Size
@@ -301,14 +296,14 @@ class AsyncImagePainter internal constructor(
 
         // If we're in inspection mode skip the image request and set the state to loading.
         if (isPreview) {
-            val request = requestInternal.value.newBuilder().defaults(imageLoader.defaults).build()
+            val request = request.newBuilder().defaults(imageLoader.defaults).build()
             updateState(State.Loading(request.placeholder?.toPainter()))
             return@trace
         }
 
         // Observe the current request and execute any emissions.
         scope.launch {
-            requestInternal
+            snapshotFlow { request }
                 .mapLatest { imageLoader.execute(updateRequest(it)).toState() }
                 .collect(::updateState)
         }
--- a/coil-compose-base/src/main/java/coil/compose/AsyncImagePainter.kt
+++ b/coil-compose-base/src/main/java/coil/compose/AsyncImagePainter.kt
@@ -1,5 +1,6 @@
 package coil.compose
 
+import android.annotation.SuppressLint
 import android.graphics.drawable.BitmapDrawable
 import android.graphics.drawable.Drawable
 import androidx.compose.foundation.Image
@@ -9,12 +10,12 @@ import androidx.compose.runtime.Composable
 import androidx.compose.runtime.NonRestartableComposable
 import androidx.compose.runtime.RememberObserver
 import androidx.compose.runtime.Stable
+import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableFloatStateOf
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.setValue
-import androidx.compose.runtime.snapshotFlow
 import androidx.compose.ui.geometry.Size
 import androidx.compose.ui.geometry.isUnspecified
 import androidx.compose.ui.graphics.ColorFilter
@@ -209,7 +210,9 @@ private fun rememberAsyncImagePainter(
     painter.filterQuality = filterQuality
     painter.isPreview = LocalInspectionMode.current
     painter.imageLoader = state.imageLoader
-    painter.request = request // Update request last so all other properties are up to date.
+    @SuppressLint("StateFlowValueCalledInComposition")
+    painter.requestInternal.value =
+        request // Update request last so all other properties are up to date.
     painter.onRemembered() // Invoke this manually so `painter.state` is set to `Loading` immediately.
     return painter
 }
@@ -248,17 +251,19 @@ class AsyncImagePainter internal constructor(
     internal var contentScale = ContentScale.Fit
     internal var filterQuality = DefaultFilterQuality
     internal var isPreview = false
+    internal val requestInternal = MutableStateFlow(request)
 
     /** The current [AsyncImagePainter.State]. */
     var state: State by mutableStateOf(State.Empty)
         private set
 
     /** The current [ImageRequest]. */
-    var request: ImageRequest by mutableStateOf(request)
-        internal set
+    val request
+        @Composable
+        get() = requestInternal.collectAsState().value
 
     /** The current [ImageLoader]. */
-    var imageLoader: ImageLoader by mutableStateOf(imageLoader)
+    var imageLoader: ImageLoader = imageLoader
         internal set
 
     override val intrinsicSize: Size
@@ -296,14 +301,14 @@ class AsyncImagePainter internal constructor(
 
         // If we're in inspection mode skip the image request and set the state to loading.
         if (isPreview) {
-            val request = request.newBuilder().defaults(imageLoader.defaults).build()
+            val request = requestInternal.value.newBuilder().defaults(imageLoader.defaults).build()
             updateState(State.Loading(request.placeholder?.toPainter()))
             return@trace
         }
 
         // Observe the current request and execute any emissions.
         scope.launch {
-            snapshotFlow { request }
+            requestInternal
                 .mapLatest { imageLoader.execute(updateRequest(it)).toState() }
                 .collect(::updateState)
         }
--- a/coil-compose-base/src/main/java/coil/compose/AsyncImagePainter.kt
+++ b/coil-compose-base/src/main/java/coil/compose/AsyncImagePainter.kt
@@ -30,6 +30,7 @@ import androidx.compose.ui.graphics.vector.ImageVector
 import androidx.compose.ui.layout.ContentScale
 import androidx.compose.ui.platform.LocalInspectionMode
 import androidx.compose.ui.unit.Constraints
+import androidx.compose.ui.util.trace
 import coil.ImageLoader
 import coil.compose.AsyncImagePainter.Companion.DefaultTransform
 import coil.compose.AsyncImagePainter.State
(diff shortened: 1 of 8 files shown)


colinrtwhite:
Thanks for taking a look at this! Though I'm a bit confused - won't `async(Dispatchers.Main.immediate)` always run sequentially (even if wrapped with `withContext(Dispatchers.Default)`) since `Dispatchers.Main.immediate` is a single threaded dispatcher?

Unfortunately I don't think we can use an non-main thread dispatcher here as it breaks returning images from the memory cache synchronously - this is the failing test.

Do you know if `RealImageLoader.executeMain` is taking up the bulk of the `AsyncImagePainter.onRemembered` time? There are some things that Coil needs to run on the main thread, but I'm curious if that's what's slowing it down at the moment or if it's something else.

mlykotom:
I sort of expected something to break, but I was a bit in a rush and wanted to start the conversation going 😇 

I pushed my wip code with experiments (`ScrollBenchmark`) that can produce the system trace and a **method trace**.
Here's the breakdown of the method trace
image

You can run the benchmark as well and get the method trace for analysis

mlykotom:
Okay, I tried a bit playing with this if something could be triggered earlier, but I'm not able to solve the memory cache blocking request. I thought there could be something running in parallel with the `async` call and `await` it at the apropriate time to basically have something like "if not ready, block". But I don't understand the details of when things are called.

colinrtwhite:
Interesting thanks for digging in! Looks like there are two main avenues to improve perf like you called out:

- **Speed up `scope.launch` in `AsyncImagePainter.onRemembered`**: I'm not sure what we can do here given the bulk of the time looks to be coroutines library-related. The result of the code inside `scope.launch` should be pretty quick.
- **Don't block when `ConstraintSizeResolver.size` is called**: I'm surprised that this is a large percentage of the runtime given `SizeResolver.size` should suspend instead of block while waiting for non-zero constraints. Do you know what could be the issue?

mlykotom:
1. Part of the `scope.launch` in `onRemembered` is because of reading the state, which happens twice - `request` and `imageLoader`. I'm thinking whether the `imageLoader` has to be as a `mutableStateOf`. I don't see this used for compose directly, so this might be a small optimization. 

2. The other thing might also be to not use `mutableStateOf` for the `request` parameter. I don't see it being read anywhere in the project, is it a niche case when someone might want to listen to this outside? Could this be a `MutableStateFlow`, which is used for everyone and used as `collectAsState()` composable if anyone needs listening to this?

3. Could the blocking nature of `ConstraintSizeResolver.size` be related to the fact it uses `Dispatchers.Main.immediate`.

mlykotom:
I experimented with removing the State variables, there's some minor change in overall duration, but the FPS impact might be too noisy

Before:
```
AsyncImagePainter.onRememberedAverageMs        min   0.1,   median   0.1,   max   0.1
AsyncImagePainter.onRememberedCount            min 130.0,   median 139.0,   max 146.0
AsyncImagePainter.onRememberedMs               min   9.3,   median  12.5,   max  16.3
frameDurationCpuMs                             P50    4.1,   P90    9.1,   P95   20.0,   P99   99.3
```

After:
```
AsyncImagePainter.onRememberedAverageMs        min   0.1,   median   0.1,   max   0.1
AsyncImagePainter.onRememberedCount            min 130.0,   median 141.0,   max 148.0
AsyncImagePainter.onRememberedMs               min   8.3,   median  10.5,   max  15.8
frameDurationCpuMs                             P50    4.4,   P90    8.6,   P95   12.0,   P99   95.1
```

mlykotom:
I played a bit more with what could be improved and the `async` call within `RealImageLoader.execute` is only needed for `Views`. So if we tweak the the code to use that only for Views, we can get rid of some coroutines overhead.

**Before (any tweaks)**
```
AsyncImagePainter.onRememberedAverage_µs       min     77.3,   median    101.5,   max    116.7
AsyncImagePainter.onRememberedCount            min    130.0,   median    130.0,   max    146.0
AsyncImagePainter.onRemembered_µs              min 10,048.4,   median 13,880.2,   max 16,900.0
timeToInitialDisplayMs                         min    256.8,   median    281.7,   max    358.9
frameDurationCpuMs                             P50       5.1,   P90       9.2,   P95      13.3,   P99     117.0
```

**After**
```
AsyncImagePainter.onRememberedAverage_µs       min     68.9,   median     78.7,   max     91.1
AsyncImagePainter.onRememberedCount            min    128.0,   median    141.0,   max    146.0
AsyncImagePainter.onRemembered_µs              min  8,956.1,   median 11,163.1,   max 13,111.9
timeToInitialDisplayMs                         min    234.1,   median    276.5,   max    325.7
frameDurationCpuMs                             P50       4.5,   P90       8.9,   P95      15.5,   P99     120.3
```

* The results would be better taken with Microbenchmark, but hard to do for public Compose library.

colinrtwhite:
@mlykotom This is awesome, thanks! Do you want to merge this in with the scroll perf benchmark?

To improve scroll perf even more, I think if we offer an opt-in option for Compose (or maybe the default in 3.x?) we should be able to move all of `executeMain` off the main thread. Currently, it only requires it to interact with views, lifecycles, and invoke `listener` methods, which aren't necessary for `AsyncImage`/`AsyncImagePainter`.

mlykotom:
Hey, sorry I was super busy this week and couldn't get to it.
I will clean up the PR to be in a submittable shape.

>To improve scroll perf even more, I think if we offer an opt-in option for Compose (or maybe the default in 3.x?) we should be able to make executeMain thread agnostic (though, I think ConstraintsSizeResolve.size will still run on the UI thread). Currently, it only requires it to interact with views, lifecycles, and invoke listener methods, which aren't necessary for AsyncImage/AsyncImagePainter.

I think this sounds really great to think about the threads in 3.x, I would vote for having it default, but definitelly needs some thinking.

coil-compose-base/src/main/java/coil/compose/AsyncImagePainter.kt
```diff
@@ -247,17 +251,19 @@ class AsyncImagePainter internal constructor(
...
     /** The current [AsyncImagePainter.State]. */
     var state: State by mutableStateOf(State.Empty)
         private set
 
     /** The current [ImageRequest]. */
-    var request: ImageRequest by mutableStateOf(request)
-        internal set
+    val request
+        @Composable
+        get() = requestInternal.collectAsState().value
```
mlykotom, line 263: @colinrtwhite what do you think about this change? Technically this changes the API, because now the `request` is only available from a composable, which wasn't before. I'm not sure how this could be done in a backwards compatible way 😬

coil-compose-base/src/main/java/coil/compose/AsyncImagePainter.kt
```diff
@@ -282,9 +288,9 @@ class AsyncImagePainter internal constructor(
     }
 
     @OptIn(ExperimentalCoroutinesApi::class)
-    override fun onRemembered() {
+    override fun onRemembered() = trace("AsyncImagePainter.onRemembered") {
```
mlykotom, line 291: I'm thinking whether the usage of `onRemembered` is used actually? This is manually called immediatelly in composition and then compose framework calls it again after the composition and it always just returns with the check on `rememberScope != null`. 

I don't fully understand (yet) the child painter and it's relation to `RememberObserver`, but feels like it could call them differently as well?

coil-compose-base/src/main/java/coil/compose/AsyncImagePainter.kt
```diff
@@ -295,14 +301,14 @@ class AsyncImagePainter internal constructor(
...
             updateState(State.Loading(request.placeholder?.toPainter()))
-            return
+            return@trace
         }
 
         // Observe the current request and execute any emissions.
         scope.launch {
-            snapshotFlow { request }
+            requestInternal
                 .mapLatest { imageLoader.execute(updateRequest(it)).toState() }
```
mlykotom, line 312: What if this is actually just a `var _request: ImageRequest`?
The request is changed from outside right before calling `onRemembered` manually, which means it's actually up-to-date here.

I don't know what's the case for the nested painters.

coil-compose-base/src/main/java/coil/compose/AsyncImagePainter.kt
```diff
@@ -247,17 +251,19 @@ class AsyncImagePainter internal constructor(
     internal var contentScale = ContentScale.Fit
     internal var filterQuality = DefaultFilterQuality
     internal var isPreview = false
+    internal val requestInternal = MutableStateFlow(request)
```
colinrtwhite, line 254: Is the a performance benefit to reading request from a `MutableStateFlow` vs. a `mutableStateOf`?

coil-compose-base/src/main/java/coil/compose/AsyncImagePainter.kt
```diff
@@ -247,17 +251,19 @@ class AsyncImagePainter internal constructor(
...
     /** The current [ImageRequest]. */
-    var request: ImageRequest by mutableStateOf(request)
-        internal set
+    val request
+        @Composable
+        get() = requestInternal.collectAsState().value
 
     /** The current [ImageLoader]. */
-    var imageLoader: ImageLoader by mutableStateOf(imageLoader)
+    var imageLoader: ImageLoader = imageLoader
```
colinrtwhite, line 266: I think we need to keep this as a `mutableStateOf` otherwise it violates `@Stable`'s contract. It's possible to change the `ImageLoader` from `rememberAsyncImagePainter` and the composition should be notified.

coil-compose-base/src/main/java/coil/compose/AsyncImagePainter.kt
colinrtwhite: I'd prefer to keep it as a non-Composable since it's then the same as the other exposed properties (`imageLoader` and `state`). Though if there's a strong performance case then we should consider changing all the exposed properties to be `@Composable`.

coil-compose-base/src/main/java/coil/compose/AsyncImagePainter.kt
colinrtwhite: I'm not too sure of Compose's behaviour, but I added this in on `onRemembered` since we want to stop any ongoing requests in `onForgotten`. This keeps the symmetry in case the painter is re-remembered after being forgotten.

The child painter is usually either a `CrossfadePainter` if crossfading is enabled or a `DrawablePainter`, which wraps the `Drawable` returned by the `ImageLoader`.

coil-compose-base/src/main/java/coil/compose/AsyncImagePainter.kt
colinrtwhite: I think we might need to read this as a state, but I can't remember the specifics. If I remember correctly there are cases were `mutableStateOf(request)` won't be equal to `_request` since it exists outside of the Composition. Also it's possible for `request` to change so we need to observe it.

coil-compose-base/src/main/java/coil/compose/AsyncImagePainter.kt
```diff
@@ -197,7 +199,7 @@ private fun rememberAsyncImagePainter(
     onState: ((State) -> Unit)?,
     contentScale: ContentScale,
     filterQuality: FilterQuality,
-): AsyncImagePainter {
+): AsyncImagePainter = trace("rememberAsyncImagePainter") {
```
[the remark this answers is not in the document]
mlykotom: This trace doesn't add any overhead (~6ns per call) and it's fine to ship to production. Compose adds bunch of custom sections as well. These usually serve for better understanding what's going on.

mlykotom (commented):
Ok, I reverted the breaking changes, so these can be done in a separate PR so the effect is smaller (~0.1ms per item). Still that gets 3ms in total when scrolling content.

```
before
ScrollBenchmark_fullCompilation
AsyncImagePainter.onRememberedAverage_µs   min    112.2,   median    112.2,   max    112.2
AsyncImagePainter.onRememberedCount        min    130.0,   median    130.0,   max    130.0
AsyncImagePainter.onRemembered_µs          min 14,586.9,   median 14,586.9,   max 14,586.9
rememberAsyncImagePainterAverage_µs        min    244.4,   median    244.4,   max    244.4
rememberAsyncImagePainterCount             min     65.0,   median     65.0,   max     65.0
rememberAsyncImagePainter_µs               min 15,888.1,   median 15,888.1,   max 15,888.1

after
ScrollBenchmark_fullCompilation
AsyncImagePainter.onRememberedAverage_µs   min     90.7,   median     90.7,   max     90.7
AsyncImagePainter.onRememberedCount        min    130.0,   median    130.0,   max    130.0
AsyncImagePainter.onRemembered_µs          min 11,785.6,   median 11,785.6,   max 11,785.6
rememberAsyncImagePainterAverage_µs        min    197.5,   median    197.5,   max    197.5
rememberAsyncImagePainterCount             min     65.0,   median     65.0,   max     65.0
rememberAsyncImagePainter_µs               min 12,839.0,   median 12,839.0,   max 12,839.0
```

coil-compose-base/src/main/java/coil/compose/AsyncImagePainter.kt
colinrtwhite: @mlykotom I think 0.2ms is probably a strong enough incentive to change the API. Though unfortunately, we can't make any breaking changes to the 2.x branch. We could make the changes on the 3.x (`main`) branch, however.

colinrtwhite:
@mlykotom Thanks for all the work on this! I'll take a look through the other optimizations this weekend. Unfortunately we can't make any binary incompatible changes to the 2.x branch, but we should consider incorporating some of them into 3.x (the `main` branch). I'll also upstream this change to `main`.

PR #2239, opened 2024-05-09
Remove compiler work-around.

colinrtwhite:


colinrtwhite:
This still fails if you run `./gradlew assemble`.

PR #2241, opened 2024-05-09
Support passing CoroutineContext to any CoroutineDispatcher methods.

colinrtwhite:


coil-core/src/commonMain/kotlin/coil3/ImageLoader.kt
```diff
@@ -202,39 +202,39 @@ interface ImageLoader {
...
             )
         }
 
         /**
-         * The [CoroutineDispatcher] that the [Interceptor] chain will be executed on.
+         * The [CoroutineContext] that the [Interceptor] chain will be executed in.
          *
          * Default: `Dispatchers.Main.immediate`
          */
-        fun interceptorDispatcher(dispatcher: CoroutineDispatcher) = apply {
```
colinrtwhite, line 219: @jingibus Do you think it's worth renaming these to `interceptorContext`, etc.? Is there a naming pattern we've followed in other libraries?

I'm leaning towards keeping the names as is since they're more discoverable and in 90% of cases users only want to set the dispatcher.

jingibus: I don't have strong opinions about it. In Cash's codebase, we tend to use `dispatcher` for the injected `val`.

This API change might be cause to modify how the defaults work, too: it is possible that you want `EmptyCoroutineContext` or any context without a dispatcher to also default to `Dispatchers.Main.immediate`? Or would it be appropriate for `EmptyCoroutineContext` to mean what it does in other contexts: "use whatever dispatcher you're already on"?

jingibus (approved):

coil-core/src/commonMain/kotlin/coil3/ImageLoader.kt
colinrtwhite: :+1: Sounds good. I'll leave the names as is for the moment.

For the defaults I think we want to leave them as is too. They're opinionated (for instance, we almost always want to fetch on a background dispatcher), but will work in any context.

PR #2251, opened 2024-05-12
Fix uri parse error in the Windows system.

Caij:

### DIFF
--- a/coil-core/src/commonTest/kotlin/coil3/UriTest.kt
+++ b/coil-core/src/commonTest/kotlin/coil3/UriTest.kt
@@ -40,6 +40,17 @@ class UriTest {
     }
 
     @Test
+    fun windows() {
+        val uri = "D:\\test\\relative\\image.jpg".toUri()
+        assertNull(uri.scheme)
+        assertNull(uri.authority)
+        assertEquals(uri.path, "D:\\test\\relative\\image.jpg")
+        assertEquals(listOf("D:", "test", "relative", "image.jpg"), uri.pathSegments)
+        assertNull(uri.query)
+        assertNull(uri.fragment)
+    }
+
+    @Test
     fun veryMalformed() {
         val uri = "/#02dkfj;anc%%2".toUri()
         assertNull(uri.scheme)
--- a/coil-core/src/commonMain/kotlin/coil3/Uri.common.kt
+++ b/coil-core/src/commonMain/kotlin/coil3/Uri.common.kt
@@ -1,5 +1,7 @@
 package coil3
 
+import coil3.util.isFileUri
+
 /**
  * A uniform resource locator.
  */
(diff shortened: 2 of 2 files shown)


Caij:
Windows local path parse error.  D:\\test\\relative\\image.jpg

```
class Factory : Fetcher.Factory<Uri> {
        override fun create(
            data: Uri,
            options: Options,
            imageLoader: ImageLoader,
        ): Fetcher? {
            if (!isFileUri(data)) return null
            return FileUriFetcher(data, options)
        }
    }
```

```
internal fun isFileUri(uri: Uri): Boolean {
    return (uri.scheme == null || uri.scheme == SCHEME_FILE) &&
        uri.path != null &&
        !isAssetUri(uri)
}
```
Here the uri.path is empty,  so the FileUriFetcher factory can't be found.

coil-core/src/commonTest/kotlin/coil3/UriTest.kt
```diff
@@ -39,6 +39,17 @@ class UriTest {
         assertEquals("something:/test/relative/image.jpg", uri.fragment)
     }
 
+    @Test
+    fun windows() {
+        val uri = "D:\\test\\relative\\image.jpg".toUri()
+        assertNull(uri.scheme)
```
colinrtwhite, line 45: Is it correct that scheme is null? Should it be `D:`?

colinrtwhite:
Thanks for taking a stab at this - unfortunately looks like the tests are failing with a `StackOverflowError`. You can run the tests locally using `./test.sh`.

coil-core/src/commonTest/kotlin/coil3/UriTest.kt
Caij: I think D is the volume information, it should be part of the path.

Caij: > Is it correct that scheme is null? Should it be `D`?

I think D is the volume information, it should be part of the path.

Caij:
> Thanks for taking a stab at this - unfortunately looks like the tests are failing with a `StackOverflowError`. You can run the tests locally using `./test.sh`.

There was an error and I fixed it, and it has been submitted.

colinrtwhite:
Sorry for the delay on merging this. After thinking about this I think we might want to handle Windows file paths a different way.

Changing the `Uri` class to handle backslash file paths technically violates the [`URI` spec](https://datatracker.ietf.org/doc/html/rfc3986). Instead the spec recommends converting the file paths to use forward slashes, which we could handle with a custom `Mapper` for Windows targets. Though I'm not sure how we would get the correct Windows path back afterwards to pass to Okio.

Caij:
> Sorry for the delay on merging this. After thinking about this I think we might want to handle Windows file paths a different way.
> 
> Changing the `Uri` class to handle backslash file paths technically violates the [`URI` spec](https://datatracker.ietf.org/doc/html/rfc3986). Instead the spec recommends converting the file paths to use forward slashes, which we could handle with a custom `Mapper` for Windows targets. Though I'm not sure how we would get the correct Windows path back afterwards to pass to Okio.

Yes, I don't think my approach is very elegant, but I can't think of a better way to handle this, looking forward to your solutions.
Our current solution is set custom fetch factory for Windows targets.

colinrtwhite:
@Caij Thanks for creating this PR - I ended up using some of your code in the final fix [here](https://github.com/coil-kt/coil/pull/2404), which should be available in the next alpha release.

PR #2317, opened 2024-06-16
Create Roborazzi test for desktop

### LINKED ISSUE #2095: Update Roborazzi tests to also run on desktop JVM
We should update our Roborazzi tests to run on the JVM using [Roborazzi's desktop support](https://github.com/takahirom/roborazzi?tab=readme-ov-file#experimental-feature-compose-desktop-support). Currently we only run the tests on Android.

ghostbear:
Resolves #2095


internal/test-roborazzi/build.gradle.kts
```diff
@@ -1,25 +1,63 @@
...
+
+roborazzi {
+    // RoborazziOptions.RecordOptions.outputDirectoryPath and roborazzi.output.dir seems to be ignored on desktop.
+    // To workaround that this is used.
+    outputDir = layout.projectDirectory.dir("src/jvmTest/snapshots/images")
+}
+
+tasks.withType<KotlinCompile>().configureEach {
+    compilerOptions {
+        freeCompilerArgs.add("-Xcontext-receivers")
```
colinrtwhite, line 61: Do we need to add `"-Xcontext-receivers"` here and above?

colinrtwhite (approved):
Looks great, thanks!

internal/test-roborazzi/build.gradle.kts
takahirom: Sorry for the inconvenience. I didn't think context receivers would take this long to be released. I'm willing to use normal parameters instead of context receivers, but for now, they are needed.

colinrtwhite: No prob! Was mostly curious if we can declare it in only one spot.

PR #2322, opened 2024-06-20
Decouple ImageLoader.execute from the main dispatcher.

colinrtwhite:


coil-compose-core/src/commonMain/kotlin/coil3/compose/AsyncImagePainter.kt
```diff
@@ -249,7 +249,7 @@ class AsyncImagePainter internal constructor(
         if (rememberScope != null) return@trace
 
         // Create a new scope to observe state and execute requests while we're remembered.
-        val scope = CoroutineScope(SupervisorJob() + Dispatchers.Main.immediate)
+        val scope = CoroutineScope(SupervisorJob() + Dispatchers.Unconfined)
```
colinrtwhite, line 252: `Dispatchers.Unconfined` will get replaced by [this](https://github.com/coil-kt/coil/pull/2220).

Issue #1543, opened 2022-11-24
Duplicate network request for images with chunked response


Fayth:
**Describe the bug**
Scenario:
* image caching disabled in coil
* requesting image via url (HttpUrlFetcher)
* server returns valid chunked response (doesn't contain `Content-Length` header)

results in two requests to given url (despite the first one being successful)

**To Reproduce**
```kotlin
val request = ImageRequest.Builder(context)
            .data("https://www.httpwatch.com/httpgallery/chunked/chunkedimage.aspx")
            .memoryCachePolicy(CachePolicy.DISABLED)
            .diskCachePolicy(CachePolicy.DISABLED)
            .build()
context.imageLoader.execute(request)
```

**Logs/Screenshots**
The cause seems to be the verification in https://github.com/coil-kt/coil/blob/main/coil-base/src/main/java/coil/fetch/HttpUriFetcher.kt#L95 which relies on explicit content length being provided in response. In the scenario above the length isn't known (okhttp reports it as -1)

**Version**
2.2.2

It might be enough to allow `contentLength() == -1` in the check mentioned above (it seems to fix the problem in my case) or possibly also peeking the response body buffer to make sure it's not empty

colinrtwhite:
@Fayth Does OkHttp have the same behaviour for that URL? Coil's `HttpUriFetcher` code is based off of OkHttp's behaviour and tries to adhere pretty closely to it.

Fayth:
I haven't used "raw" okhttp much, but from what I tried, there's only one GET request when using this sample request:
```
val request: Request = Request.Builder()
            .url(url)
            .cacheControl(CacheControl.FORCE_NETWORK)
            .build()
OkHttpClient().newCall(request).execute().use { response -> return response.body?.string() }
```
Looking through okhttp source, I think the only place where it uses content length in higher-level logic is ensuring there's no body for 204/205 responses.

ferinagy:
Hello, we just ran into this for `2.6.0`. I tried to create a failing test in `HttpUriFetcherTest`:

```kotlin
@Test
fun `chunked response`() = runTestAsync {
    val url = server.url(IMAGE).toString()

    server.enqueueChunkedImage(IMAGE)
    server.enqueueChunkedImage(IMAGE)

    val fetcher = newFetcher(url, Options(context, diskCachePolicy = CachePolicy.DISABLED))
    fetcher.fetch()

    assertEquals(1, server.requestCount)
}

[...]

PR #2363, opened 2024-07-09
Fix duplicate network call for chunked responses

ferinagy:
Fixes https://github.com/coil-kt/coil/issues/1543


colinrtwhite (approved):
This looks great, thanks! I'll include this in the `2.7.0` release, which should be out soon.

PR #2403, opened 2024-07-30
Replace ImageLoader.Builder.networkObserverEnabled with ConnectivityChecker interface.

colinrtwhite:


coil-network-core/src/androidMain/kotlin/coil3/network/ConnectivityChecker.kt
```diff
@@ -0,0 +1,52 @@
...
+import android.annotation.SuppressLint
+import android.net.ConnectivityManager
+import android.net.NetworkCapabilities.NET_CAPABILITY_INTERNET
+import android.os.Build.VERSION.SDK_INT
+import androidx.annotation.RequiresApi
+import androidx.core.content.getSystemService
+import coil3.PlatformContext
+import coil3.network.internal.isPermissionGranted
+
+actual fun ConnectivityChecker(context: PlatformContext): ConnectivityChecker {
```
colinrtwhite, line 13: cc @tech31 I'd love your feedback on this if possible. This replaces the `NetworkObserver` interface with public `ConnectivityChecker`. It removes usages of `NetworkCallback` as we don't need to be kept up to date with network changes. Instead it calls `getNetworkCapabilities` for the default network on each image request (on a background thread).

tech31: You could do this, but it really isn't the recommended way as you will be making a bunch of synchronous calls. The general suggestion is to rely on the async methods for best reliability and performance.

In your case you would:

- use `ConnectivityManager#registerDefaultNetworkCallback`
- use the passed in callback to understand if you have a network available and similarly, you can use the `onCapabilitiesChanged` event to know the network's capabilities
- assign the returned network capabilities to a local variable and check against that variable on each request to determine if you have connectivity or not
- if the variable is null (such as being set to null when no default network is available) or if the variable doesn't have the capability `VALIDATED`, then return false
- otherwise return true

That said, you can continue doing it as you are in this change, it's just not the recommended way of doing it.

coil-network-core/src/androidMain/kotlin/coil3/network/ConnectivityChecker.kt
```diff
@@ -0,0 +1,52 @@
...
+
+@RequiresApi(23)
+@SuppressLint("MissingPermission")
+private class ConnectivityCheckerApi23(
+    private val connectivityManager: ConnectivityManager,
+) : ConnectivityChecker {
+    override fun isOnline(): Boolean {
+        val activeNetwork = connectivityManager.activeNetwork
+        val capabilities = connectivityManager.getNetworkCapabilities(activeNetwork)
+        return capabilities != null && capabilities.hasCapability(NET_CAPABILITY_INTERNET)
```
tech31, line 39: Just one note on this; checking `NET_CAPABILITY_INTERNET` isn't entirely correct if the question you're trying to answer is "do I have connectivity". The internet capability just means this network _can_ use the internet, not that it is _connected_ to the internet.

https://developer.android.com/reference/android/net/NetworkCapabilities#NET_CAPABILITY_INTERNET

If you want to know whether the network is currently connected and able to use the internet, you want to check `NET_CAPABILITY_VALIDATED`.

https://developer.android.com/reference/android/net/NetworkCapabilities#NET_CAPABILITY_VALIDATED

PR #2408, opened 2024-08-03
Fix some broken links in docs

hakonschia:
Some 404 links I found in the docs

There are also some in the changelogs like to `Precision.kt` for [0.9.0](https://github.com/coil-kt/coil/blob/main/CHANGELOG.md#090---december-30-2019). Not sure if it makes sense to retroactively change a changelog


colinrtwhite:
Thanks! I'm going to move the existing docs into a separate 2.x subfolder with the old Dokka docs, and will merge this right after that.

colinrtwhite (approved):
Thanks!

PR #2460, opened 2024-08-29
Refactor static and animated ImageDecoder to share source creation utils.

colinrtwhite:


coil-core/src/androidMain/kotlin/coil3/decode/StaticImageDecoder.kt
```diff
@@ -106,40 +108,45 @@ class StaticImageDecoder(
...
+    }
 
-            val metadata = metadata
-            return when {
-                metadata is AssetMetadata -> {
-                    ImageDecoder.createSource(options.context.assets, metadata.filePath)
-                }
-                metadata is ContentMetadata -> try {
+    val metadata = metadata
+    return when {
```
revonateB0T, line 128: `when (val metadata = metadata)`
See https://github.com/Kotlin/KEEP/blob/master/proposals/val-in-when-subject.md

coil-core/src/androidMain/kotlin/coil3/decode/StaticImageDecoder.kt
```diff
@@ -106,40 +108,45 @@ class StaticImageDecoder(
...
-                    ImageDecoder.createSource(options.context.resources, metadata.resId)
-                }
-                metadata is ByteBufferMetadata -> {
-                    ImageDecoder.createSource(metadata.byteBuffer)
-                }
-                else -> null
+                    return@run ImageDecoder.createSource { asset }
+                } catch (_: ErrnoException) {}
             }
+            ImageDecoder.createSource(options.context.contentResolver, metadata.uri.toAndroidUri())
```
revonateB0T, line 142: I'm hesitated if we just fail to `BitmapFactory` on fd doesn't support seek or SDK == 28.
Opening another fd with contextResolver is not cheap on SAF condition, and may eat all performance benefits `ImageDecoder` give us.

revonateB0T: Although opening another fd is cheaper that "squash to direct bytebuffer" for AnimateImageDecoder, but it only happens if user disable rewrite framerate && SDK == 28, that's conner use case, when fd doesn't support seek, rewind is impossible, so AnimatedImageDecoder will load source to memory fully as well.

coil-core/src/androidMain/kotlin/coil3/decode/StaticImageDecoder.kt
```diff
@@ -106,40 +108,45 @@ class StaticImageDecoder(
...
+        }
+        metadata is ResourceMetadata && metadata.packageName == options.context.packageName -> {
+            ImageDecoder.createSource(options.context.resources, metadata.resId)
+        }
+        metadata is ByteBufferMetadata && (SDK_INT >= 30 || metadata.byteBuffer.isDirect) -> {
```
revonateB0T, line 147: For static image on SDK 29, 30
decode a `HeapByteBuffer` is safe as ImageDecoder will decode it immediately.

coil-core/src/androidMain/kotlin/coil3/decode/StaticImageDecoder.kt
colinrtwhite: Unfortunately we can't use the when subject since we have extra `&&` statements (e.g. `metadata is ResourceMetadata && metadata.packageName == options.context.packageName`).

revonateB0T (approved):
Sorry that I can't resolve these review conversation, so approve here.

coil-core/src/androidMain/kotlin/coil3/decode/StaticImageDecoder.kt
revonateB0T: Friendly remind: maybe we can use guards later.
https://github.com/Kotlin/KEEP/blob/guards/proposals/guards.md

PR #2499, opened 2024-09-26
Update utils.kt

GeorgCantor:
The `foldIndexed` function is a higher-order function that allows you to accumulate a value (in this case, the sum) while also keeping track of the index. This eliminates the need for manual index management and enhances readability.


colinrtwhite:
Thanks for creating this, though in this case I'm trying to keep the implementation as similar to `sumBy` (from the STD lib) as possible.

PR #2552, opened 2024-10-15
Update to Dokka 2.0.0.

colinrtwhite:


buildSrc/src/main/kotlin/coil3/projects.kt
```diff
@@ -30,12 +30,12 @@ fun Project.androidLibrary(
     sourceSets["main"].resources {
         srcDirs("src/commonMain/resources", "src/jvmCommonMain/resources")
     }
-    if (project.name in publicModules) {
+    if (false && project.name in publicModules) {
```
colinrtwhite, line 33: Commenting this out to apply the plugin manually for testing.

build.gradle.kts
```diff
@@ -34,10 +32,10 @@ buildscript {
 
 plugins {
     alias(libs.plugins.baselineProfile) apply false
-    alias(libs.plugins.binaryCompatibility)
-    alias(libs.plugins.dokka)
     alias(libs.plugins.poko) apply false
+    alias(libs.plugins.binaryCompatibility)
     alias(libs.plugins.spotless)
+    id("org.jetbrains.dokka")
```
chris-horner, line 38: Can we move dokka back into the `[plugins]` section of `libs.versions.toml` so this can be tidier?
```suggestion
    alias(libs.plugins.dokka)
```

chris-horner (approved):

build.gradle.kts
colinrtwhite: Unfortunately we can't due to a [gradle issue](https://github.com/gradle/gradle/issues/20084). Added a link to it!

PR #2566, opened 2024-10-18
Config nullability

Jawnnypoo:
With API 35, the nullability of config was fixed, so we can remove this elvis suppression and comment!


colinrtwhite (approved):
Nice!

Issue #2349, opened 2024-07-02
AsyncImage fills the screen if one dimension is not specified and the image is not resolved


ParsleyJ:
**Describe the bug**
The AsyncImage composable takes up all the possible space when the image (provided via URL) is not resolved.

**To Reproduce**
In the following snippet, the rendered screen should contain just a red box. Instead nothing is displayed, as the _empty_ AsyncImage takes up the full height of the screen, pushing the box to the bottom, outside the screen.

```Kotlin
@Composable
fun TestScreen() {
    Column(Modifier.fillMaxSize()) {

        AsyncImage(
            modifier = Modifier.fillMaxWidth(.5f),
            model = "https://inexistent.url/missingImage.png",
            contentDescription = "Test image",
            contentScale = ContentScale.FillWidth
        )

        Box(Modifier.background(color = Color.Red).size(50.dp))
        
    }
}
```

**Version**
kt-coil, coil-compose 2.6.0

RobertTadevosyan:
@ParsleyJ the expected result is that the async image would have a 0 height ?

ParsleyJ:
@RobertTadevosyan 

Yes

PR #2578, opened 2024-10-21
Don't fill the available space in AsyncImage when the painter has no child painter.

colinrtwhite:
Resolves: https://github.com/coil-kt/coil/issues/2349


colinrtwhite:
This creates some issues with `CrossfadePainter` that need to be resolved.

PR #2678, opened 2024-11-12
docs: add important usage note for Compose Multiplatform

AungThiha:
To help new users of coil in Compose Multiplatform find the guide easier. This can help new users avoid below problem I had.

When I first tried, I used okhttp and was stuck for a while until I carefully read the error logs and found something about compatibility issue with one of the iOS CPU architecture. Then, I started googling and found nothing. Eventually, I carefully read the documentation. Only then, I found what's wrong.

----
Anecdote aside, the text I added is basically a word for word copy-paste from the official document. The only thing I added is `> **⚠️ Important:**`

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@ Import the Compose library and a [networking library](https://coil-kt.github.io/
 implementation("io.coil-kt.coil3:coil-compose:3.0.2")
 implementation("io.coil-kt.coil3:coil-network-okhttp:3.0.2")
 ```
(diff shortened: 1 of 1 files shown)


colinrtwhite:
@AungThiha This is already mentioned at the top of the getting started docs [here](https://coil-kt.github.io/coil/getting_started/#getting-started). I'd prefer to keep the quickstart readme as simple as possible and avoid adding notes/warnings as it adds mental load.

AungThiha:
@colinrtwhite understood but it's not in the readme, which made me confused. 
If you'd like to keep readme simple, I'd like to request the removal of quickstart from readme and only link it to the docs.
I can make a PR for that if you're alright with that request.

PR #2685, opened 2024-11-14
Add test to verify DiskLruCache entries persist deletion without flush.

colinrtwhite:


colinrtwhite:
This isn't very useful for the situation we're trying to test.

PR #2699, opened 2024-11-18
Guard against Dispatchers.Main.immediate being unavailable.

colinrtwhite:
Coil relies on immediate dispatching to resolve from its memory cache synchronously. `Dispatchers.Main.immediate` is available almost everywhere except test and preview environments (e.g. Paparazzi, Roborazzi, AndroidX screenshot testing lib, and Android Studio previews). Technically the main thread can be overridden in these cases with `Dispatchers.setMain`, but we want to avoid requiring that. In those cases, fall back to `Dispatchers.Unconfined`. This works fine for most cases as Coil makes no assumptions about the context it is in and its components are thread safe except:

- Cases where we need to touch a view or lifecycle. This is only an issue with `ImageRequest`s launched off the main thread that have a `ViewTarget`.
- `AsyncImagePainter` has no problem receiving the `Image` callback on any thread, but user code may be expecting the listeners to be invoked on the main thread (even though this is not part of Coil's API contract).

As `Dispatchers.Unconfined` will only end up being used in test/non prod code I feel OK about making this change since in those cases you should also set `ImageLoader.Builder.coroutineContext(EmptyCoroutineContext)` to load images synchronously. It may make sense to set `ImageLoader.Builder.coroutineContext(EmptyCoroutineContext)` automatically if `Dispatchers.Main.immediate` can't be resolved, but I'm unsure.

### DIFF
--- a/README-ko.md
+++ b/README-ko.md
@@ -1,4 +1,4 @@
-﻿![Coil](logo.svg)
+![Coil](logo.svg)
 
 [Android](https://www.android.com/) 및 [Compose Multiplatform](https://www.jetbrains.com/lp/compose-multiplatform/)을 위한 이미지 로딩 라이브러리입니다. Coil의 특징은 다음과 같습니다.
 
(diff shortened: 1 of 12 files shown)


colinrtwhite:
This is an imperfect solution, but guards against the issue for now. We can better address it in a `3.1` release.

PR #2712, opened 2024-11-24
Corrected Korean Readme typos and made Korean more natural

WonJoongLee:
## What's Changed
- Markdown was not applied correctly.
- The code in `build.gradle.kts` was also translated, but code should not be translated.
- Additionally, some Korean text was made more natural.


colinrtwhite (approved):
Thank you!

PR #2725, opened 2024-11-29
Remove dependency on Dispatchers.Main.immediate for immediate dispatching in Compose.

colinrtwhite:
Fixes an issue with Coil's public composables where its memory cache check could miss the first frame due to dispatching. This case occurs with Paparazzi/Roborazzi/AndroidX screenshot tests and is due to the dispatcher provided by `rememberCoroutineScope()` either not supporting immediate dispatching, `Dispatchers.Main.immediate` being unavailable, or `Dispatchers.Main.immediate` using a different underlying thread.

To fix this I tried a couple things:

- Using `CoroutineStart.UNDISPATCHED` with a `StateFlow`. This works for initial composition, but does not resolve synchronously from the memory cache on subsequent requests (e.g. when the `AsyncImage(model)` changes).
- Attempting to create a `CoroutineDispatcher` that re-enables dispatching after the first time the context's `CoroutineDispatcher` changes. As far as I can tell this isn't possible at the moment with `CoroutineInterceptor`'s hooks.

I settled on using a custom `CoroutineDispatcher` that acts like `Dispatchers.Unconfined` until after the `CoroutineContext`'s dispatcher changes. This is safe for Coil's internals as it's all thread safe. Aside from fixing the above issue this has a number of benefits:

- This ensures we have consistent behaviour in all environments instead of falling back to `Dispatchers.Unconfined` if `Dispatchers.Main.immediate` is unavailable.
- `kotlinx-coroutines-swing` no longer needs to be added manually as a dependency on JVM.


coil-compose-core/src/commonMain/kotlin/coil3/compose/internal/ForwardingUnconfinedCoroutineScope.kt
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * A [CoroutineDispatcher] that delegates to [Dispatchers.Unconfined] while [unconfined] is true
+ * and [delegate] when [unconfined] is false.
+ */
+internal class ForwardingUnconfinedCoroutineDispatcher(
+    private val delegate: CoroutineDispatcher,
+) : CoroutineDispatcher() {
+    private val _unconfined = atomic(true)
+    var unconfined by _unconfined
```
zach-klippenstein, line 36: Consider documenting that after setting this to false, it's ignored even if set to true again.

coil-compose-core/src/commonMain/kotlin/coil3/compose/internal/ForwardingCoroutineContext.kt
```diff
@@ -0,0 +1,36 @@
+package coil3.compose.internal
+
+import kotlin.coroutines.CoroutineContext
+
+/**
+ * A special [CoroutineContext] implementation that lets us observe changes to its elements.
+ */
+internal class ForwardingCoroutineContext(
```
zach-klippenstein, line 8: Oo this is a neat trick.

zach-klippenstein (commented):
This is really clever, glad you were able to figure out a solution and even get rid of that swing dep!

coil-compose-core/src/commonMain/kotlin/coil3/compose/AsyncImagePainter.kt
```diff
@@ -213,22 +219,26 @@ class AsyncImagePainter internal constructor(
...
-                    if (previewHandler != null) {
-                        // If we're in inspection mode use the preview renderer.
-                        val request = updateRequest(input.request, isPreview = true)
-                        previewHandler.handle(input.imageLoader, request)
-                    } else {
-                        // Else, execute the request as normal.
-                        val request = updateRequest(input.request, isPreview = false)
-                        input.imageLoader.execute(request).toState()
+        val originalDispatcher = scope.coroutineContext.dispatcher ?: Dispatchers.Unconfined
+        val scope = ForwardingUnconfinedCoroutineScope(scope.coroutineContext)
```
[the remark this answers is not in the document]
colinrtwhite: @zach-klippenstein Great point! I reworked the implementation to hide more of the details + updated the naming to `DeferredDispatch` to make it more clear. I added `originalDispatcher` to the context since that gets passed through `launch`.

PR #2774, opened 2025-01-06
Support 0 as Expires header value in CacheControlCacheStrategy.

### LINKED ISSUE #2773: Exception when parsing Expires: 0 header
**Describe the bug**
An exception occurs when trying to parse the special `Expires: 0` header in the [`CacheControlCacheStrategy`](https://github.com/coil-kt/coil/blob/3.0.4/coil-network-cache-control/src/commonMain/kotlin/coil3/network/cachecontrol/CacheControlCacheStrategy.kt#L121). According to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Expires) `0` is allowed as a special value.

```
kotlinx.datetime.DateTimeFormatException: Failed to parse an instant from '0'
	at kotlinx.datetime.Instant$Companion.parse(Instant.kt:79)
	at coil3.network.cachecontrol.CacheControlCacheStrategy$Computation.<init>(CacheControlCacheStrategy.kt:121)
        ...
```

**To Reproduce**
Request an image with `Expires: 0` as a response header.

**Version**
Coil 3.0.4

JanMalch:
Closes #2773

### DIFF
--- a/coil-network-cache-control/src/commonMain/kotlin/coil3/network/cachecontrol/CacheControlCacheStrategy.kt
+++ b/coil-network-cache-control/src/commonMain/kotlin/coil3/network/cachecontrol/CacheControlCacheStrategy.kt
@@ -118,7 +118,10 @@ class CacheControlCacheStrategy @JvmOverloads constructor(
                         servedDateString = value
                     }
                     name.equals("Expires", ignoreCase = true) -> {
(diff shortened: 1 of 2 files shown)


colinrtwhite (approved):
Thanks! Also thanks for adding a test!

PR #2884, opened 2025-03-11
Fix transformation with default response in FakeImageLoaderEngine

carlonzo:
When using the default interceptor for a default response, the chain used was not the transformed one. This caused the Request to not have a NONE Transition.Factory, and causing tests to still running transitions.

For example this caused issues with our Paparazzi tests where, requests with default result where still having the crossfade animation causing the tests to be flaky.

### DIFF
--- a/coil-test/src/commonMain/kotlin/coil3/test/FakeImageLoaderEngine.kt
+++ b/coil-test/src/commonMain/kotlin/coil3/test/FakeImageLoaderEngine.kt
@@ -55,17 +55,17 @@ class FakeImageLoaderEngine private constructor(
 
     override suspend fun intercept(chain: Interceptor.Chain): ImageResult {
         val request = requestTransformer.transform(chain.request)
-        val requestValue = RequestValue(request, chain.size)
+        val newChain = chain.withRequest(request)
+        val requestValue = RequestValue(request, newChain.size)
         _requests.emit(requestValue)
 
         var result: ImageResult? = null
         for (interceptor in interceptors) {
-            val newChain = chain.withRequest(request)
             result = interceptor.intercept(newChain)
             if (result != null) break
         }
         if (result == null) {
-            result = defaultInterceptor.intercept(chain)
+            result = defaultInterceptor.intercept(newChain)
         }
 
         _results.emit(ResultValue(requestValue, result))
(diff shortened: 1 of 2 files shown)


coil-test/src/commonMain/kotlin/coil3/test/FakeImageLoaderEngine.kt
```diff
@@ -55,17 +55,17 @@ class FakeImageLoaderEngine private constructor(
...
     override suspend fun intercept(chain: Interceptor.Chain): ImageResult {
         val request = requestTransformer.transform(chain.request)
-        val requestValue = RequestValue(request, chain.size)
+        val newChain = chain.withRequest(request)
+        val requestValue = RequestValue(request, newChain.size)
         _requests.emit(requestValue)
 
         var result: ImageResult? = null
         for (interceptor in interceptors) {
-            val newChain = chain.withRequest(request)
```
carlonzo, line 63: honestly I am not sure why a new chain was generated for each interceptor

coil-test/src/commonMain/kotlin/coil3/test/FakeImageLoaderEngine.kt
```diff
@@ -55,17 +55,17 @@ class FakeImageLoaderEngine private constructor(
 
     override suspend fun intercept(chain: Interceptor.Chain): ImageResult {
         val request = requestTransformer.transform(chain.request)
-        val requestValue = RequestValue(request, chain.size)
+        val newChain = chain.withRequest(request)
+        val requestValue = RequestValue(request, newChain.size)
```
carlonzo, line 59: use the transformed chain size

coil-test/src/commonMain/kotlin/coil3/test/FakeImageLoaderEngine.kt
```diff
@@ -55,17 +55,17 @@ class FakeImageLoaderEngine private constructor(
...
 
         var result: ImageResult? = null
         for (interceptor in interceptors) {
-            val newChain = chain.withRequest(request)
             result = interceptor.intercept(newChain)
             if (result != null) break
         }
         if (result == null) {
-            result = defaultInterceptor.intercept(chain)
+            result = defaultInterceptor.intercept(newChain)
```
carlonzo, line 68: the actual problem was here, the intercepted chain was the original

colinrtwhite (approved):
Good find, thanks! I think you just need to run `./gradlew spotlessApply` to pass CI.

coil-test/src/commonMain/kotlin/coil3/test/FakeImageLoaderEngine.kt
colinrtwhite: Agreed - this seems like a logic mistake.

Issue #2952, opened 2025-04-24
Some images are not displayed correctly with 270 orientation


psedoykin:
**Describe the bug**
Some images are not displayed correctly with 270 orientation. If I use ImageDecoder then image is displayed incorrectly, if I set imageDecoderEnabled(false) and use  BitmapFactory it's work correctly

**To Reproduce**
```
 AsyncImage(
        modifier = modifier
            .focusRequester(focusRequester),
        imageLoader = imageLoader,
        model = ImageRequest.Builder(LocalContext.current)
            .data(uri)
            .build(),
        contentDescription = "",
        clipToBounds = false,
    )
```

**Version**
3.1.0

psedoykin:
{"Attachments":[{"__type":"ItemIdAttachment:#Exchange","ItemId":{"__type":"ItemId:#Exchange","ChangeKey":null,"Id":"AAMkADE2MGUxZTY3LWJjZWItNDIwMy1iYjU1LTRiMzBiZGViODY4ZABGAAAAAAAGUnixWem1TpT9WGbZuosIBwCMFZenI/voT62r8gARACtLAAAG6GoxAACMFZenI/voT62r8gARACtLAAExoG9wAAA="},"Name":"[Jira] (KOS-22184) [Gallery] Some images are opened in incorrect orientation","IsInline":false}]}

psedoykin:
[raw_arw_example(1).zip](https://github.com/user-attachments/files/19887344/raw_arw_example.1.zip)

Zimbelstern:
Hi @psedoykin, for someone having the same issue (wrong orientation of images), could you explain your workaround? I can set `imageDecoderEnabled(false)` in the `ImageLoader.Builder` but how to use `BitmapFactory`?

psedoykin:
> Hi [@psedoykin](https://github.com/psedoykin), for someone having the same issue (wrong orientation of images), could you explain your workaround? I can set `imageDecoderEnabled(false)` in the `ImageLoader.Builder` but how to use `BitmapFactory`?

Hi,
If you set imageDecoderEnabled(false) coil start use BitmapFactory. 

You also check it in code: 
https://github.com/coil-kt/coil/blob/782e77e0d3c1cbed2ad4d8eff57ff4dae5910b7a/coil-core/src/androidMain/kotlin/coil3/RealImageLoader.android.kt#L107

Zimbelstern:
[...]

PR #2958, opened 2025-04-27
`StaticImageDecoder` behaves like `RESPECT_PERFORMANCE`

revonateB0T:
Note that this will not resolve** https://github.com/coil-kt/coil/issues/2952
As `BitmapFactory` will not return correct mimetype for ARW image.

Fixes: https://github.com/coil-kt/coil/issues/2952


colinrtwhite (approved):
Thanks!

PR #2961, opened 2025-04-28
Use kotlinx.io's new Okio interop module.

colinrtwhite:
Need to figure out why this causes `ClosedByteChannelException` on non-JVM platforms ~and add a regression test~.

Reverts coil-kt/coil#2957

EDIT: Fixed in kotlinx.io [here](https://github.com/Kotlin/kotlinx-io/pull/453).


coil-network-ktor2/src/nonJvmCommonMain/kotlin/coil3/network/ktor2/internal/utils.nonJvmCommon.kt
```diff
@@ -14,7 +14,6 @@ internal actual suspend fun ByteReadChannel.writeTo(sink: BufferedSink) {
         val packet = readRemaining(buffer.size.toLong())
         if (packet.isEmpty) break
 
-        // TODO: Figure out how to remove 'buffer' and read directly into 'sink'.
```
colinrtwhite, line 17: Remove this since it isn't something that we're likely to do for only Ktor 2 now that Ktor 3 has a better solution.

chris-horner (approved):

Issue #2959, opened 2025-04-27
`AsyncImage` in `movableContentOf` flickering after d3694d316398602b517caa9193eb68dacc39ddb2


revonateB0T:
```
@Composable
fun Sample() {
    Box(modifier = Modifier.size(100.dp)) {
        val content = remember {
            movableContentOf {
                AsyncImage(
                    model = "https://avatars.githubusercontent.com/u/52722434",
                    contentDescription = null
                )
            }
        }
        val state by remember {
            flow {
                while (true) {
                    delay(1000)
                    emit(true)
                    delay(1000)
                    emit(false)
                }
            }
        }.collectAsState(false)
        if (state) {
            Spacer(modifier = Modifier)
            content()
        } else {
            content()
        }
    }
}
```

Related PR: #2795

andkulikov:
Yes, it might indeed be caused by #2795. Modifier nodes are detached and then attached again in movable content use case, while remembered objects are not recreated.

@colinrtwhite, I think we can try to workaround it somehow. Practically, we only need to reload an image if the input did change.  If the request for this input object was finished already, then the painter was forgotten, and then remembered again, we don't really need to reload anything. It will still work fine for list items reuse use case, as there we also reset the input - https://github.com/coil-kt/coil/blob/6f916ad77e2027795eaec447f60830b98d0dc4ca/coil-compose-core/src/commonMain/kotlin/coil3/compose/internal/ContentPainterModifier.kt#L182. Wdyt?

colinrtwhite:
@andkulikov I think that sounds good to me! Feel free to send over a PR or I can take a closer look once I'm back from vacation on May 27th.

revonateB0T:
@andkulikov Hi, any update on this? We should resolve this regression ASAP.

PR #2995, opened 2025-05-22
Implement InFlightRequestStrategy and DeDupeInFlightRequestStrategy

luciofm:
This fixes https://github.com/coil-kt/coil/issues/1461, to avoid multiple parallel network requests for the same URL.

I've basically created a simple `InFlightRequestStrategy` interface, with a `DEFAULT` implementation that does nothing.

And a concrete implementation `DeDupeInFlightRequestStrategy`, that check for a current in-flight request for the same url and waits for it.

Then on `NetworkFetcher` I've simply wrapped the `fetch()` method with `InFlightRequestStrategy.apply()`. 

With the DEFAULT implementation, it will just invoke the lambda (call doFetch()). And using the DeDupe strategy, it will check for in-flight requests and wait on it.

Note: I haven't update the `.api` files yet...

### DIFF
--- a/coil-network-core/src/commonMain/kotlin/coil3/network/InFlightRequestStrategy.kt
+++ b/coil-network-core/src/commonMain/kotlin/coil3/network/InFlightRequestStrategy.kt
@@ -1,0 +1,61 @@
+package coil3.network
+
+import coil3.fetch.FetchResult
+import kotlin.jvm.JvmField
+import kotlinx.coroutines.CompletableDeferred
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
+import okio.Closeable
+import okio.use
+
+interface InFlightRequestStrategy {
+    suspend fun apply(key: String, block: (suspend () -> FetchResult)): FetchResult {
+        return block()
+    }
+
+    companion object {
+        @JvmField
+        val DEFAULT: InFlightRequestStrategy = object : InFlightRequestStrategy {}
+    }
+}
+
+class DeDupeInFlightRequestStrategy() : InFlightRequestStrategy {
+    private val inFlightRequests = mutableMapOf<String, InFlightRequest>()
+    private val mutex: Mutex = Mutex()
+
+    override suspend fun apply(
+        key: String,
+        block: (suspend () -> FetchResult),
+    ): FetchResult {
+        var shouldWait = false
+        val request = mutex.withLock {
+            inFlightRequests[key]?.also { shouldWait = true } ?: addRequest(key)
+        }
+
+        if (shouldWait) {
+            request.deferrable.await()
+        }
+
+        return request.use {
+            block()
+        }.also {
+            mutex.withLock {
+                inFlightRequests.remove(key)
+            }
+        }
+    }
+
+    // must be called with the Mutex already acquired
+    private fun addRequest(key: String): InFlightRequest {
+        return InFlightRequest(key, CompletableDeferred()).also { inFlightRequests[key] = it }
+    }
+
+    data class InFlightRequest(
+        val key: String,
+        val deferrable: CompletableDeferred<Unit>,
+    ) : Closeable {
+        override fun close() {
+            deferrable.complete(Unit)
+        }
+    }
+}
--- a/coil-network-core/src/commonMain/kotlin/coil3/network/NetworkFetcher.kt
+++ b/coil-network-core/src/commonMain/kotlin/coil3/network/NetworkFetcher.kt
@@ -35,10 +35,17 @@ class NetworkFetcher(
     private val networkClient: Lazy<NetworkClient>,
     private val diskCache: Lazy<DiskCache?>,
     private val cacheStrategy: Lazy<CacheStrategy>,
+    private val inFlightRequestStrategy: Lazy<InFlightRequestStrategy>,
     private val connectivityChecker: ConnectivityChecker,
 ) : Fetcher {
 
     override suspend fun fetch(): FetchResult {
+        return inFlightRequestStrategy.value.apply(diskCacheKey) {
+            doFetch()
+        }
+    }
+
+    private suspend fun doFetch(): FetchResult {
         var snapshot = readFromDiskCache()
         try {
             // Fast path: fetch the image from the disk cache without performing a network request.
@@ -262,10 +269,12 @@ class NetworkFetcher(
     class Factory(
         networkClient: () -> NetworkClient,
         cacheStrategy: () -> CacheStrategy = { CacheStrategy.DEFAULT },
+        inFlightRequestStrategy: () -> InFlightRequestStrategy = { InFlightRequestStrategy.DEFAULT },
         connectivityChecker: (PlatformContext) -> ConnectivityChecker = ::ConnectivityChecker,
     ) : Fetcher.Factory<Uri> {
         private val networkClientLazy = lazy(networkClient)
         private val cacheStrategyLazy = lazy(cacheStrategy)
+        private val inFlightRequestStrategyLazy = lazy(inFlightRequestStrategy)
         private val connectivityCheckerLazy = singleParameterLazy(connectivityChecker)
 
         override fun create(
@@ -280,6 +289,7 @@ class NetworkFetcher(
                 networkClient = networkClientLazy,
                 diskCache = lazy { imageLoader.diskCache },
                 cacheStrategy = cacheStrategyLazy,
+                inFlightRequestStrategy = inFlightRequestStrategyLazy,
                 connectivityChecker = connectivityCheckerLazy.get(options.context),
             )
         }
--- a/coil-network-core/src/commonTest/kotlin/coil3/network/NetworkFetcherTest.kt
+++ b/coil-network-core/src/commonTest/kotlin/coil3/network/NetworkFetcherTest.kt
@@ -45,6 +45,7 @@ class NetworkFetcherTest : RobolectricTest() {
             diskCache = lazyOf(null),
             cacheStrategy = lazyOf(CacheStrategy.DEFAULT),
             connectivityChecker = ConnectivityChecker(context),
+            inFlightRequestStrategy = lazyOf(InFlightRequestStrategy.DEFAULT)
         ).fetch()
 
         assertIs<SourceFetchResult>(result)
--- a/coil-network-ktor2/src/commonMain/kotlin/coil3/network/ktor2/KtorNetworkFetcher.kt
+++ b/coil-network-ktor2/src/commonMain/kotlin/coil3/network/ktor2/KtorNetworkFetcher.kt
@@ -5,6 +5,7 @@ package coil3.network.ktor2
 import coil3.PlatformContext
 import coil3.network.CacheStrategy
 import coil3.network.ConnectivityChecker
+import coil3.network.InFlightRequestStrategy
 import coil3.network.NetworkClient
 import coil3.network.NetworkFetcher
 import coil3.network.ktor2.internal.KtorNetworkClient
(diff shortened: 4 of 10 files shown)


coil-network-core/src/commonMain/kotlin/coil3/network/InFlightRequestStrategy.kt
```diff
@@ -0,0 +1,61 @@
...
+class DeDupeInFlightRequestStrategy() : InFlightRequestStrategy {
+    private val inFlightRequests = mutableMapOf<String, InFlightRequest>()
+    private val mutex: Mutex = Mutex()
+
+    override suspend fun apply(
+        key: String,
+        block: (suspend () -> FetchResult),
+    ): FetchResult {
+        var shouldWait = false
+        val request = mutex.withLock {
```
revonateB0T, line 31: I suggest to use non-suspend mutex here as critical region is small & non-suspend.

coil-network-core/src/commonMain/kotlin/coil3/network/InFlightRequestStrategy.kt
```diff
@@ -0,0 +1,61 @@
...
+            inFlightRequests[key]?.also { shouldWait = true } ?: addRequest(key)
+        }
+
+        if (shouldWait) {
+            request.deferrable.await()
+        }
+
+        return request.use {
+            block()
+        }.also {
```
revonateB0T, line 41: Given we launch 2 same requests A, B. A is active, B is waiting for A.
What about A is cancelled? we doesn't remove it from inFlightRequests here.
Moreover, now that B is not cancelled, should we cancel & relaunch network request here? it does harm performance.

revonateB0T:
I suggest to look into our `MergeInteceptor` https://github.com/FooIbar/EhViewer/blob/1548e16fbe7250785d5d13e55615131777ac11b6/app/src/main/kotlin/com/hippo/ehviewer/coil/MergeInterceptor.kt#L34

It handles same-url-request debounce well using kotlin flow.
Given request order:
Launch A -> Launch B -> Cancel A
where A, B use same URL, launched network request is executed as is.

Given request order:
Launch A -> Cancel A -> Launch B
where A, B use same URL, delay 100ms on each operation, launched network request is executed as is, we avoid relaunch request with a 100ms cancel-delay trade-off.

coil-network-core/src/commonMain/kotlin/coil3/network/InFlightRequestStrategy.kt
luciofm: Fair enough.

coil-network-core/src/commonMain/kotlin/coil3/network/InFlightRequestStrategy.kt
revonateB0T: > If A is cancelled, the request will be removed from the `inFlightRequests` map and B will proceed...
> 
> If there are more requests waiting, on this particular case, all of them will trigger a network request.

But we did't remove it, the `also` block is not getting executed due to `CancellationException`, it's not wrapped by FetchResult.

luciofm: If the request is canceled, the Deferrable will be completed, and will not block other requests form going through, but I'll move to a try/finally to remove the request from the map.

luciofm: I wanted to make it really simple, and as less invasive as possible on the current code/functionality while solving the most common use cases...

Lets see what @colinrtwhite has to say about this

revonateB0T: > I wanted to make it really simple, and as less invasive as possible on the current code/functionality while solving the most common use cases...
> 
> Lets see what @colinrtwhite has to say about this

He's on vacation https://github.com/coil-kt/coil/issues/2959#issuecomment-2875068132

luciofm: Maybe we could have 2 implementations, the current one proposed on the PR renamed as `SimpleInFlightRequestStrategy`, a simple strategy that handles the most common cases:
- The first network request succeeds and all other waiters are unblocked to hit the disk cache in parallel
- All queued network requests for the same URL would fail. We unblock all waiting requests to fail in parallel.

And have a second, more complex, implementation
-  One request is allowed to go through, all others are waiting on a `Channel<Unit>`
  - The request succeeds:
    - The channel is closed, and all waiting coroutines are allowed to run in parallel, hitting the disk cache.
  - The request fails:
    - Tries to send on the channel, if fails (no one waiting), close the channel and remove the `InFlightRequest` from the map
    - If send is successful, this means another coroutine will wake up and run the request, this new coroutine will now be the channel owner/controller and responsible for the cleanup. Rinse and repeat.

```kotlin
class DeDupeInFlightRequestStrategy() : InFlightRequestStrategy {
    private val inFlightRequests = mutableMapOf<String, InFlightRequest>()
    private val lock = SynchronizedObject()

    @OptIn(ExperimentalAtomicApi::class)
    override suspend fun apply(
        key: String,
        block: (suspend () -> FetchResult),
    ): FetchResult {
        // This coroutine found an existing request and should attempt to wait on its channel.
        var shouldWaitForSignal = false
        // This coroutine is controlling the channel. This can be the first coroutine for the key
        // or one that received the baton after a prior failure
        var currentChannelController = false
        // Is this channel controller also responsible for cleaning up the channel and map entry.
        // A controller might not be responsible for cleanup if it successfully passes the baton to
        // another waiter upon its own failure.
        var isResponsibleForCleanup = false
        
        val request = synchronized(lock) {
            inFlightRequests[key]?.also {
                // An InFlightRequest already exists for this key. This coroutine must wait.
                shouldWaitForSignal = true
            } ?: addRequest(key).also {
                // No existing request. This coroutine is the first and the channel controller
                currentChannelController = true
                // Initially responsible for cleanup.
                isResponsibleForCleanup = true
            }
        }

        if (shouldWaitForSignal) {
            // This coroutine is a waiter. It suspends until the channel receives or is closed.
            // Receiving Unit means a previous executor failed, and this waiter should take over.
            if (request.channel.receiveCatching().getOrNull() != null) {
                // Successfully received a signal (Unit). This waiter is now the channel controller.
                currentChannelController = true
                // And also responsible for cleaning up after success.
                isResponsibleForCleanup = true
            } else {
                // Channel was closed (by a successful executor or the last in a chain of failures).
                // This waiter will proceed to run block() hoping for a cache hit.
                // It's not the channel controller for this attempt post-signal, nor
                // responsible for cleanup.
            }
        }
        
        // If !currentChannelController at this point, it means:
        // 1. This coroutine was a waiter.
        // 2. It didn't receive a direct signal (Unit) to become the next executor
        //    (channel was likely closed).
        // 3. It will run block() (e.g., to hit cache) but won't manage the channel or map entry.
        try {
            return block()
        } catch (ex: Exception) {
            // This coroutine was the designated executor and its block() failed.
            // Attempt to pass execution responsibility (baton) to another waiting coroutine.
            val successfullyPassedBaton = request.channel.trySend(Unit).isSuccess
            if (successfullyPassedBaton) {
                // Baton was passed. This coroutine is no longer the channel controller nor 
                // responsible for cleanup.
                currentChannelController = false
                isResponsibleForCleanup = false
            } else {
                // The baton was not passed (no waiter or channel closed), this coroutine remains
                // the channel controller and responsible for the cleanup in the finally block.
            }
            throw ex
        } finally {
            if (currentChannelController && isResponsibleForCleanup) {
                // This coroutine was the channel owner AND is responsible for cleanup.
                // This happens if:
                // 1. It executed block() successfully.
                // 2. It executed block(), failed, AND could not pass the baton to a waiter.
                synchronized(lock) {
                    request.channel.close()
                    inFlightRequests.remove(key)
                }
            }
        }
    }

    // must be called with the lock already acquired
    private fun addRequest(key: String): InFlightRequest {
        return InFlightRequest(key).also { inFlightRequests[key] = it }
    }

    data class InFlightRequest(
        val key: String,
        val channel: Channel<Unit> = Channel()
    )
}
```

coil-network-core/src/commonMain/kotlin/coil3/network/InFlightRequestStrategy.kt
```diff
@@ -0,0 +1,173 @@
...
+            if (channel.receiveCatching().getOrNull() != null) {
+                currentChannelController = true
+                isResponsibleForCleanup = true
+            }
+        }
+
+        try {
+            return block()
+        } catch (ex: Exception) {
+            val successfullyPassedBaton = channel.trySend(Unit).isSuccess
```
colinrtwhite, line 159: Will this launch a new request if the first request is cancelled and it falls back to a second observer? I'm wondering if we can create an implementation that refcounts the observers and only cancels the request if the refcount drops to 0.

colinrtwhite (commented):
@luciofm Thanks for working on this! I left a few comments, but I think these are the main things we should focus on/think about:

- `InFlightRequestStrategy.apply` currently couples us to the disk cache key. We might not be able to change this especially if `block` must capture the disk cache reads as well (instead of only the network request).
- `InFlightRequestStrategy.apply` `block` captures `Option`, which has a reference to the locally-scope `Context`. This could leak a `Fragment`/`Activity` if `apply` is cancelled, but `block` isn't.
- I think we probably only want to ship 2 request strategies - not 3: `InFlightRequestStrategy.DEFAULT` (i.e. no deduping) and ideally a version of deduping that uses refcounting and only executes the network request exactly once (instead of passing the baton on a failure).
- We'll need to preserve the existing functions/constructors that don't have `InFlightRequestStrategy` arguments. This is to avoid breaking binary compatibility.

coil-network-core/src/commonMain/kotlin/coil3/network/InFlightRequestStrategy.kt
```diff
@@ -0,0 +1,173 @@
...
+import kotlin.jvm.JvmField
+import kotlinx.atomicfu.locks.SynchronizedObject
+import kotlinx.atomicfu.locks.synchronized
+import kotlinx.coroutines.CompletableDeferred
+import kotlinx.coroutines.channels.Channel
+import okio.Closeable
+import okio.use
+
+interface InFlightRequestStrategy {
+    suspend fun apply(key: String, block: (suspend () -> FetchResult)): FetchResult {
```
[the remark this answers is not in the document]
luciofm: For now I think we could add a this limitation to the `InFlightRequestStrategy` documentation, but creating a `diskCacheKeyFactory` makes sense for these cases...

About the leaking `Context`, I've added a new `safeDiskCacheKey` that copies `options.diskCacheKey` if available, or just use the `url`.

coil-network-core/src/commonMain/kotlin/coil3/network/InFlightRequestStrategy.kt
luciofm: This works like a serial queue until `block()` succeeds/returns. 
If block throws - either cancelation or it fails (404 for example), it will awake exactly one waiter to try again.
If block succeeds, it will close the channel, releasing all waiters to proceed.

luciofm:
I think I've fixed all/most your comments, specially the binary compatibility ones.

> * I think we probably only want to ship 2 request strategies - not 3: `InFlightRequestStrategy.DEFAULT` (i.e. no deduping) and ideally a version of deduping that uses refcounting and only executes the network request exactly once (instead of passing the baton on a failure).
Agree with having just a DEFAULT that does nothing and one implementation... I've moved the `SimpleInFlightRequestStrategy` as a sample on `InFlightRequestStrategy` doc.

I've added a new implementation for discussion (we should keep just one in the end). This one has an optional delay in removing the request key from the `inFlightRequests` map, to avoid creating new channels on scrolling for example. If you don't think this is a good idea, we can just remove the delay, but I think this implementation is cleaner than the first one... It works basically the same.

luciofm:
Closing, this got implemented with #3326 

Thanks @colinrtwhite ❤️

Issue #3009, opened 2025-06-11
Some SVG images are not loaded properly


kitswas:
**Describe the bug**

Badges from Shields.io with a logo are missing the logo.

Only a warning is shown in LogCat.
```
Kumiho-Kumiho           W  isSupportedFormat: Format is not supported
```

Or, with explicit `SvgDecoder`:
```
HWUI                    D  --- Failed to create image decoder with message 'unimplemented'
```

**To Reproduce**

Try to load: (The Github logo will be missing)

[![badge](https://img.shields.io/github/stars/kitswas/VirtualGamePad?style=social)](https://img.shields.io/github/stars/kitswas/VirtualGamePad?style=social)

OR

[![badge](https://img.shields.io/github/v/release/kitswas/VirtualGamePad-PC?logo=github)](https://img.shields.io/github/v/release/kitswas/VirtualGamePad-PC?logo=github)

```kotlin
        AsyncImage(
            model = ImageRequest.Builder(LocalContext.current)
                .data(imageUrl)
                .decoderFactory(SvgDecoder.Factory())
                .crossfade(true)
                .build(),
            contentDescription = contentDescription,
            contentScale = ContentScale.FillHeight,
            modifier = Modifier
                .height(25.dp),
            onError = { error ->
                Log.e(
                    TAG,
                    "Failed to load badge image: ${error.result.throwable.message}",
                    error.result.throwable
                )
            }
        )
```

**Version**

3.2.0 (Latest at the time of reporting)

colinrtwhite:
This is a rendering issue with the underlying SVG parser that we use, [`androidsvg`](https://github.com/BigBadaboom/androidsvg). Unfortunately it's not possible for us to fix or work-around in Coil; please report the issue to their tracker.

In Coil `3.3.0` it's now possible to replace the underlying parser via the `Svg.Parser` interface so you can swap the parser with a different library.

colinrtwhite:
I wrote up some thoughts of how we could work around `androidsvg`'s limitations [here](https://github.com/coil-kt/coil/issues/3090). Hopefully this can be a better long term solve than depending on `androidsvg`.

Issue #2924, opened 2025-04-04
Few SVGs not loading in Compose Multiplatform iOS


johnskoshy:
**Describe the bug**
Few svgs not loading in iOS

<img width="431" alt="Image" src="https://github.com/user-attachments/assets/1cc653bc-6605-486c-bd15-fbc573b7d931" />

Logs:
Successful (NETWORK) - https://cdn.coinranking.com/jkDf8sQbY/usdc.svg
Successful (NETWORK) - https://cdn.coinranking.com/dSCnSLilQ/hedera.svg
Successful (NETWORK) - https://cdn.coinranking.com/BUvPxmc9o/ltcnew.svg

**To Reproduce**

Code snippet:

AsyncImage(
    model = ImageRequest.Builder(LocalPlatformContext.current)
        .data(coin.iconUrl)
        .build(),
    contentDescription = null,
    contentScale = ContentScale.Fit,
    modifier = Modifier.padding(4.dp).clip(CircleShape).size(40.dp)
)

Icons not loading: 
https://cdn.coinranking.com/jkDf8sQbY/usdc.svg
https://cdn.coinranking.com/dSCnSLilQ/hedera.svg
https://cdn.coinranking.com/BUvPxmc9o/ltcnew.svg

**Version**
coil-compose = "3.1.0"

coil-compose = { module = "io.coil-kt.coil3:coil-compose", version.ref = "coil-compose" }
coil-compose-core = { module = "io.coil-kt.coil3:coil-compose-core", version.ref = "coil-compose" }
coil-network-ktor = { module = "io.coil-kt.coil3:coil-network-ktor3", version.ref = "coil-compose" }
coil-core = { module = "io.coil-kt.coil3:coil", version.ref = "coil-compose" }
coil-svg = { module = "io.coil-kt.coil3:coil-svg", version.ref = "coil-compose" }

Tested on iPhone 16 Pro iOS 18.4

colinrtwhite:
Similar issue to https://github.com/coil-kt/coil/issues/2895. This is a rendering issue with Skia's SVG support. You'll need to file an issue at the [Skia issue tracker](https://issues.skia.org/issues).

Issue #2895, opened 2025-03-20
SVG classes not working properly on iOS


OskarPersson:
I have these two SVGs

class.svg

```svg
<svg xmlns="http://www.w3.org/2000/svg" id="svg1" version="1.1" viewBox="0 0 568.9 556.6"><defs><style>.cls-13{fill:none; stroke-width:4px; stroke:#000}</style></defs><path id="path15" d="M21.3 224.2c9.7-.9 21.9-5.6 31-8.7 5.2-2.6 15.4-7.8 20.6-10.5l10.5-1.7 10.3-3.5 10.3-1.7 10.3-1.7 10.3-3.5 10.5-1.7h51.6l10.3-1.7 10.3-3.5 10.5-1.7 10.3-3.5 10.3-1.7c5.1-1.8 15.4-5.4 20.6-7.1l10.5-5.2 10.3-1.7 10.3-1.7 10.3-3.5 10.5-1.7 10.3-3.5 10.3-1.7c12.7-2.9 28.4-1.4 41.3-1.7h61.9" class="cls-13"/></svg>
```

inline.svg

```svg
<svg xmlns="http://www.w3.org/2000/svg" id="svg1" version="1.1" viewBox="0 0 568.9 556.6"><path id="path15" d="M21.3 224.2c9.7-.9 21.9-5.6 31-8.7 5.2-2.6 15.4-7.8 20.6-10.5l10.5-1.7 10.3-3.5 10.3-1.7 10.3-1.7 10.3-3.5 10.5-1.7h51.6l10.3-1.7 10.3-3.5 10.5-1.7 10.3-3.5 10.3-1.7c5.1-1.8 15.4-5.4 20.6-7.1l10.5-5.2 10.3-1.7 10.3-1.7 10.3-3.5 10.5-1.7 10.3-3.5 10.3-1.7c12.7-2.9 28.4-1.4 41.3-1.7h61.9" style="fill:none; stroke-width:4px; stroke:#000"/></svg>
```

The only difference between them is that the first one uses a class for its styling and the second one has the styling inline. On Android they render the same:

<img src="https://github.com/user-attachments/assets/a268d71c-0739-41d5-9bad-9cfcde209b98" width="200">

But on iOS the class one does not render like it should. It seems to ignore the `fill: none` attribute:

<img src="https://github.com/user-attachments/assets/b201ffa5-ceb1-469e-93f7-8ac1ba0578f1" width="200">

[...]

Issue #2853, opened 2025-02-20
Update Coil svg decoder


T8RIN:
Currently, it relies on android svg, but there is some issue that svgs cannot be parsed with some attributes, you can see them here

https://github.com/deckerst/aves/issues/1428

https://github.com/T8RIN/ImageToolbox/issues/1697

deckerst fixed those issues in his fork of android svg

colinrtwhite:
It looks like his fork is currently [only published to Jitpack](https://github.com/deckerst/aves/blob/5c297c1daf9a22c82b6b36e51d3ddcdbe6c5f720/android/app/build.gradle#L175), which we can't use as every Coil user would need to add a dependency on Jitpack, which is notoriously flaky.

@deckerst is this fork only for internal use in Aves? If it's for public use do you have plans to upload it to Maven Central and maintain it long term? Migrating Coil to use a fork of the android svg project is a big change that'll have downstream effects on all of Coil's users.

deckerst:
Sorry, it's not intended for public use, as I can't dedicate time to other projects.

The fix I did was simple though, so I could make a PR, then people could lobby the original owner for an official release (they seem to maintain the repo but not release anymore).

T8RIN:
That would be great, cause forking full code of coil-svg and android svg just to add one || check sounds not like a good idea :)

colinrtwhite:
@deckerst Sounds good 👍🏻. It looks like the author has [mentioned](https://github.com/coil-kt/coil/blob/main/coil-svg/src/commonMain/kotlin/coil3/svg/internal/Svg.kt#L9) plans to do a release in the future, but it's not clear on what timeline.

@T8RIN I think your best bet for now is probably to use Gradle's [dependency substitution](https://docs.gradle.org/current/userguide/resolution_rules.html#sec:dependency-substitution-rules) to replace `com.caverock:androidsvg-aar:1.4` with `com.github.deckerst:androidsvg:cc9d59a88f`. That should be OK as they appear to have the same public API.

Ultimately I'd like to migrate away from `com.caverock:androidsvg-aar`, however there don't appear to be any better alternatives that don't include native libraries.

T8RIN:
Thanks!

zhtw1234-source:
？

PR #3012, opened 2025-06-13
Create a public SVG parser interface to enable custom SVG parsing/rendering implementations.

colinrtwhite:
A common issue we get is that an SVG doesn't render correctly on Android or other platforms (e.g. [one](https://github.com/coil-kt/coil/issues/3009), [two](https://github.com/coil-kt/coil/issues/2924), [three](https://github.com/coil-kt/coil/issues/2895), [four](https://github.com/coil-kt/coil/issues/2853)). Especially on Android, [the SVG parser we use](https://github.com/BigBadaboom/androidsvg) hasn't had a new release in a long time and has many rendering fixes on master that haven't been released yet.

Currently we don't have a great solution to these problems in Coil as there's no way for users to use a custom SVG parser if default ones don't work. This PR supports passing in a custom `Svg.Parser` to `SvgDecoder` so users will now be able to hook up their own parser instead of having to run a fully custom `Decoder`, which is more work.


coil-svg/src/commonMain/kotlin/coil3/svg/Svg.kt
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * Represents an SVG with mutable properties.
+ *
+ * NOTE: This interface is provided to enable custom SVG parsing/rendering implementations, however
+ * it intentionally only provides hooks for the properties Coil uses. Additionally, these classes
+ * are very likely to change in the future.
+ */
+@ExperimentalCoilApi
+interface Svg {
```
colinrtwhite, line 18: I'm not stoked about the API on this class (especially the getters and setters), but it should do the job for now.

colinrtwhite:
Requested Copilot as well as I want to kick the tires on it 😅

PMARZV:
do non-android platform scale with density?? I say it because i dont think so and devs can modify density (for desktop especially). it would be good to be able to pass a value of density scaling (maybe it is not needed but im not 100% sure)

colinrtwhite:
@PMARZV Could you open a separate enhancement request for that please?

PR #3046, opened 2025-07-02
Add request transformation example to Recipes

afollestad:
A relatively simple example that shows how we can add `width` & `height` query parameters to HTTP request URLs.

### DIFF
--- a/docs/recipes.md
+++ b/docs/recipes.md
@@ -129,3 +129,56 @@ AsyncImage(
     },
 )
 ```
+
+## Transforming Requests
+
+You may need to transform HTTP requests used to fetch images. In this example, we'll use an [Interceptor](https://coil-kt.github.io/coil/api/coil-core/coil3.intercept/-interceptor) to append `width` & `height` query parameters to the request URL.
+
+```kotlin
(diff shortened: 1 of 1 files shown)


docs/recipes.md
```diff
@@ -129,3 +129,56 @@ AsyncImage(
...
+                )
+                .build()
+            return chain.withRequest(transformedRequest).proceed()
+        } else {
+            // Width & height aren't available, i.e. because of infinite constraints.
+            chain.proceed()
+        }
+    }
+
+    private val ImageRequest.uri: Uri?
```
afollestad, line 166: Is there a multi-platform friendly way to add query parameters to a `coil3.Uri`?

colinrtwhite: Hmm there isn't at the moment, but we should consider adding one in a separate PR

colinrtwhite (approved):
Awesome! Thanks for adding this

docs/recipes.md
afollestad: Filed https://github.com/coil-kt/coil/issues/3047

afollestad:
@colinrtwhite you'll have to merge this since I don't have permissions 😄

PR #3054, opened 2025-07-04
Add native Linux support

Thomas-Vos:
This PR adds support for the native Linux target.

Unfortunately, compose does not currently support Linux, but skiko does support it.


gradle/libs.versions.toml
```diff
@@ -15,7 +15,7 @@ okhttp = "4.12.0"
 okio = "3.15.0"
 paparazzi = "1.3.5"
 roborazzi = "1.45.1"
-skiko = "0.9.4.2"
+skiko = "0.9.21"
```
colinrtwhite, line 18: Is updating Skiko required to support the new target? We keep our Skiko version the same as the latest stable Compose to prevent issues

Thomas-Vos:
Thanks for merging the PR! Once compose supports native Linux as well, I'll create another PR to add it to coil. But it will take some time: https://github.com/JetBrains/compose-multiplatform-core/pull/2027

PR #3065, opened 2025-07-11
Support setting a custom main coroutine context for tests.

colinrtwhite:
A common issue we've run into is "why isn't this running synchronously" - especially in tests. Ideally Coil should handle these cases automatically if possible, but we can also provide a self-serve way to work-around any issues.

This PR adds an API for Coil's main thread usage so `Dispatchers.Main.immediate` can be replaced in tests without having to use `Dispatchers.setMain()`. This should help with test frameworks that use a different main thread/handler than the one `Dispatchers.Main.immediate` knows about. Example usage:

```kotlin
val fakeImageLoader = ImaageLoader.Builder(context)
    .mainCoroutineContext { UnconfinedTestDispatcher() }
    .build()
```


theisenp (approved):

coil-core/src/commonMain/kotlin/coil3/ImageLoader.kt
```diff
@@ -213,6 +218,32 @@ interface ImageLoader {
...
+         * The [CoroutineContext] that any UI/main thread work will be executed in.
+         *
+         * NOTE: This should typically only be used in tests, as `View`s and `Lifecycle`s will be
+         * accessed in this context and those operations will throw if not performed on a valid
+         * thread.
+         *
+         * Default: `Dispatchers.Main.immediate`
+         */
+        @ExperimentalCoilApi
+        fun mainCoroutineContext(initializer: () -> CoroutineContext) = apply {
```
colinrtwhite, line 243: We need to have a lazy method as `Dispatchers.main.immediate` can throw on desktop in certain (expected) cases.

Issue #3024, opened 2025-06-29
Allow passing custom density to new Svg interface


PMARZV:
**Is your feature request related to a problem? Please describe.**
After [New Svg Interface](https://github.com/coil-kt/coil/pull/3012) merges, it would be nice to have a way to pass a custom density to the Svg Parser interface to be able to scale svg to density on nonAndroid platforms. In desktop targets, devs often change the default density to scale it down since some components are way too big for desktop usage. The downside is that Svg parsed in Coil does not take that density into account (since its just drawing into bitmap, if my understanding is correct)

**Describe the solution you'd like**
Let us have a way to pass density for SvgParser consumption (which i see that nonAndroid platforms do not have the extension density, and the check done in default parser only applies to Android)

PR #3081, opened 2025-07-16
Add density function to SvgDecoder.

colinrtwhite:
Resolves: https://github.com/coil-kt/coil/issues/3024

### DIFF
--- a/coil-svg/src/commonMain/kotlin/coil3/svg/SvgDecoder.kt
+++ b/coil-svg/src/commonMain/kotlin/coil3/svg/SvgDecoder.kt
@@ -1,6 +1,7 @@
 package coil3.svg
 
 import coil3.ImageLoader
+import coil3.PlatformContext
 import coil3.asImage
 import coil3.decode.DecodeResult
 import coil3.decode.DecodeUtils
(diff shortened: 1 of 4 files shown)


coil-svg/src/commonMain/kotlin/coil3/svg/SvgDecoder.kt
```diff
@@ -27,21 +29,21 @@ import okio.use
...
  *  SVG will be rendered at draw time. Rendering at draw time is more memory efficient, but
  *  depending on the complexity of the SVG, can be slow.
- * @param scaleToDensity If true and the target size is [Size.ORIGINAL], multiplies the source
- *  dimensions of the image by the device's display density. NOTE: This only affects Android.
  */
 class SvgDecoder(
     private val source: ImageSource,
     private val options: Options,
     val parser: Svg.Parser = Svg.Parser.DEFAULT,
+    val density: PlatformContext.() -> Float = NO_DENSITY,
```
colinrtwhite, line 44: We don't need to maintain binary compatibility here as the new constructor with `parser` hasn't been released yet.

chris-horner (approved):

Issue #3092, opened 2025-07-25
Use embedded thumbnails from video files if available


hardcore-sushi:
**Is your feature request related to a problem? Please describe.**

When creating a thumbnail from a video, the video frames must be decoded, possibly from the middle of the file, which may take some time, especially if the file needs to be fetched from the network and/or decrypted. Also, not displaying the same thumbnails as other file explorers might be confusing.

**Describe the solution you'd like**
The API could offer an option to use the embedded thumbnail provided by the video file metadata, and fallback to the existing image extraction method if not available.

**Additional context**

[MediaMetadataRetriever.getEmbeddedPicture()](https://developer.android.com/reference/android/media/MediaMetadataRetriever#getEmbeddedPicture()) could be used. Example usage: https://developer.android.com/social-and-messaging/guides/media-thumbnails#create-thumbnail

I'd like to use this feature for video thumbnails in the DroidFS file explorer.

colinrtwhite:
This sounds like a good addition to me :+1:. We could add something like `ImageRequest.Builder.preferVideoFrameEmbeddedThumbnail(true)` or something similar - similar to how other `ImageRequest` [options are passed through](https://github.com/coil-kt/coil/blob/main/coil-video/src/main/java/coil3/video/imageRequests.kt#L23). PR welcome!

colinrtwhite:
Resolved by https://github.com/coil-kt/coil/pull/3107.

PR #3107, opened 2025-08-01
Use embedded thumbnails from video files if available (issue-3092)

Khudoyshukur:
Please note that after renovate gradle upgrade, the project is not building in Android Studio
image

I have run the './test.sh' with downgraded gradle config (gradle version 8.11.1), and everything is fine


Khudoyshukur:
This PR addresses issue #3092

coil-video/src/main/java/coil3/video/VideoFrameDecoder.kt
```diff
@@ -97,30 +98,49 @@ class VideoFrameDecoder(
...
+                rawBitmap = BitmapFactory.decodeByteArray(
+                    embeddedPicture,
+                    0,
+                    embeddedPicture.size
+                )?.also {
+                    srcWidth = it.width
+                    srcHeight = it.height
+                }
+            } catch (throwable: Throwable) {
+                // we encountered an error while trying to decode the embedded thumbnail.
```
colinrtwhite, line 114: I think we should probably shouldn't catch this here. If the user opted into embedded thumbnails and the decoder fails to decode it, we should propagate up the exception. The use can disable `preferVideoFrameEmbeddedThumbnail` and retry if they want.

coil-video/src/androidTest/java/coil3/video/VideoFrameDecoderTest.kt
```diff
@@ -219,4 +219,107 @@ class VideoFrameDecoderTest {
         val expected = context.decodeBitmapAsset("video_frame_1.jpg")
         actual.assertIsSimilarTo(expected)
     }
+
+    @Test
+    fun embeddedThumbnailPresentAndEmbeddedThumbnailOptionEnabled() = runTest {
```
colinrtwhite, line 224: Thanks for adding tests!

colinrtwhite (approved):
Looks great! Left one comment, but otherwise we're good to merge once tests pass. It looks like they fail on an API 26 emulator:

```
coil3.video.VideoFrameDecoderTest > embeddedThumbnailNotPresentAndEmbeddedThumbnailOptionEnabled[emulator-5554 - 8.0.0] FAILED 
	java.lang.IllegalStateException: The images are not visually similar. Expected: 0.98; Actual: -0.41216287859559503.
	at coil3.test.utils.BitmapsKt.assertIsSimilarTo(bitmaps.kt:54)

coil3.video.VideoFrameDecoderTest > specificFrameIndex[emulator-5554 - 8.0.0] SKIPPED 

coil3.video.VideoFrameDecoderTest > embeddedThumbnailNotPresentAndEmbeddedThumbnailOptionDisabled[emulator-5554 - 8.0.0] FAILED 
	java.lang.IllegalStateException: The images are not visually similar. Expected: 0.98; Actual: -0.41216287859559503.
	at coil3.test.utils.BitmapsKt.assertIsSimilarTo(bitmaps.kt:54)

coil3.video.VideoFrameDecoderTest > embeddedThumbnailPresentAndEmbeddedThumbnailOptionDisabled[emulator-5554 - 8.0.0] FAILED 
	java.lang.IllegalStateException: The images are not visually similar. Expected: 0.98; Actual: -0.2931417502459996.
	at coil3.test.utils.BitmapsKt.assertIsSimilarTo(bitmaps.kt:54)
```

coil-video/src/main/java/coil3/video/VideoFrameDecoder.kt
Khudoyshukur: Done

Khudoyshukur:
@colinrtwhite 

The problem is that I was using [.set(Extras.Key.videoFrameIndex, 807)] which is available in api 28 and above.
Changed the logic to use [Extras.Key.videoFramePercent] in tests

PR #3118, opened 2025-08-08
Fix incorrect class name in recipes.md

Johnett:
The documentation in `recipes.md`(section:- https://github.com/coil-kt/coil/blob/main/docs/recipes.md#transforming-requests) references `FastlyCoilInterceptor`, but the actual class defined in the example is `UrlSizeInterceptor`.

This PR updates the reference to match the correct class name.


colinrtwhite (approved):
Thanks!

PR #3134, opened 2025-08-21
Sync the Visible property to the child drawable

liujingxing:
I customize Decoder to load webp Animated via [APNG4Android](https://github.com/penfeizhou/APNG4Android), it will automatically stop animation when drawable is not visible, but when I set crossfade to true globally and drawable is not visible, the animation will never stop


colinrtwhite (approved):
Thanks!

PR #3138, opened 2025-08-22
Relax JVM 11 bytecode requirement for Android target.

colinrtwhite:
I thought this was required by Compose 1.8.0 for all targets, however it turns out it's only required for `jvmMain` (desktop). We can go back to JVM 8 bytecode for Android.

### DIFF
--- a/buildSrc/src/main/kotlin/coil3/utils.kt
+++ b/buildSrc/src/main/kotlin/coil3/utils.kt
@@ -2,6 +2,9 @@ package coil3
 
 import kotlin.math.pow
 import org.gradle.api.Project
+import org.gradle.kotlin.dsl.withType
+import org.jetbrains.kotlin.gradle.dsl.JvmTarget
+import org.jetbrains.kotlin.gradle.tasks.KotlinJvmCompile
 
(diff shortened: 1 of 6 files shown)


buildSrc/src/main/kotlin/coil3/utils.kt
```diff
@@ -49,6 +52,20 @@ val Project.versionCode: Int
 val Project.enableComposeMetrics: Boolean
     get() = booleanProperty("enableComposeMetrics") { false }
 
+// Compose 1.8.0 requires JVM 11 only for the JVM target.
+fun Project.applyJvm11OnlyToJvmTarget() {
+    tasks.withType<KotlinJvmCompile>().configureEach {
+        when {
+            name.contains("kotlinjvm", ignoreCase = true) -> {
+                compilerOptions.jvmTarget.set(JvmTarget.JVM_11)
```
colinrtwhite, line 60: Setting `compilerOptions` inside the `kotlin` DSL didn't seem to work 🤷🏻.

chris-horner (approved):

PR #3137, opened 2025-08-22
Cache non-200 network responses in the disk cache.

colinrtwhite:
This was a regression from Coil 2.0 -> 3.0.

Fixes https://github.com/coil-kt/coil/issues/3128

PR #3139, opened 2025-08-22
Restrict which response codes are cacheable.

colinrtwhite:
Follow up to https://github.com/coil-kt/coil/pull/3137. I realized that we shouldn't cache response codes like 500s.


chris-horner (approved):
I should have caught this too

abdulazizumarovich:
When should we expect this change to be released?

lon9man:
@colinrtwhite what do you think about some option to adjust statuses, which should be cacheable?
```
private val CACHEABLE_STATUS_CODES = setOf(300, 301, 404, 405, 410, 414, 501)
```

example:
AWS S3 returns 403 status, when file is not available. and it is very high chance, that it will not be available forever. so we should cache requests with such status. it will be very helpful!

to reload such file from S3 developer can add some version as query-parameter (http://example.com?version=yyyymmdd). 

for now i am implementing custom logic to cache such statuses, which messes original logic in multiple places.

thanks!

Issue #2994, opened 2025-05-21
Crossfade does not actually crossfade


chrisbanes:
**Describe the bug**
As per the title. If I have an image loaded, and then change the `model` to some other URL, the new image is shown as a 'jump cut' and not a crossfade.

The issue seems to be here: https://github.com/coil-kt/coil/blob/12df129b8068392092a23e493100ac03fcae25e0/coil-compose-core/src/androidMain/kotlin/coil3/compose/AsyncImagePainter.android.kt#L29

The current painter is only used if it is the `Loading` state, which seems a strange choice?

**Version**
3.2.0

colinrtwhite:
I believe this was to keep consistency with how `CrossfadeTransition` behaves with `ImageView`s. i.e. we only crossfade from placeholder -> result. Each new request applies its placeholder (sometimes null) and we crossfade to the result when it's loaded. Not crossfading between subsequent image requests is expected behaviour.

If you want to crossfade between subsequent image requests you might be able to use the `transform` param on `AsyncImage` intercept the current state, though that might not work as we need to change the previous state...

It's a common request to crossfade between subsequent requests so I'd like to figure out a better way to support this without breaking the existing behaviour.

chrisbanes:
For now we're just using an `AnimatedContent` and cross-fading between the two `AsyncImage`s. 

That is closer to what we want, but this will crossfade as soon as the `model` changes, whereas ideally we'd only start crossfading when the second image has been fetched and ready to be displayed.

mipastgt:
This is also an annoying issue for me which cost me a lot of time because at first I could not believe that the observed behaviour would make sense to anybody until I found the explanation in this issue.

I am wondering whether it would be possible and would help to be able to configure the current image as the placeholder for the next image. If crossfading from the placeholder to the result works, then this should do the trick.

n2oghost:
I also ran into this issue PLUS the fact that the fade duration doesn't work at all. In my case I need a very slow true crossfade and because of the above two issues I ended up with the following workaround:
````
package com.meditationmoments.meditationmoments.core.ui

import androidx.compose.foundation.Image
import androidx.compose.runtime.Composable
import androidx.compose.runtime.getValue
import androidx.compose.runtime.mutableStateOf
import androidx.compose.runtime.remember
import androidx.compose.runtime.setValue
import androidx.compose.ui.Modifier
import androidx.compose.ui.graphics.painter.Painter
import androidx.compose.ui.layout.ContentScale
[...]

PR #3141, opened 2025-08-23
Add crossfade between subsequent image results in AsyncImage

tak8997:
### Description
Currently, `AsyncImage`'s crossfade only applies from placeholder → result.
When switching between different image models, the new image replaces the old one immediately ("jump cut") without a crossfade.  
This PR introduces an optional parameter to support **crossfading from the previous successful image result → the new result**,  skipping the placeholder.

### Motivation
- Addresses [#2994](https://github.com/coil-kt/coil/issues/2994).
- Many developers expect true crossfade between subsequent requests (not only placeholder → result).

### Notes
- Default behavior remains unchanged to avoid breaking existing apps.
- When enabled, fade duration now properly applies to subsequent requests.
- Tested with multiple consecutive image requests to confirm smooth transition.


coil-core/src/commonMain/kotlin/coil3/request/imageRequests.kt
```diff
@@ -30,6 +30,15 @@ expect val ImageRequest.crossfadeMillis: Int
 
 internal const val DEFAULT_CROSSFADE_MILLIS = 200
 
+fun ImageRequest.Builder.crossfadeBetweenImages(enable: Boolean) = apply {
```
colinrtwhite, line 33: Can you add a short doc comment for this method?

Also we should add `ImageLoader.Builder.crossfadeBetweenImages` so this can be configured per-`ImageLoader`.

coil-core/src/commonMain/kotlin/coil3/request/imageRequests.kt
```diff
@@ -30,6 +30,15 @@ expect val ImageRequest.crossfadeMillis: Int
...
+    extras[crossfadeBetweenImagesKey] = enable
+}
+
+val ImageRequest.crossfadeBetweenImages: Boolean
```
colinrtwhite, line 37: Please annotate this and `crossfadeBetweenImages` with `@ExperimentalCoilApi`.

coil-core/src/commonMain/kotlin/coil3/request/imageRequests.kt
```diff
@@ -30,6 +30,15 @@ expect val ImageRequest.crossfadeMillis: Int
...
+    get() = getExtra(crossfadeBetweenImagesKey)
+
+private val crossfadeBetweenImagesKey = Extras.Key(default = false)
```
colinrtwhite, line 40: We should also add:

```kotlin
val Extras.Key.Companion.crossfadeBetweenImages: Extras.Key<Boolean>
    get() = crossfadeBetweenImagesKey
```

tak8997: ba668de5a748eebea68a43312027a6b31b07d27a
please check this commit

tak8997:
> Thanks for implementing this. I think this looks good for Compose, but do you know how we would also add support to `View`s?
> 
> Ideally, I'd like for this `ImageRequest` param to automatically support `View`s and `Compose`. Currently I could see someone using this param with an `ImageView` and being confused it doesn't work.

tak8997:
> Thanks for implementing this. I think this looks good for Compose, but do you know how we would also add support to `View`s?
> 
> Ideally, I'd like for this `ImageRequest` param to automatically support `View`s and `Compose`. Currently I could see someone using this param with an `ImageView` and being confused it doesn't work.

340a445f8ba3b357814828fad725476d41a0c327
please check this commit

colinrtwhite (commented):
Thanks for working on this. Thinking more about it - I think we should make this solution more generic. Currently we're coupling this behaviour to crossfade, but really we want the ability to use the existing image as the `placeholder` for the next image request. If we do that then both the view and Compose-based crossfade implementations should automatically crossfade from what the previous image is. It would also allow users to skip setting the placeholder if they want to go from first loaded image -> second loaded image without setting a placeholder and without a crossfade.

Unfortunately we don't have knowledge of what the existing image is with views, but we do with Compose/`AsyncImagePainter`. As such, I think it's OK to only support this functionality in Compose. Here's what we should do:

- Revert the changes to `Target` to support views.
- Rename the config option `crossfadeBetweenImages` to `useExistingImageAsPlaceholder(enable: Boolean)` (open to better names) and move it to an `imageRequest.kt` file in `coil-compose-core` (as it's only supported in Compose). We should note in the method description that this config option will only work in Compose.
- Update `AsyncImagePainter` to use the existing `painter` in `onStart` if `placeholder == null` and `useExistingImageAsPlaceholder` is enabled on the `ImageRequest`.

Thanks again for working and iterating on this!

Issue #3152, opened 2025-08-28
Timeout options for ImageRequest.Builder


oungsi2000:
**Is your feature request related to a problem? Please describe.**
When I tried to set a timeout for a Coil 3 ImageRequest, I found it complicated. I had to either use a custom OkHttpClient with a dedicated interceptor or manually wrap the imageLoader.execute call in a withTimeout block

example: 
```kotlin
fun ImageView.loadImage(
    url: String?,
    block: ImageRequest.Builder.() -> Unit = {},
) {
    findViewTreeLifecycleOwner()?.lifecycleScope?.launch {
        runCatching {
            withTimeout(2000) {
                val request =
                    ImageRequest
                        .Builder(this@loadImage.context)
                        .data(url)
                        .apply { block() }
                        .build()

                val result = this@loadImage.context.imageLoader.execute(request)
                this@loadImage.setImageBitmap(result.image?.toBitmap())
            }
        }.onFailure {
            this@loadImage.load(R.drawable.img_fallback)
        }
    }
}
```

I'd like to simplify this as shown in the second code example.

```kotlin
fun ImageView.loadImage(
    url: String?,
    block: ImageRequest.Builder.() -> Unit = {},
) {
    this.load(url) {
        fallback(R.drawable.img_fallback)
        timeout(2000)
        block()
    }
}
```

I found that the ImageView.load extension function internally uses imageLoader.enqueue, which returns a Disposable with an associated Job. Therefore, if there's no simpler way to handle this, I'd like to propose a pull request for the code I've suggested.

colinrtwhite:
What's your use case? Are you looking to have a timeout specifically for Coil's network request? Or overall for an image request?

oungsi2000:
While this is for an image request overall, it seems okay to separate the network timeout from the rest !
I wanted to simplify the process of loading an error image when the timeout limit set in option for ImageRequest.Builder is exceeded.

colinrtwhite:
@oungsi2000 Why do you want a timeout separate from the network timeout? Reading a file or Android resource should be very unlikely to get stuck.

oungsi2000:
@colinrtwhite I apologize for the confusion. 

I wasn't arguing that network requests and overall image requests must be separated. I was suggesting that we apply a timeout to the overall image request, but based on the question, 

> What's your use case? Are you looking to have a timeout specifically for Coil's network request? Or overall for an image request?

[...]

PR #3154, opened 2025-08-30
Add timeout option in ImageRequest.Builder

oungsi2000:
this implements #3152 

I've created timeout option for ImageRequest.Builder. 
Error Image defined by ```error()```is shown when timeout is exceeded

I just simply wrapped ```enqueue()```and ```èxecute()``` with ```withTimeout()``` function. 
so that the entire image loading process is subject to a timeout.

If tests are needed, I'll add some test cases.

Thanks


oungsi2000:
Sorry for submitting this PR without permission first. If it's a problem, I'll close both the issue and the PR.

Issue #1688, opened 2023-03-26
Placeholder affects image scaling when we use CrossfadeTransition and the placeholder image is larger


tfcporciuncula:
**Describe the bug**
When I have `CrossfadeTransition.Factory()` passed to the `ImageLoader`, I get a different final scale in an `AsyncImage` if it uses a placeholder that is larger than the final image. Everything is fine if I remove the crossfade, remove the placeholder, or use a placeholder that matches the size of the final image.

**To Reproduce**
Minimum reprodicible example here: https://github.com/tfcporciuncula/coil-crossfade

You can clone the repo, run the code, and observe that:

- If we keep `CrossfadeTransition.Factory()` around, we'll get different scaling results depending on whether we keep or remove the placeholder.
- If we remove `CrossfadeTransition.Factory()`, the scaling is the same regardless if we have the placeholder or not.
- If we keep `CrossfadeTransition.Factory()` but change the placeholder from `R.drawable.placeholder_larger` to `R.drawable.placeholder`, which has the exact same size as our final imagem, then everything is fine.

**Logs/Screenshots**

Image on the left is what we get with crossfade and the large placeholder. Image on the right is the correct scaling which we get by either removing the crossfade, removing the placeholder, or using a placeholder that matches the final image size.

![1](https://user-images.githubusercontent.com/750114/227771995-33a69506-1ba2-4bd2-816a-c4869202d995.png) ![2](https://user-images.githubusercontent.com/750114/227771998-61ddd330-742a-4102-b844-a8c8314d2ce5.png)

We're using `ContentScale.Crop` above, but we can move to `ContentScale.Fit` to make the difference more visible:

![w](https://user-images.githubusercontent.com/750114/227772144-55858b3d-d08d-430a-8559-8fb36b889c41.png) ![c](https://user-images.githubusercontent.com/750114/227772141-8c97fb71-ae7e-48ab-8e38-b2d4d7059c6c.png)

**Version**
Coil 2.3.0, reproduced on a Pixel 4 running Android 13 and on a Pixel 6 emulator (1080x2400 420dpi) running Android 10 (screenshots from the latter).

colinrtwhite:
Thanks for the repro project! At first glance, seems like this could be related to: https://github.com/coil-kt/coil/issues/1505

tfcporciuncula:
[...]

PR #3175, opened 2025-09-14
Fix CrossfadePainter intrinsic size logic

tak8997:
### Solution
Refactored the intrinsic size computation logic to:
- Prioritize the end image's intrinsic size when available
- Fall back to start image's size only when end image size is unspecified

### Changes
- Ensures ContentScale is applied correctly to the final image regardless of placeholder dimensions

### Motivation
- Addresses #1688

### DIFF
--- a/coil-compose-core/src/commonMain/kotlin/coil3/compose/CrossfadePainter.kt
+++ b/coil-compose-core/src/commonMain/kotlin/coil3/compose/CrossfadePainter.kt
@@ -99,12 +99,14 @@ class CrossfadePainter(
 
         val isStartSpecified = startSize.isSpecified
         val isEndSpecified = endSize.isSpecified
-        if (isStartSpecified && isEndSpecified) {
-            return Size(
-                width = maxOf(startSize.width, endSize.width),
-                height = maxOf(startSize.height, endSize.height),
-            )
+
+        if (isEndSpecified) {
+            return endSize
         }
+        if (isStartSpecified) {
+            return startSize
+        }
+
         if (preferExactIntrinsicSize) {
             if (isStartSpecified) return startSize
             if (isEndSpecified) return endSize


coil-compose-core/src/commonMain/kotlin/coil3/compose/CrossfadePainter.kt
```diff
@@ -99,12 +99,14 @@ class CrossfadePainter(
 
         val isStartSpecified = startSize.isSpecified
         val isEndSpecified = endSize.isSpecified
-        if (isStartSpecified && isEndSpecified) {
-            return Size(
-                width = maxOf(startSize.width, endSize.width),
-                height = maxOf(startSize.height, endSize.height),
-            )
```
tak8997, line 106: May I ask why you chose to use maxOf for startSize and endSize?

coil-compose-core/src/commonMain/kotlin/coil3/compose/CrossfadePainter.kt
```diff
@@ -99,12 +99,14 @@ class CrossfadePainter(
...
+
+        if (isEndSpecified) {
+            return endSize
+        }
```
tak8997, line 105: In my opinion, we could simply set the intrinsicSize based on the endSize. What do you think?

colinrtwhite:
@tak8997 Thanks for looking into that bug. Unfortunately I don't think we can make this change without flagging as this is a fairly sizeable behaviour change for existing users. Could you:

- Rebase on `main`. I added a couple `CrossfadePainter` tests as it didn't have any.
- Add a configuration option (using the same `Extras` pattern as the previous PR) that is false by default to enable this new sizing behaviour.
- Add a regression test in `CrossfadePainterTest` to ensure that [this ticket](https://github.com/coil-kt/coil/issues/1688) is fixed.

You'll likely need to add a secondary `@Deprecated(level = DeprecationLevel.HIDDEN)` constructor to `CrossfadePainter` to preserve binary compatibility.

coil-compose-core/src/commonMain/kotlin/coil3/compose/CrossfadePainter.kt
colinrtwhite: It was to ensure that the placeholder isn't cut off. It's also to have consistency with `CrossfadeDrawable`.

coil-compose-core/src/commonMain/kotlin/coil3/compose/CrossfadePainter.kt
colinrtwhite: That would be a significant behaviour change and I'm not sure about the ramifications. How would crossfading from a painter with an intrinsic size to a painter with no intrinsic size (e.g. `ColorPainter`) inside an `AsyncImage` with no constraints look? Would it have 0 size?

tak8997:
> @tak8997 Thanks for looking into that bug. Unfortunately I don't think we can make this change without flagging as this is a fairly sizeable behaviour change for existing users. Could you:
> 
> * Rebase on `main`. I added a couple `CrossfadePainter` tests as it didn't have any.
> * Add a configuration option (using the same `Extras` pattern as the previous PR) that is false by default to enable this new sizing behaviour.
> * Add a regression test in `CrossfadePainterTest` to ensure that [this ticket](https://github.com/coil-kt/coil/issues/1688) is fixed.
> 
> You'll likely need to add a secondary `@Deprecated(level = DeprecationLevel.HIDDEN)` constructor to `CrossfadePainter` to preserve binary compatibility.

78086de1c69d98e1f3a2cacbb9d226cf73be52c6
683e13197d8fd9121d7aee7ffbb8061785c2166e
sorry for late. please check this🙏

tak8997:
> @tak8997 Thanks for looking into that bug. Unfortunately I don't think we can make this change without flagging as this is a fairly sizeable behaviour change for existing users. Could you:
> 
> * Rebase on `main`. I added a couple `CrossfadePainter` tests as it didn't have any.
> * Add a configuration option (using the same `Extras` pattern as the previous PR) that is false by default to enable this new sizing behaviour.
> * Add a regression test in `CrossfadePainterTest` to ensure that [this ticket](https://github.com/coil-kt/coil/issues/1688) is fixed.
> 
> You'll likely need to add a secondary `@Deprecated(level = DeprecationLevel.HIDDEN)` constructor to `CrossfadePainter` to preserve binary compatibility.

78086de1c69d98e1f3a2cacbb9d226cf73be52c6
683e13197d8fd9121d7aee7ffbb8061785c2166e
ff6d24345bb1ee8ef408151d051265ef98aaef81
sorry for late. please check this🙏

coil-compose-core/src/commonMain/kotlin/coil3/compose/CrossfadePainter.kt
colinrtwhite, line 37: Could you add a `@param preferEndFirstIntrinsicSize` entry to the method description? We can copy it from [here](https://github.com/coil-kt/coil/pull/3175/files?new_files_changed=true#diff-e10b48ac0addc8fee50dd3ffa0b7e20885d2c92be1bcbd56f406180791ef892bR48).

coil-compose-core/src/commonMain/kotlin/coil3/compose/CrossfadePainter.kt
```diff
@@ -45,6 +45,7 @@ class CrossfadePainter(
     val timeSource: TimeSource = TimeSource.Monotonic,
     val fadeStart: Boolean = true,
     val preferExactIntrinsicSize: Boolean = false,
+    val preferEndFirstIntrinsicSize: Boolean = false
```
colinrtwhite, line 48: We'll need to add a `@Deprecated("Kept for binary compatibility.", level = DeprecationLevel.HIDDEN)` version of this constructor without the `preferEndFirstIntrinsicSize` argument to preserve binary compatibility. Check out `NetworkClient` for an example .

colinrtwhite (commented):
Thanks for implementing this! Left two remaining comments otherwise LGTM

coil-compose-core/src/commonMain/kotlin/coil3/compose/CrossfadePainter.kt
tak8997: 50ab6102543d524ed9f7e15a04b36c2bf7c7c9bc
please check this

PR #3201, opened 2025-10-11
skip Main dispatcher in enqueue when not required

k2w4t4h:
### Summary
Skip dispatching to the main dispatcher when it isn’t required by the request. This removes an extra main-thread hop in RealImageLoader.enqueue and reduces scheduling overhead for background/non‑UI requests.

### Motivation
Previously, enqueue always started on the main dispatcher even when the request didn’t need it, adding unnecessary context switching and latency. I ported the same decision logic already used in 'execute' into 'enqueue'.

### DIFF
--- a/coil-core/src/commonMain/kotlin/coil3/RealImageLoader.kt
+++ b/coil-core/src/commonMain/kotlin/coil3/RealImageLoader.kt
@@ -34,6 +34,7 @@ import kotlinx.atomicfu.atomic
 import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.CoroutineExceptionHandler
 import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.CoroutineStart
 import kotlinx.coroutines.Deferred
 import kotlinx.coroutines.SupervisorJob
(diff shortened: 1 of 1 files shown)


colinrtwhite:
Thanks for proposing this, though I don't think we should make this change. `needsExecuteOnMainDispatcher` was added as a optimization for Compose which uses `execute`, however are a few reasons this is less beneficial for `enqueue`:

- With `execute` it lets us avoid all the coroutines machinery whereas we always have to call `scope.async` in `enqueue`. This means we get less of a performance benefit (though we would still get some).
- We needed to add this for Compose as it uses a different main dispatcher in its context that doesn't support immediate dispatching. `enqueue` doesn't have this problem as there's no existing coroutine context and `Dispatchers.Main.immediate` should already avoid dispatching if the request is started on the main thread.
- Changing this could change the behaviour of some requests. `ImageRequest.Builder.target {}` would now execute on an IO thread instead of the main thread which could be unexpected and difficult to debug for users. This is less important for `execute` as the result is returned at the call site, however `ImageRequest.Builder.target` is the main way to get the result from an `enqueue'd request.

PR #3228, opened 2025-11-12
Prevent multiple parallel network requests for the same URL

tak8997:
### Summary
This PR introduces a mechanism to prevent multiple parallel network requests for the same URL in Coil.

### Details
- Uses `HashMap` and `CompletableDeferred` to track in-flight requests.
- If a request for the same URL is already in progress, subsequent requests will wait for the first to complete and reuse its result.

### Motivation
- Address #1461


coil-network-core/src/commonMain/kotlin/coil3/network/NetworkFetcher.kt
```diff
@@ -208,6 +242,28 @@ class NetworkFetcher(
...
+                .joinToString(separator = ",") { "${it.first}:${it.second}" }
+            append(headers)
+            append('|')
+            append(options.networkCachePolicy)
+            append('|')
+            append(options.diskCachePolicy)
+            append('|')
+            append(options.diskCacheKey)
+        }
+    }
```
tak8997: I would appreciate some advice on how to make a request key

tak8997:
@colinrtwhite Could you please check this PR?

PR #3259, opened 2025-12-13
Fix maxBitmapSize calculation with FILL for preventing too large bitmap crash on Android platform

irgaly:
AsyncImage on Compose will exceeds maxBitmapSize limitation, and leads crash on some Android platforms.

* relates:  #1349

## Reproducer

* platform: Android 14 Emulator
* coil v3.3.0
* Reproducer condition:
    * **The image has larger size than maxBitmapSize**
    * **The incoming constraints of AsyncImage are Infinity**
    * **The Scale config is FILL**

```kotlin
Box(Modifier.padding(padding)) {
    Column(
        Modifier
            // IntrinsicSize leads the constraints to Infinity
            .width(IntrinsicSize.Max)
            .background(Color.Yellow.copy(alpha = 0.4f))
    ) {
        AsyncImage(
            model = ImageRequest.Builder(LocalPlatformContext.current)
                // 8000px x 4000px image
                .data("https://placehold.jp/8000x4000.png")
                .build(),
            contentDescription = null,
            Modifier.fillMaxWidth().aspectRatio(4f / 1f),
            // ContentScale.Crop is used on calculation as FILL
            contentScale = ContentScale.Crop,
        )
        Text("Title text")
        Text("Body text. Body text. Body text.")
    }
}
```

This layout is like this:

image

In this example, the AsyncImage receives `Infinity` constraints in ConstraintsSizeResolver.

image

Then, resolved size is `Size(width = Undefined, height = Undefined)` (Size.ORIGINAL).

image

Then, DecodeUtils.computeSizeMultiplier() will calculate as `Scale.FILL` to `(dstWidth, dstHeight) = (4096, 4000)`. dstWidth is limited to maxBitmapSize. So the `multiplier = 1.0f` will be used.

image

Finally, AsyncImage will draw the bitmap with 8000px x 4000px, and this will cause crash on some Android versions (old versions or hardware limitation?).

> java.lang.RuntimeException: Canvas: trying to draw too large(128000000bytes) bitmap.

image

---

### Fix

The problem is `DecodeUtils.computeSizeMultiplier()`. When it computes with `Scale.FILL`, the multiplier and resulting bitmap will be larger than the size of maxBitmapSize.

I fixed computeSizeMultiplier() on this PR, by limiting the multiplier after calculating that with `FILL` or `FIT`. The multiplier will be limited to the both of the maxBitmapSize.width and the maxBitmapSize.height.

image

### DIFF
--- a/coil-core/api/android/coil-core.api
+++ b/coil-core/api/android/coil-core.api
@@ -360,9 +360,12 @@ public final class coil3/decode/DecodeUtils {
 	public static final field INSTANCE Lcoil3/decode/DecodeUtils;
 	public static final fun calculateInSampleSize (IIIILcoil3/size/Scale;)I
 	public static final fun computeDstSize-sEdh43o (IILcoil3/size/Size;Lcoil3/size/Scale;Lcoil3/size/Size;)J
-	public static final fun computeSizeMultiplier (DDDDLcoil3/size/Scale;)D
-	public static final fun computeSizeMultiplier (FFFFLcoil3/size/Scale;)F
-	public static final fun computeSizeMultiplier (IIIILcoil3/size/Scale;)D
+	public static final fun computeSizeMultiplier (DDDDLcoil3/size/Scale;Lcoil3/size/Size;)D
+	public static final fun computeSizeMultiplier (FFFFLcoil3/size/Scale;Lcoil3/size/Size;)F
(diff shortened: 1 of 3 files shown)


irgaly:
I fixed lint error and updated API dump.

irgaly:
`./gradlew iosSimulatorArm64Test macosArm64Test` is passed on my local Mac and my forked repository https://github.com/irgaly/coil/actions/runs/20220087681/job/58040384774

PR #3286, opened 2026-01-15
Make BitmapFetcher available in common code

dsvdmeer:
In our compose multiplatform project we noticed that on iOS Coil was unable to display in-memory images (`coil3.Bitmap` instances) because there is no fetcher registered. On Android this does work out of the box due to the existence of a `BitmapFetcher`. I could see no reason why this should be implemented only for Android, and so in this PR I've moved the implementation to common code.

The Android implementation originally called `data.toDrawable(options.context).asImage()`, but the `toDrawable(...)` step in-between is basically a no-op as far as I can tell, so it should be safe to just leave that out.

### DIFF
--- a/coil-core/src/commonMain/kotlin/coil3/fetch/BitmapFetcher.kt
+++ b/coil-core/src/commonMain/kotlin/coil3/fetch/BitmapFetcher.kt
@@ -1,20 +1,18 @@
 package coil3.fetch
 
-import android.graphics.Bitmap
+import coil3.Bitmap
 import coil3.ImageLoader
 import coil3.asImage
 import coil3.decode.DataSource
 import coil3.request.Options
-import coil3.util.toDrawable
 
 internal class BitmapFetcher(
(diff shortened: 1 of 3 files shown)


coil-core/src/commonMain/kotlin/coil3/fetch/BitmapFetcher.kt
```diff
@@ -1,20 +1,18 @@
...
-import coil3.util.toDrawable
 
 internal class BitmapFetcher(
     private val data: Bitmap,
-    private val options: Options,
 ) : Fetcher {
 
     override suspend fun fetch(): FetchResult {
         return ImageFetchResult(
-            image = data.toDrawable(options.context).asImage(),
```
colinrtwhite: We should probably use `asImage(resources)` here on Android (for automatic density scaling), however that's a separate issue. You're right `data.toDrawable(options.context).asImage()` is equivalent to `data.asImage()`.

PR #3304, opened 2026-02-07
  Add lint check for kotlin.error() misuse inside ImageRequest.Builder. (#2887)

### LINKED ISSUE #2887: Conflicting error extension function can lead to Unexpected Runtime Exceptions

The name of the extension function `error` for the `ImageRequest.Builder` class is confusing. I'm using Coil in a view-based Android project. I was using the `ImageView.load` function inside of a fragment and would use `error(@DrawableRes drawableResId: Int)`. But since it's an extension function, it requires importing. When I was writing code, the IDEA automatically imported the `error(message: Any)` function from Kotlin stdlib, which throws IllegalStateException and causes a runtime exception.

Could the `error` function be renamed to `onError`? Or the library should include a lint check that warns the user when there is an `error(message: Any)` function (from stdlib) usage in the context of `ImageRequest.Builder` class.

Amir-yazdanmanesh:
## Summary                                                                                                                                                               
  - Add new `coil-lint` module with `ErrorFunctionDetector` that warns when `kotlin.error()` is accidentally used inside `ImageRequest.Builder` lambdas (e.g.,
  `imageView.load {}`, `AsyncImage`, `SubcomposeAsyncImage`)                                                                                                               
  - IDE auto-import often selects `kotlin.error()` instead of Coil's `error()` extension, causing runtime `IllegalStateException` instead of setting an error drawable     

  ## Test plan
  - [x] Unit test: `kotlin.error()` inside `load {}` triggers warning
  - [x] Unit test: Coil's `error()` extension does not trigger warning
  - [x] Unit test: `kotlin.error()` outside Coil context does not trigger warning
  - [x] Unit test: `kotlin.error()` in regular lambdas does not trigger warning

  Closes #2887

### DIFF
--- a/coil-lint/build.gradle.kts
+++ b/coil-lint/build.gradle.kts
@@ -1,0 +1,41 @@
+import coil3.setupPublishing
+import com.vanniktech.maven.publish.JavaLibrary
+import com.vanniktech.maven.publish.JavadocJar
+import org.gradle.api.attributes.java.TargetJvmVersion
+
+plugins {
+    id("org.jetbrains.kotlin.jvm")
+    id("com.vanniktech.maven.publish.base")
+}
+
+setupPublishing {
+    configure(JavaLibrary(javadocJar = JavadocJar.Empty()))
+}
+
+kotlin {
+    jvmToolchain(17)
+}
+
+// Lint dependencies require JVM 11+, so we must request JVM 17 compatible dependencies.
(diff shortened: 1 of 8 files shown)


colinrtwhite (approved):
Awesome, thank you!

Amir-yazdanmanesh:
> Awesome, thank you!

you’re welcome bro

PR #3305, opened 2026-02-09
Use web worker to decode images on Web

terrakok:
This change introduces asynchronous image decoding on the Web platform using a web worker. This prevents blocking the main thread during image decoding operations.

The decoding logic in `SkiaImageDecoder` has been refactored to use a new `expect` function, `decodeBitmap`. The Web implementation of this function, `decodeBitmap`, leverages a web worker for decoding and includes optimizations to read image dimensions for JPEG and PNG files directly, avoiding the need to decode the full image just to get its size.

The JVM and Native platforms provide synchronous implementations as they don't have the same main-thread blocking concerns for this operation. A new dependency on `kotlinx-browser` has been added for the Web targets.

## Before
```kotlin
val bitmap = Bitmap()
bitmap.allocN32Pixels(outWidth, outHeight)
Canvas(bitmap).use { canvas ->
    // heavy UI web thread blocking call
    canvas.drawImageRect(
        image = image,
        src = Rect.makeWH(srcWidth.toFloat(), srcHeight.toFloat()),
        dst = Rect.makeWH(outWidth.toFloat(), outHeight.toFloat()),
    )
}
return bitmap
```

**There are frame drops:**
coil-before-opt

## After
```kotlin
// async decodes an image to a bitmap on a special web worker. doesn't block UI thread :)
val webBitmap = decodeBytesToBitmap(bytes, width, height)

// pass bitmap ArrayBuffer to the skiko memory
val skikoData = webBitmap.passToSkiko()

val colorInfo = ColorInfo(
    ColorType.RGBA_8888,
    ColorAlphaType.UNPREMUL,
    ColorSpace.sRGB,
)
val imageInfo = ImageInfo(colorInfo, width, height)
val image = Image.makeRaster(imageInfo, skikoData, imageInfo.minRowBytes)
return Bitmap.makeFromImage(image)
```

**There are no frame drops:**
coil-after-opt

The change is prepared together with @eymar 🚀

### DIFF
--- a/coil-core/karma.config.d/karma.conf.js
+++ b/coil-core/karma.config.d/karma.conf.js
@@ -1,0 +1,13 @@
+config.customLaunchers = {
+    ChromeForComposeTests: {
+        base: "Chrome",
+        flags: [
+            "--no-sandbox",
+            "--disable-search-engine-choice-screen",
+            "--disable-setuid-sandbox",
+            "--enable-webgl",
+            "--ignore-gpu-blocklist",
+            "--in-process-gpu"
+        ]
+    }
+}


colinrtwhite (approved):
Looks great, thanks for implementing this! Left a few minor comments also looks like `jsNodeTest` is failing.

terrakok:
@colinrtwhite 

Compose for Web is not designed to be used in a node environment. Tests used to work because there were no explicit skiko references.
New changes import "org.jetbrains.skiko.wasm.awaitSkiko" function. The function is declared in "js-reexport-symbols.mjs" module, which is not possible to use in nodejs.

colinrtwhite:
@terrakok Makes sense! I remember migrating over to nodeJS tests as I couldn't get the browser tests to run successfully on Github Actions. Unfortunately looks like they're still failing when re-enabled. I'll take another run at fixing them in a separate PR soon unless you beat me to it.

Issue #3234, opened 2025-11-18
Alignment not working when crossfade and placeholder is present


xVemu:
**Describe the bug**
When both crossfade and placeholder are present:
<img width="507" height="511" alt="Image" src="https://github.com/user-attachments/assets/8dad8089-3859-45c3-8d5a-9b9f0ebaae51" />

When crossfade or painter is not present:

<img width="510" height="509" alt="Image" src="https://github.com/user-attachments/assets/ab28453e-11c1-4721-8464-c80412ecccff" />

---
**To Reproduce**
```kt
val platformContext = LocalPlatformContext.current

AsyncImage(
    ImageRequest.Builder(platformContext).crossfade(true)
        .data("https://www.totalwine.com/media/sys_master/twmmedia/ha5/h01/8799883329566.png")
        .build(),
    contentDescription = "todo.name",
    alignment = Alignment.BottomEnd,
    placeholder = ColorPainter(Color.Red),
    modifier = Modifier
        .size(400.dp)
        .background(Color.Blue),
)
```

**Version**

io.coil-kt.coil3:coil-compose
io.coil-kt.coil3:coil-network-ktor3
io.coil-kt.coil3:coil-svg
all version 3.3.0

I tested it on Android 36 and Desktop with same result.

kairowan:
I've submitted a fix for this in #3335. The root cause is that `CrossfadePainter.drawPainter()` uses `inset()` to forcibly center the painter, which overrides the `alignment` parameter already handled by `ContentPainterModifier.draw()`. The fix removes this centering logic and lets the outer modifier handle alignment properly.

PR #3335, opened 2026-02-28
Fix: CrossfadePainter ignores alignment parameter

kairowan:
## Problem

When using `AsyncImage` with `crossfade(true)` and a custom `alignment` parameter (e.g., `Alignment.TopCenter`), the alignment is ignored during the crossfade transition. The image always appears centered regardless of the alignment setting.

## Root Cause

The `CrossfadePainter.drawPainter()` method uses `inset()` to forcibly center the painter, which overrides the `alignment` parameter that is already handled by `ContentPainterModifier.draw()`.

## Solution

Remove the centering logic from `CrossfadePainter.drawPainter()` and let it draw directly to the full canvas size. The outer `ContentPainterModifier` already handles alignment properly.

## Changes

- Removed `inset()` centering logic in `drawPainter()`
- Removed unused `computeDrawSize()` method
- Removed unused imports

## Testing

Existing unit tests pass as they use same-sized painters where centering has no effect.


kairowan:
Fixes #3234

PR #3345, opened 2026-03-06
feat: Add default `accept` header on Android

KamilKurde:
This PR adds `accept` header for Android used by some CDNs, so that the best possible image format could be delivered by the CDN, instead of falling back to "default" one (which often is just `image/jpeg`).

It also retains `accept` header if it was explicitly set by the user.

### DIFF
--- a/coil-network-core/src/androidHostTest/kotlin/coil3/network/AcceptHeadersTest.kt
+++ b/coil-network-core/src/androidHostTest/kotlin/coil3/network/AcceptHeadersTest.kt
@@ -1,0 +1,149 @@
+package coil3.network
+
+import android.graphics.ImageDecoder
+import android.os.Build
+import coil3.Extras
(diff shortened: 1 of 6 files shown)


colinrtwhite:
Thanks for implementing this. After thinking about this, I'd prefer to avoid setting default headers by default in Coil's network artifact like this for a couple reasons:

- **It changes all Android HTTP requests by default**: this could have behavioural changes for users depending on how the server handles the new header.
- **Accept priorities are arguable in some cases**: Should WebP always be preferred over PNG/JPEG? In some cases the user might want a larger lossless PNG instead of a lossy WebP.
- **Already user-configurable**: It's already possible to set default headers like `accept` via a `coil3.Interceptor` that sets `ImageRequest.Builder.httpHeaders`.

Notably, Glide also doesn't set a default `accept` header either.

PR #3401, opened 2026-04-24
Evaluate cache strategy before enforcing failure response code

### LINKED ISSUE #3391: Non-200 (or 304) cached response never expired
https://github.com/coil-kt/coil/blob/3.4.0/coil-network-core/src/commonMain/kotlin/coil3/network/NetworkFetcher.kt#L82-L86

```kotlin
                cacheResponse = snapshot.toNetworkResponseOrNull()
                if (cacheResponse != null) {
                    throwIfFailureResponseCode(cacheResponse)

                    readResult = cacheStrategy.value.read(cacheResponse, newRequest(), options)
                    if (readResult.response != null) {
                        return SourceFetchResult(
                            source = snapshot.toImageSource(),
                            mimeType = getMimeType(url, readResult.response.headers[CONTENT_TYPE]),
                            dataSource = DataSource.DISK,
                        )
                    }
                }

```

If cached response is non-200 (or 304), it will always be thrown via throwIfFailureResponseCode and never have chance to ask cacheStrategy (aka `cache-control`)

leesgit:
## Summary
Resolves #3391.

`NetworkFetcher` cached non-2xx responses listed in `CACHEABLE_STATUS_CODES` (300, 301, 404, 405, 410, 414, 501) to disk, but then unconditionally threw `HttpException` on the next read before `CacheStrategy.read()` was invoked. As a result, cache-aware strategies such as `CacheControlCacheStrategy` could never evaluate whether the cached failure had expired — a cached 404 would throw forever, regardless of `Cache-Control` headers.

## Changes
- Moved `throwIfFailureResponseCode(cacheResponse)` until after `cacheStrategy.read(...)` returns a response.
- The check now runs only when the strategy opts to reuse the cached response. When the strategy returns a fresh `NetworkRequest` instead (e.g. because the cached entry is stale), the fetcher proceeds to the network as expected.

This keeps the existing behavior for `CacheStrategy.DEFAULT` (which always returns the cached response) and unblocks `CacheControlCacheStrategy` from honoring expiration on cached failures.

## Testing
- `./gradlew :coil-network-core:jvmTest --tests coil3.network.NetworkFetcherTest` — all 5 tests pass.
- Existing `cached404ResponseThrows` still passes (default strategy preserves current behavior).
- New `cached404IsRefreshedWhenCacheStrategyRequestsNewRequest` verifies that when a strategy returns `ReadResult(request)`, the cached 404 no longer short-circuits with `HttpException` and the network request is issued.
- `./gradlew :coil-network-core:spotlessCheck` — passes.

### DIFF
--- a/coil-network-core/src/commonMain/kotlin/coil3/network/NetworkFetcher.kt
+++ b/coil-network-core/src/commonMain/kotlin/coil3/network/NetworkFetcher.kt
@@ -81,10 +81,13 @@ class NetworkFetcher(
                 // Return the image from the disk cache if the cache strategy agrees.
                 cacheResponse = snapshot.toNetworkResponseOrNull()
                 if (cacheResponse != null) {
-                    throwIfFailureResponseCode(cacheResponse)
-
                     readResult = cacheStrategy.value.read(cacheResponse, newRequest(), options)
                     if (readResult.response != null) {
+                        // Only enforce the failure code check when the strategy opts to use the
+                        // cached response. This lets cache-aware strategies (e.g. cache-control)
+                        // issue a fresh request when a cached failure (e.g. 404) has expired.
+                        throwIfFailureResponseCode(readResult.response)
+
                         return SourceFetchResult(
                             source = snapshot.toImageSource(),
                             mimeType = getMimeType(url, readResult.response.headers[CONTENT_TYPE]),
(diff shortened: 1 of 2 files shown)


colinrtwhite:
Thanks this looks like the proper fix.

colinrtwhite (approved):

PR #3458, opened 2026-06-19
fix: Fix JVM memory cache sizing to use Runtime.maxMemory()

gajera1080:
### Description
This PR fixes a performance issue on Desktop JVM where the memory cache size was artificially constrained, resulting in excessive cache evictions under heavy load.

Previously, `totalAvailableMemoryBytes()` in `nonAndroidMain` used a hardcoded fallback of `512MB`. Since `nonAndroidMain` encompasses JVM desktop targets, this severely limited the default memory cache (15%) to just 76.8MB, completely ignoring the JVM's actual allocated maximum heap size. 

**Changes:**
- Removed the `512MB` fallback `actual` implementation from `nonAndroidMain`.
- Shifted the fallback into `nonJvmCommonMain` so it still gracefully applies to JS/Wasm/Apple Native targets where querying system memory isn't fully supported.
- Added a dedicated `jvmMain` implementation that accurately queries the heap limit using `Runtime.getRuntime().maxMemory()`, with an 8GB fallback for environments where it is undefined (`Long.MAX_VALUE`).

### DIFF
--- a/coil-core/src/jvmMain/kotlin/coil3/util/contexts.jvm.kt
+++ b/coil-core/src/jvmMain/kotlin/coil3/util/contexts.jvm.kt
@@ -1,0 +1,7 @@
+package coil3.util
+
+import coil3.PlatformContext
+
+internal actual fun PlatformContext.totalAvailableMemoryBytes(): Long {
+    return Runtime.getRuntime().maxMemory()
+}
--- a/coil-core/src/jvmMain/kotlin/coil3/util/contexts.jvm.kt
+++ b/coil-core/src/jvmMain/kotlin/coil3/util/contexts.jvm.kt
@@ -3,5 +3,10 @@ package coil3.util
 import coil3.PlatformContext
 
 internal actual fun PlatformContext.totalAvailableMemoryBytes(): Long {
-    return Runtime.getRuntime().maxMemory()
+    val maxMemory = Runtime.getRuntime().maxMemory()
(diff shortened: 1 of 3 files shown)


coil-core/src/jvmMain/kotlin/coil3/util/contexts.jvm.kt
```diff
@@ -0,0 +1,7 @@
+package coil3.util
+
+import coil3.PlatformContext
+
+internal actual fun PlatformContext.totalAvailableMemoryBytes(): Long {
+    return Runtime.getRuntime().maxMemory()
```
colinrtwhite: `Runtime.getRuntime().maxMemory()` [can return](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/Runtime.html#maxMemory()) `Long.MAX_VALUE` if it's undefined. Let's handle that case and return 8GB as a fallback.

colinrtwhite:
Thanks - this looks good to add though:

> OutOfMemoryErrors under heavy load

I know this is likely an LLM-generated description, but this isn't possible. If the memory cache is too small images will only be evicted quicker - not OOM.

gajera1080:
Thanks for the review! Good catch on the cache evictions vs OOM. I've updated the description and added the 8GB fallback for Long.MAX_VALUE. Let me know if you need anything else!

colinrtwhite (approved):

PR #3469, opened 2026-06-20
Fix GIF filter quality in Compose

### LINKED ISSUE #2939: FilterQuality parameter is ignored when displaying Gif
**Describe the bug**
Hi all,
I have the need to display pixel art images in my app. 
When using static images everything works fine, in fact, when I apply FilterQuality.None to AsyncImage the app displays the image as I want.

Without FilterQuality.None applied:
Image

With FilterQuality.None applied:
Image

The problem occurs when I want to display Gif.
In particular seems like FilterQuality is ignored by Coil.

This is what I expected:
Image

and this is what I get:
Image

Am I doing something wrong?

**To Reproduce**
Just use AsyncImage with FilterQuality.None applied 

```
AsyncImage(
      model = /* use a gif */,
      contentDescription = contentDescription,
      filterQuality = FilterQuality.None,
     )
```

**Version**
Coil 3.0.4 - Pixel 9 Pro XL (but not limited to it) (Android 15)

qureshiayaz29:
## Summary
- apply Compose FilterQuality.None to Android Drawable-backed images
- propagate the flag into Coil's ScaleDrawable children so AnimatedImageDrawable can receive it
- add Android host tests for DrawableImage and ScaleDrawable-backed images

Fixes #2939

## Tests
- ./gradlew :coil-core:checkKotlinAbi :coil-compose-core:checkKotlinAbi :coil-compose-core:testAndroidHostTest :coil-compose-core:spotlessKotlinCheck --console=plain
- ./test.sh

### DIFF
--- a/coil-compose-core/src/androidMain/kotlin/coil3/compose/ImagePainter.android.kt
+++ b/coil-compose-core/src/androidMain/kotlin/coil3/compose/ImagePainter.android.kt
@@ -1,6 +1,5 @@
 package coil3.compose
 
-import android.graphics.drawable.Drawable
 import androidx.compose.ui.graphics.Canvas
 import androidx.compose.ui.graphics.FilterQuality
 import androidx.compose.ui.graphics.asImageBitmap
@@ -12,7 +11,6 @@ import coil3.DrawableImage
 import coil3.Image
 import coil3.PlatformContext
 import coil3.asDrawable
-import coil3.size.ScaleDrawable
 import com.google.accompanist.drawablepainter.DrawablePainter
 
 actual fun Image.asPainter(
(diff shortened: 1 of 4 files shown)


coil-compose-core/src/androidMain/kotlin/coil3/compose/ImagePainter.android.kt
```diff
@@ -22,9 +24,18 @@ actual fun Image.asPainter(
...
+        drawable = asDrawable(context.resources).mutate().apply {
+            setFilterBitmapRecursive(filterQuality != FilterQuality.None)
+        },
     )
     else -> ImagePainter(this)
 }
 
 internal actual val Canvas.nativeCanvas get() = nativeCanvas
+
+private fun Drawable.setFilterBitmapRecursive(filter: Boolean) {
```
colinrtwhite: Let's remove this method, inline the call to `setFilterBitmap` at the call site, and have `ScaleDrawable` implement `setFilterBitmap` and `isFilterBitmap` and delegate those methods to the `child` similar to other methods on that class.

coil-compose-core/src/androidUnitTest/kotlin/coil3/compose/ImagePainterTest.kt
```diff
@@ -0,0 +1,55 @@
...
+    fun drawableImageAsPainter_appliesFilterQuality() {
+        val drawable = TestDrawable()
+
+        drawable.asImage().asPainter(context, FilterQuality.None)
+
+        assertFalse(drawable.filterBitmap)
+    }
+
+    @Test
+    fun scaleDrawableImageAsPainter_appliesFilterQualityToChild() {
```
colinrtwhite: Let's remove this test and only validate the cases where `FilterQuality.None` as well as `FilterQuality.Low` for `TestDrawable`.

colinrtwhite (approved):
Thanks!

PR #3470, opened 2026-06-21
Fix crash when measuring AsyncImage with inverted constraints

### LINKED ISSUE #3459: Exception thrown when `min > max` in `coerceIn` of `Constraints.constrainWidth/Height`
**Describe the bug**

Exception thrown when `min > max` in `coerceIn`, in [utils.kt](https://github.com/coil-kt/coil/blob/66ddf1c2c2b86f00908e2e3c7eb07862b44560ce/coil-compose-core/src/commonMain/kotlin/coil3/compose/internal/utils.kt#L206)

```kotlin
internal fun Constraints.constrainWidth(width: Float) =
    width.coerceIn(minWidth.toFloat(), maxWidth.toFloat())

internal fun Constraints.constrainHeight(height: Float) =
    height.coerceIn(minHeight.toFloat(), maxHeight.toFloat())
```

**To Reproduce**

Reproducing is tricky. In our case, it involved two things coinciding: collapsing of our in-house-built top bar, and Coil's async image load succeeding.

### Claude's details are helpful in this case

**How the crash happens:**

`coerceIn(min, max)` throws `IllegalArgumentException` when `min > max`. The crash message `maximum 0.0 is less than minimum 160.0` means the incoming `Constraints` have `minWidth=160` and `maxWidth=0` — an invalid state.

The stack trace shows this triggered during Compose snapshot observation (`DerivedSnapshotState → observe`), which points to `AsyncImagePainter.painter` (backed by `mutableStateOf`) changing state when the image loads:

```kotlin
// AsyncImagePainter.kt:158
private var painter: Painter? by mutableStateOf(null)  // triggers recomposition on load

// AsyncImagePainter.kt:202-203
override val intrinsicSize: Size
    get() = painter?.intrinsicSize ?: Size.Unspecified  // reads the mutableStateOf
```

When the image finishes loading, `painter` changes from `null` to a `BitmapPainter` with intrinsic size (160px wide), triggering `invalidateMeasurement()`. On the next layout pass, `modifyConstraints` receives constraints that have `maxWidth=0` (the header has collapsed) but `minWidth=160` (set by a prior `modifyConstraints` call that set `minWidth = constrainWidth(scaledWidth)` = 160 when the header was still open). Coil then calls `coerceIn(160.0, 0.0)` → crash.

**Root cause summary:** A race between the collapsing header animation (which drives `maxWidth` toward 0) and the async image load completing (which triggers re-measurement with stale `minWidth=160` constraints). The bug is in Coil 3's `modifyConstraints` — it doesn't guard against `minWidth > maxWidth` before calling `coerceIn`.

**Version**

Coil version 3.5.0

### Claude's graceful solution

Return max when min > max.

```kotlin
  internal fun Constraints.constrainHeight(height: Float): Float {
      val min = minHeight.toFloat()
      val max = maxHeight.toFloat()
      return if (min > max) max else height.coerceIn(min, max)
  }
```

  Returning max (0 in this crash) means: "the image gets the space the parent can actually provide", which is the
  correct graceful degradation — the composable occupies zero size during the collapse transition, then resizes
  normally once constraints stabilize.

ChristophyBarth:
### Summary
The `Float` overloads of `Constraints.constrainWidth` / `constrainHeight` in `coil-compose-core` use the standard-library `coerceIn`, which throws when `minWidth > maxWidth`:

```
java.lang.IllegalArgumentException: Cannot coerce value to an empty range:
maximum 0.0 is less than minimum 160.0.
```

`modifyConstraints` can receive inverted constraints during transient layout states. For example, a collapsing parent drives `maxWidth` to 0 on the same frame an image finishes loading and re-measures with a larger `minWidth`. `constrainWidth(intrinsicWidth)` then calls `coerceIn(160f, 0f)` and crashes (full stack trace in #3459).

### Fix
Switch the two overloads to Compose's `fastCoerceIn`, which clamps (min, then max) without validating the range, returning `max` for an inverted range instead of throwing. Compose's own `Int` `Constraints.constrainWidth` / `constrainHeight` already do exactly this, so the change brings the `Float` overloads in line with them. For every valid constraint the result is identical to `coerceIn`, so there is no behavior change to normal layouts.

> #3459 suggested `if (min > max) max else coerceIn(...)`. `fastCoerceIn` is behaviorally identical for that case but reuses Compose's own helper and matches the integer overloads, so I went with it.

### Tests
Added `ConstrainSizeTest` (`commonTest`) covering normal clamping plus the inverted-constraint regression (throws on `main`, passes with the fix). The inverted `Constraints` is built via Compose's internal non-validating packer, since the public factory rejects `min > max`.

Verified locally:
- `./gradlew checkKotlinAbi spotlessCheck`: green (change is `internal`-only, so no ABI files change)
- `./gradlew :coil-compose-core:jvmTest :coil-compose-core:testAndroidHostTest`: green

Fixes #3459.


coil-compose-core/src/commonTest/kotlin/coil3/compose/internal/ConstrainSizeTest.kt
```diff
@@ -0,0 +1,24 @@
+package coil3.compose.internal
+
+import androidx.compose.ui.unit.Constraints
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+class ConstrainSizeTest {
```
colinrtwhite: I think we can remove these tests and `ConstrainSizeInvertedTest`. The comment on `constrainWidth` should be enough of a guard against accidentally reintroducing this behaviour.

colinrtwhite (commented):
Thanks - I think we can merge this once the tests are removed (see the other comment for rationale).

coil-compose-core/src/commonTest/kotlin/coil3/compose/internal/ConstrainSizeTest.kt
ChristophyBarth: Right.

ChristophyBarth:
> Thanks - I think we can merge this once the tests are removed (see the other comment for rationale).

Thanks for the review. I have torn the files down.

colinrtwhite (approved):

PR #3481, opened 2026-06-26
fix: Query physical memory on Apple targets for cache sizing

gajera1080:
### Description
This PR replaces the hardcoded `512MB` fallback for `totalAvailableMemoryBytes()` on Apple targets (iOS/macOS) with a real implementation that queries `NSProcessInfo.processInfo.physicalMemory`. 

This enables Coil's memory cache limit (which defaults to 15%) to automatically scale to the hardware specs of the running Apple device rather than being artificially capped at `76.8MB` for all iOS and macOS users.

The 512MB fallbacks have been cleanly moved to target-specific `jsCommon` and `linux` source sets to avoid duplicate actual declarations.

### Testing
- [x] Verified code formatting with `./gradlew spotlessCheck`.
- [x] Verified JVM, JS, WasmJS, and Linux target compilation with `./gradlew :coil-core:compileKotlinJvm :coil-core:compileKotlinJs :coil-core:compileKotlinWasmJs :coil-core:compileKotlinLinuxX64 :coil-core:compileKotlinLinuxArm64`.

### DIFF
--- a/coil-core/src/appleMain/kotlin/coil3/util/contexts.apple.kt
+++ b/coil-core/src/appleMain/kotlin/coil3/util/contexts.apple.kt
@@ -1,0 +1,8 @@
+package coil3.util
+
+import coil3.PlatformContext
+import platform.Foundation.NSProcessInfo
+
+internal actual fun PlatformContext.totalAvailableMemoryBytes(): Long {
+    return NSProcessInfo.processInfo.physicalMemory.toLong()
+}
--- a/coil-core/src/appleMain/kotlin/coil3/util/contexts.apple.kt
+++ b/coil-core/src/appleMain/kotlin/coil3/util/contexts.apple.kt
@@ -4,5 +4,8 @@ import coil3.PlatformContext
 import platform.Foundation.NSProcessInfo
 
 internal actual fun PlatformContext.totalAvailableMemoryBytes(): Long {
-    return NSProcessInfo.processInfo.physicalMemory.toLong()
+    // A single app process is budgeted a fraction of the total physical memory.
+    // We estimate the process budget to be 25% of physical memory, capped at 1 GB.
+    val physicalMemory = NSProcessInfo.processInfo.physicalMemory.toLong()
+    return (physicalMemory / 4).coerceAtMost(1024L * 1024 * 1024)
 }
--- a/coil-core/src/jsCommonMain/kotlin/coil3/util/contexts.jsCommon.kt
+++ b/coil-core/src/jsCommonMain/kotlin/coil3/util/contexts.jsCommon.kt
@@ -1,0 +1,7 @@
+package coil3.util
+
+import coil3.PlatformContext
+
+internal actual fun PlatformContext.totalAvailableMemoryBytes(): Long {
+    return 512L * 1024 * 1024 // 512 MB
+}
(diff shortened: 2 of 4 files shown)


coil-core/src/appleMain/kotlin/coil3/util/contexts.apple.kt
```diff
@@ -0,0 +1,8 @@
+package coil3.util
+
+import coil3.PlatformContext
+import platform.Foundation.NSProcessInfo
+
+internal actual fun PlatformContext.totalAvailableMemoryBytes(): Long {
+    return NSProcessInfo.processInfo.physicalMemory.toLong()
```
colinrtwhite: The issue with using `NSProcessInfo.processInfo.physicalMemory` is that this returns the total RAM the device has - not just the RAM available to the process. That's the more applicable number as we only want to consume a small subset of what's budgeted for the app.

gajera1080: Updated `totalAvailableMemoryBytes()` to return an estimated process memory budget (25% of physical RAM, capped at 1 GB) instead of the total hardware RAM. This prevents over-allocating default memory cache sizes (e.g., capping it at 150MB instead of 600MB+ on 4GB devices).

coil-core/src/appleMain/kotlin/coil3/util/contexts.apple.kt
```diff
@@ -0,0 +1,11 @@
+package coil3.util
+
+import coil3.PlatformContext
+import platform.Foundation.NSProcessInfo
+
+internal actual fun PlatformContext.totalAvailableMemoryBytes(): Long {
+    // A single app process is budgeted a fraction of the total physical memory.
+    // We estimate the process budget to be 25% of physical memory, capped at 1 GB.
```
colinrtwhite, line 8: What's the rationale behind this size? Did Codex/Claude suggest this? Did you test to see if this matches available per-app RAM budgets in practice?

gajera1080: This heuristic was suggested by my AI coding assistant to address the lack of a static process memory limit API on Apple platforms.

I haven't run target-specific profiling to verify exact Jetsam limits on physical devices, but mathematically:

**Standard devices (4 GB+ RAM):** The budget caps at 1 GB, yielding a predictable 150 MB default memory cache (`15% of 1 GB`). This matches standard configurations in other image libraries.
**Lower-end watchOS/tvOS (1-2 GB RAM):** It scales down gracefully to 38 MB–76 MB default caches.

colinrtwhite: > I haven't run target-specific profiling to verify exact Jetsam limits on physical devices

We should figure out how much is typically allocated to iOS apps by checking that ourselves and make a heuristic based on that. I wouldn't trust AI estimation for something this important.

> This matches standard configurations in other image libraries.

Can you provide a source for this?

gajera1080: **Kingfisher (Apple):** Defaults memory limit to 25% of physical memory (totalMemory / 4) (see `line 293` in [ImageCache.swift](https://github.com/onevcat/Kingfisher/blob/master/Sources/Cache/ImageCache.swift)).

PR #3485, opened 2026-06-27
Properly convey typed asset file uri open requirement

nift4:
There are various special cases where a content:// uri MUST be opened with openTypedAssetFileDescriptor and some specific MIME type set, such as MediaStore's thumbnails (including music thumbnails already supported). If you open the exact same Uri using the normal openAssetFileDescriptor, in the album case you get an error and in the music case you get the .mp3 file as file descriptor instead. So it's important to split ContentMetadata to allow conveying how the Uri needs to be opened exactly. Otherwise, (image) decoders that work like VideoFrameDecoder, using the content uri directly, would accidentally open .mp3 file or album (which can't be opened) instead of the thumbnail. Prefereably, the decoder directly uses AssetFileDescriptor, which doesn't require special handling as the uri was already opened. So for the benefit of StaticImageDecoder, this is extracted into superclass BaseContentMetadata, from which ContentMetadata and TypedContentMetadata inherit. (Note that uri is NOT extracted into superclass because the semantics differ, any user must always check which subclass it is before using the uri, hence adding uri field to BaseContentMetadata would just mislead API consumers.)

If there are multiple decoders like StaticImageDecoder that require seekable AssetFileDescriptor, we don't need to check multiple times, store the check result as Lazy in BaseContentMetadata.

### DIFF
--- a/coil-core/src/androidMain/kotlin/coil3/decode/ImageSource.kt
+++ b/coil-core/src/androidMain/kotlin/coil3/decode/ImageSource.kt
@@ -1,6 +1,10 @@
 package coil3.decode
 
 import android.content.res.AssetFileDescriptor
+import android.os.Bundle
+import android.system.ErrnoException
+import android.system.Os
+import android.system.OsConstants.SEEK_SET
 import androidx.annotation.DrawableRes
 import coil3.Uri
 
@@ -12,12 +16,45 @@ class AssetMetadata(
 ) : ImageSource.Metadata()
 
 /**
- * Metadata containing the [uri] and associated [assetFileDescriptor] of a `content` URI.
+ * Abstract superclass for metadata relating to the asset file descriptor obtained from a
+ * `content` URI.
+ */
+abstract class BaseContentMetadata : ImageSource.Metadata() {
+    abstract val assetFileDescriptor: AssetFileDescriptor
+    val seekable by lazy {
+        try {
+            Os.lseek(
+                assetFileDescriptor.fileDescriptor, assetFileDescriptor.startOffset,
+                SEEK_SET,
+            )
+            true
+        } catch (_: ErrnoException) {
+            false
+        }
+    }
+}
+
+/**
+ * Metadata containing the [uri] and associated [assetFileDescriptor] of a `content` URI
+ * that can be opened without special considerations.
  */
 class ContentMetadata(
     val uri: Uri,
-    val assetFileDescriptor: AssetFileDescriptor,
-) : ImageSource.Metadata()
+    override val assetFileDescriptor: AssetFileDescriptor,
+) : BaseContentMetadata()
+
+/**
+ * Metadata containing the [uri] and associated [assetFileDescriptor] of a `content` URI
+ * that must be opened with [android.content.ContentResolver.openTypedAssetFileDescriptor] (as
+ * opposed to [android.content.ContentResolver.openAssetFileDescriptor]), while also passing the
+ * [mimeTypeFilter] and [opts] (if present) to the content resolver.
+ */
+class TypedContentMetadata(
+    val uri: Uri,
+    val mimeTypeFilter: String,
+    val opts: Bundle?,
+    override val assetFileDescriptor: AssetFileDescriptor,
+) : BaseContentMetadata()
 
 /**
  * Metadata containing the [packageName], [resId], and [density] of an Android resource.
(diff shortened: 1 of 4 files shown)


colinrtwhite:
Thanks - can you add an instrumentation regression test that reproduces the behaviour?

coil-core/src/androidMain/kotlin/coil3/decode/StaticImageDecoder.kt
```diff
@@ -144,14 +144,10 @@ fun ImageSource.toImageDecoderSourceOrNull(
...
-            try {
-                // Ensure the file descriptor supports lseek.
-                // https://github.com/coil-kt/coil/issues/2434
-                val asset = metadata.assetFileDescriptor
-                Os.lseek(asset.fileDescriptor, asset.startOffset, SEEK_SET)
-                return ImageDecoder.createSource { asset }
-            } catch (_: ErrnoException) {}
+        // Ensure the file descriptor supports lseek.
+        // https://github.com/coil-kt/coil/issues/2434
+        metadata is BaseContentMetadata && SDK_INT >= 29 && metadata.seekable -> {
```
colinrtwhite, line 149: Existing users are only looking for `ContentMetadata`. Asking them to change to look for `BaseContentMetadata` is a pretty big behaviour change that could be unexpected. I'd definitely like to avoid this if possible.

coil-core/src/androidMain/kotlin/coil3/decode/ImageSource.kt
```diff
@@ -12,12 +16,45 @@ class AssetMetadata(
...
+/**
+ * Metadata containing the [uri] and associated [assetFileDescriptor] of a `content` URI
+ * that must be opened with [android.content.ContentResolver.openTypedAssetFileDescriptor] (as
+ * opposed to [android.content.ContentResolver.openAssetFileDescriptor]), while also passing the
+ * [mimeTypeFilter] and [opts] (if present) to the content resolver.
+ */
+class TypedContentMetadata(
+    val uri: Uri,
+    val mimeTypeFilter: String,
+    val opts: Bundle?,
```
colinrtwhite, line 55: It looks like we don't use `mimeTypeFilter` or `opts` in this new class. Instead of introducing new classes to the API maybe it would be better to add a backward compatible new field to `ContentMetadata` (e.g. `typed: Boolean = false`) to indicate if it needs to be opened with `openTypedAssetFileDescriptor`?

nift4: Hi, the rationale behind that was is that all of Uri, mimeTypeFilter and opts must be the same to get the same result. If you open a MediaStore
audio uri with null opts (instead of setting EXTRA_SIZE), even though you set image/* MIME type, it would NOT open thumbnail, but .mp3 file, instead. (You could say it's a MediaStore bug, but nothing we can change anyway) So for the content Uri to be meaningful these two properties are also required, because they are part of the identity of the fd.

If these are removed, downstream Decoders or similar cannot call openTypedAssetFileDescriptor to get an image fd. Of course, as you said, there currently isn't any Decoder in coil3 that requires opening the content Uri other than the video one, so these fields aren't currently used. But they are required if such a Decoder would ever be added. Do you think they should be removed, or should they be kept for future-proof-ness?

nift4:
> can you add an instrumentation regression test that reproduces the behaviour?

It's only possible for this to be an issue if some custom Decoder is used. I can write an example one l, if it's OK to add a custom Decoder to the tests for this?

coil-core/src/androidMain/kotlin/coil3/decode/StaticImageDecoder.kt
nift4: Hi, the impact of this change is limited because the default fetcher sill always return ContentMetadata, unless a music thumbnail is loaded. If then only default decoders are used it's also no issue.

If a music thumbnail is loaded and the user uses a custom Decoder that also uses ContentMetadata it would impact them.

But even if a user is impacted, I guess it shouldn't cause an image to fail to load, only to fall back to okio Buffer based loading.

I think it's an OK tradeoff to make, because the alternative, using ContentMetadata for typed too, combined with a Decoder that isn't updated for this change, would end up in image loading failure, which is worse than using okio Buffer to load but succeeding.

edit: but of course if you still think it's a problem I can change it, as if a decoder is updated it doesn't impact it either way.

PR #3494, opened 2026-07-05
Add allowPartialImage option for StaticImageDecoder

### LINKED ISSUE #3482: [Feature Request] Make onPartialImageListener configurable in StaticImageDecoder
Starting with [Coil 3.0.4](https://coil-kt.github.io/coil/changelog/#304-november-25-2024), StaticImageDecoder hardcodes onPartialImageListener to always return true. This causes partial images (e.g., from incomplete downloads or partially written content URIs) to be rendered silently instead of throwing an exception.

Relevant code in [StaticImageDecoder.kt:](https://github.com/coil-kt/coil/blob/39e7d982f6d1979979d3402d468cc77e6510de39/coil-core/src/androidMain/kotlin/coil3/decode/StaticImageDecoder.kt#L87)
```kotlin
private fun ImageDecoder.configureImageDecoderProperties() {
    onPartialImageListener = ImageDecoder.OnPartialImageListener { true }
    // ...
}
```

### Rationale:
In many production scenarios, displaying a partial or corrupted image is undesirable. We would prefer the decoding to fail so that it can be caught by a RetryInterceptor or handled via error state UI.
Currently, to disable this behavior, we have to:
1. Disable the default StaticImageDecoder via ImageLoader.Builder option [enableStaticImageDecoder](https://github.com/coil-kt/coil/blob/39e7d982f6d1979979d3402d468cc77e6510de39/coil-core/src/androidMain/kotlin/coil3/RealImageLoader.android.kt#L104).
2. Provide a full "clone" of StaticImageDecoder as a custom decoder just to change this one line.
3. Manually keep this clone in sync with upstream changes in the Coil codebase.
Having a configurable option would allow us to use the optimized StaticImageDecoder while still maintaining control over how partial images are handled.

### Proposed Solution:
Expose a new property (e.g., allowPartialImage) in ImageRequest and Options via Extras. Since this is specific to Android's ImageDecoder, it could be implemented as an extension in the androidMain source set, similar to bitmapConfig.
Example API usage:
```kotlin
    .allowPartialImage(false) // Set globally
    .build()

// Or per-request
    .data(uri)
    .allowPartialImage(false) 
    .build()
```

Implementation in StaticImageDecoder:

```kotlin
private fun ImageDecoder.configureImageDecoderProperties() {
    onPartialImageListener = ImageDecoder.OnPartialImageListener { options.allowPartialImage }
    // ...
}
```

This ensures StaticImageDecoder remains flexible for users who require strict image integrity. On a side note, I am interested in contributing this feature! If this proposal looks good to the maintainers, please let me know and I can put together a Pull Request.

frozenstar80:
## Summary

Closes #3482

Starting with Coil 3.0.4, `StaticImageDecoder` hardcodes `onPartialImageListener` to always return `true`, which silently renders partial or corrupted images instead of surfacing a failure. This blocks `RetryInterceptor` and error-state UI from working correctly.

This PR exposes a new Android-only `allowPartialImage` option backed by `Extras`, following the existing pattern of platform-specific extras (e.g. `bitmapConfig`).

## API Changes

New Android-only extensions (in `androidMain`):

```kotlin
// Global default on ImageLoader
ImageLoader.Builder(context)
    .allowPartialImage(false)
    .build()

// Per-request override
ImageRequest.Builder(context)
    .data(uri)
    .allowPartialImage(false)
    .build()

### CODE REVIEW

**Comment by frozenstar80** [comment]
@colinrtwhite  Can you please verify this change for the enhancement in allowpartialimage

File: coil-core/src/androidMain/kotlin/coil3/request/imageRequests.android.kt [hunk 1]
```diff
@@ -255,6 +255,33 @@ val Options.allowConversionToBitmap: Boolean
 val Extras.Key.Companion.allowConversionToBitmap: Extras.Key<Boolean>
     get() = allowConversionToBitmap
 
+// endregion
+// region allowPartialImage
+
+/**
+ * Allow [android.graphics.ImageDecoder] to decode partial images.
+ *
+ * If false, decoding will fail when [android.graphics.ImageDecoder] detects incomplete image data.
```
colinrtwhite, line 264: We should add `Default: true` similar to other configuration option docs.

colinrtwhite (approved):
Thanks - in the future please avoid opening PRs for enhancements that haven't been confirmed yet as they might not be accepted.

PR #3495, opened 2026-07-05
Use real memory and disk space on Apple targets (iOS/macOS)

frozenstar80:
## Summary

Previously `totalAvailableMemoryBytes()` and `remainingFreeSpaceBytes()` were hardcoded to 512 MB and 4 GB for all non-JVM targets, including Apple (iOS/macOS/tvOS/watchOS). This caused the memory cache to be sized against a fixed constant instead of actual device RAM.

On an iPhone 16 Pro (8 GB RAM), the image cache was capped at ~77 MB (15% of 512 MB) instead of a proportional share of the real 8 GB, leaving significant memory unused.

## Changes

Splits the `nonJvmCommonMain` fallback into per-platform actuals following the same `appleMain` / `jsCommonMain` / `linuxMain` pattern already used by `addAppleComponents`:

- **`appleMain`**: reads `NSProcessInfo.processInfo.physicalMemory` for real RAM; queries `NSFileManager.attributesOfFileSystemForPath` for real free disk space
- **`jsCommonMain`**: retains 512 MB / 4 GB constants (no browser/Node.js API available)
- **`linuxMain`**: retains 512 MB / 4 GB constants (TODOs left for `sysinfo()` / `statvfs()` follow-up)

## Testing

- All existing unit tests unchanged
- Build requires Java 21 (paparazzi plugin); validated file structure locally

review-path-agent

### DIFF
--- a/coil-core/src/androidMain/kotlin/coil3/decode/StaticImageDecoder.kt
+++ b/coil-core/src/androidMain/kotlin/coil3/decode/StaticImageDecoder.kt
@@ -15,6 +15,7 @@ import coil3.annotation.InternalCoilApi
 import coil3.asImage
 import coil3.decode.BitmapFactoryDecoder.Companion.DEFAULT_MAX_PARALLELISM
 import coil3.fetch.SourceFetchResult
+import coil3.request.allowPartialImage
 import coil3.request.Options
 import coil3.request.allowRgb565
 import coil3.request.bitmapConfig
@@ -84,7 +85,7 @@ class StaticImageDecoder(
     }
 
     private fun ImageDecoder.configureImageDecoderProperties() {
-        onPartialImageListener = ImageDecoder.OnPartialImageListener { true }
+        onPartialImageListener = ImageDecoder.OnPartialImageListener { options.allowPartialImage }
         allocator = if (options.bitmapConfig.isHardware) {
             ImageDecoder.ALLOCATOR_HARDWARE
         } else {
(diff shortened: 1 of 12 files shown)


coil-core/src/androidMain/kotlin/coil3/decode/StaticImageDecoder.kt
```diff
@@ -84,7 +85,7 @@ class StaticImageDecoder(
     }
 
     private fun ImageDecoder.configureImageDecoderProperties() {
-        onPartialImageListener = ImageDecoder.OnPartialImageListener { true }
```
colinrtwhite, line 87: This PR contains the `allowPartialImage` changes, which are unrelated.

coil-core/src/appleMain/kotlin/coil3/util/contexts.apple.kt
```diff
@@ -0,0 +1,12 @@
+package coil3.util
+
+import coil3.PlatformContext
+import platform.Foundation.NSProcessInfo
+
+internal actual fun PlatformContext.totalAvailableMemoryBytes(): Long {
+    // NSProcessInfo.physicalMemory returns the total physical RAM as a ULong.
+    // Clamp to Long.MAX_VALUE for safety on theoretical future hardware.
+    val physicalMemory = NSProcessInfo.processInfo.physicalMemory
```
colinrtwhite, line 9: The issue with using `NSProcessInfo.processInfo.physicalMemory` is that this returns the total RAM the device has - not just the RAM available to the process. That's the more applicable number as we only want to consume a small subset of what's budgeted for the app.

PR #3504, opened 2026-07-13
Adds support for loading images from blob URLs (object URLs) on web (JS/Wasm) targets

terrakok:
## Summary

Adds support for loading images from blob URLs (object URLs) on web (JS/Wasm) targets, e.g. `blob:http://localhost:8081/a3953f30-e6ef-47b6-8750-731ce8afe0f8` created via `URL.createObjectURL(...)`.

Previously such a request failed **before it ever reached a fetcher**: a `String` request is mapped to a `Uri` via `StringMapper.toUri()`, and `parseUri` threw an `IndexOutOfBoundsException` on the nested-scheme shape of a blob URL (`blob:http://…`), because it tracked an `authorityStartIndex` past the already-set `pathStartIndex`. This violated `toUri`'s documented "will not throw if the URI is malformed" contract.

## Changes

- **Fix `parseUri` (`coil-core`, common):** an authority may not begin after the path has already started. Added a `pathStartIndex == -1` guard to the authority-detection branch so nested-scheme URIs (`blob:http://…`) parse without throwing and `Uri.toString()` still round-trips the original string. Verified against all existing `UriTest` cases with no regressions.
- **`BlobUriFetcher` (web only):** matches on the raw string prefix `blob:` (object URLs have a nested origin such as `blob:http://…` or `blob:null/…`, so the parsed scheme isn't reliable), reads the referenced bytes via the browser `fetch()` API, and returns a `SourceFetchResult` (`DataSource.MEMORY`), mirroring `DataUriFetcher`. Decoding then flows through the existing `SkiaImageDecoder`/web-worker path.
- Byte reading is an `expect`/`actual` (`jsMain` + `wasmJsMain`) because the `Promise` / `ArrayBuffer → ByteArray` interop differs between the two targets.
- **Auto-registration** via the ServiceLoader eager-init hook pattern already used by the Ktor network fetchers (`jsMain`/`wasmJsMain`), so blob support is enabled automatically on web with no user configuration.

## Testing

- `UriTest`: added cases asserting `blob:http://…` no longer throws and round-trips, and that `blob:null/…` parses with scheme `blob`.
- `BlobUriFetcherTest` (`jsCommonTest`): factory accepts `blob:` URIs (http and opaque origins) and rejects non-blob URIs.

## Notes

- No public API changes — all additions are internal.

### DIFF
--- a/coil-core/src/commonMain/kotlin/coil3/Uri.kt
+++ b/coil-core/src/commonMain/kotlin/coil3/Uri.kt
@@ -215,7 +215,10 @@ private fun parseUri(
                 ) {
                     if (index + 2 < original.length &&
                         original[index + 1] == '/' &&
-                        original[index + 2] == '/'
+                        original[index + 2] == '/' &&
+                        // An authority cannot start after the path has already begun. This guards
+                        // against nested-scheme URIs (e.g. "blob:http://host/id") throwing.
+                        pathStartIndex == -1
                     ) {
                         // Standard URI with an authority (e.g. "file:///path/image.jpg").
                         openScheme = false
--- a/coil-core/src/commonTest/kotlin/coil3/UriTest.kt
+++ b/coil-core/src/commonTest/kotlin/coil3/UriTest.kt
@@ -259,6 +259,20 @@ class UriTest {
     }
 
     @Test
+    fun blobWithHttpOrigin() {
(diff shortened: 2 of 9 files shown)


colinrtwhite (approved):
Sorry for the delay on this! Looks good to merge to me - thanks!

PR #3528, opened 2026-08-03
Pin GitHub Actions to commit SHAs and declare workflow permissions

Totara-thib:
Hi, drive-by CI hardening in two commits, one logical change each.

**Commit 1 pins every action to its commit sha**, versions kept as comments. A tag like `@v6` is a movable pointer: whoever controls the action, or anyone who compromises it, can re-point it and your next run executes their code with the job's token. The sensitive spots here are the deploy jobs: `deploy-snapshot` holds your Sonatype credentials and `deploy-docs` can push to the repo. Same pattern as the tj-actions/changed-files incident (CVE-2025-30066). Renovate keeps opening its bump PRs exactly as before, it understands sha pins with a version comment, same as its recent spotless and screenshot updates.

**Commit 2 adds `permissions:` blocks.** Without one, every job's GITHUB_TOKEN inherits the repository default scope. The scopes were derived from what each job actually does with the token, not guessed: `contents: read` at the workflow level, and `contents: write` only on `deploy-docs` because `deploy_docs.sh` runs `mkdocs gh-deploy`, which pushes the gh-pages branch. The snapshot deploy authenticates through the Sonatype secrets, the GitHub token itself is unused there.

One heads up: if the org uses an Actions allowlist in settings, patterns written against tags (like `owner/action@v6`) stop matching once refs are shas and workflows refuse to start. Entries need to be `owner/action@*` in that case.

All shas were resolved from the upstream repos and cross checked against their release tags. No workflow logic changes.

Found with the Plumber CLI (https://github.com/getplumber/plumber), verified on main. I will also open a PR which adds it to CI so this does not quietly drift back, that one is a bonus, this PR stands on its own.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,14 +19,14 @@ jobs:
         runs-on: ubuntu-latest
         timeout-minutes: 30
         steps:
-            - uses: actions/checkout@v7
-            - uses: actions/setup-java@v5
+            - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+            - uses: actions/setup-java@b6effb05e454b25005698d916606bdc6ffcbf961 # v5.7.0
               with:
                   distribution: 'zulu'
                   java-version: ${{ env.JAVA_VERSION }}
 
             - name: Check Gradle wrapper
-              uses: gradle/actions/wrapper-validation@v6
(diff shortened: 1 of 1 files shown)


colinrtwhite:
Thanks for opening this, however after thinking about it I'd prefer to keep the major version syntax even though it's less secure. We only use first party Github actions except the popular Android emulator runner action, which have a lot of scrutiny. Also I'd like to avoid the extra PR churn of manually updating for every minor/patch release.

PR #3534, opened 2026-08-09
Avoid internal Skiko APIs on web

colinrtwhite:
Replaces direct internal Skiko access on JS and Wasm with bindings to the packaged module and adds regression coverage for module initialization and image decoding.


terrakok:
You introduced a critical issue here: since wasm and js code uses `@file:JsModule("./js-reexport-symbols.mjs")`, this will be leaked to the published `klib` and break users apps with CMP 1.13.

PR #3559, opened 2026-08-28
Completely remove the Node polyfill plugin

colinrtwhite:
This isn't necessary - we had it enabled to work around an Okio-related warning in our tests. Remove it to reduce the maintenance burden; I'm a bit worried this test dependency could leak into the main artifact one day accidentally.


chris-horner (approved):
🔥

PR #3566, opened 2026-09-02
Fix cancellation handling in deduplicated requests

### LINKED ISSUE #3562: DeDupeConcurrentRequestStrategy permanently deadlocks a key when a waiter is canceled
# `DeDupeConcurrentRequestStrategy` permanently deadlocks a key when a waiter is canceled

> [!IMPORTANT]
> `receiveCatching()` is called **outside** the `try`/`finally` in `DeDupeConcurrentRequestStrategy.apply`. A waiter that is canceled while parked therefore never runs `release()`, `observerCount` stays above zero forever, and the key is never removed from `concurrentRequests`. Every subsequent request for that key suspends forever.

---

## Describe the bug

`DeDupeConcurrentRequestStrategy` leaks its observer count when a **waiter** coroutine is canceled while suspended, which permanently disables cleanup for that key. After it happens, no image with that URL will ever load again for the lifetime of the `ImageLoader` — which, via `SingletonImageLoader.Factory`, is the lifetime of the process.

This is not an exotic edge case. It is the normal lifecycle of a `LazyColumn`/`LazyRow` that shows the same image URL in more than one visible item: the duplicate requests are exactly the ones `DeDupeConcurrentRequestStrategy` parks, and fling-recycling cancels them in the same frame.

**Symptom:** a set of images that share a URL stop loading — permanently. Scrolling back over them does not help, and there is no error, no log line, and no `onError` callback. The requests simply never complete.

### Root cause

[`ConcurrentRequestStrategy.kt#L45-L75`](https://github.com/coil-kt/coil/blob/main/coil-network-core/src/commonMain/kotlin/coil3/network/ConcurrentRequestStrategy.kt#L45-L75), unchanged in `3.4.0`, `3.5.0`, `3.6.0` and current `main`:

```kotlin
override suspend fun apply(
    key: String,
    block: suspend () -> FetchResult,
): FetchResult {
    var shouldWait = true
    val request = synchronized(lock) {
        concurrentRequests.getOrPut(key) {
            shouldWait = false
            Request()
        }
    }.acquire()                      // <-- observerCount++

    if (shouldWait) {
        request.channel.receiveCatching()   // <-- (!) OUTSIDE the try/finally
    }

    try {
        return block().also {
            request.markSucceeded()
        }
    } catch (e: Exception) {
        request.channel.trySend(Unit)
        throw e
    } finally {
        request.release {                   // <-- observerCount--  (never reached by a canceled waiter)
            synchronized(lock) {
                concurrentRequests -= key
            }
        }
    }
}
```

`receiveCatching()` only swallows channel *closure*. It does **not** swallow job cancellation — a canceled coroutine still throws `CancellationException` out of the suspension point. Because that call sits before the `try`, the stack unwinds without ever entering it, so the matching `release()` never runs.

### Walkthrough

Five `LazyColumn` items share one URL. Item 1 becomes the leader, items 2–5 park in `receiveCatching()`. The user flings and all five Compose jobs are canceled:

| Step | Coroutine | Effect on `observerCount` | Map entry |
| :--- | :--- | :--- | :--- |
| 1 | leader + 4 waiters call `acquire()` | `0 → 5` | present |
| 2 | waiters 2–5 canceled while parked | **unchanged (5)** — `release()` skipped | present |
| 3 | leader canceled, `finally` runs `release()` | `5 → 4` | present |
| 4 | `--observerCount <= 0`? **no** (4). `hasSucceeded`? **no**. | — | **never closed, never removed** |
| 5 | any later request for the key | `4 → 5, 6, 7…` | parks in `receiveCatching()` **forever** |

From step 4 onward the `--observerCount <= 0` branch in `Request.release` is dead for that key. Only a *successful* fetch could still remove the entry — and no fetch can ever succeed, because every new caller sees `shouldWait == true` and parks behind a channel that is never closed and holds no tokens.

### Secondary defects on the same path

| # | Location | Issue |
| :-: | :--- | :--- |
| 1 | `catch (e: Exception)` | Misses `Throwable`s that are not `Exception`s. An `OutOfMemoryError` from `NetworkResponseBody.readBuffer()` skips `trySend(Unit)` entirely, so parked waiters hang **deterministically** — see Repro 3. |
| 2 | `channel.trySend(Unit)` | The failure path emits exactly **one** token to wake exactly one waiter. Delivered to a receiver that is concurrently being canceled, the element is passed to `onCancellation` and — with no `onUndeliveredElement` on the channel — silently discarded. The baton is lost. |
| 3 | `.acquire()` outside `synchronized(lock)` | A caller can read a `Request` under the lock, be preempted, and `acquire()` it after the leader has already closed it and removed it from the map — two concurrent network fetches for one URL, which is what the class exists to prevent. |
| 4 | `release()` → `cleanup()` | Takes `Request.lock` and then the map lock. `apply` takes them in the opposite order and only escapes an AB‑BA deadlock because `synchronized(lock) { … }.acquire()` exits the block before calling `.acquire()`. Moving `.acquire()` inside the lock — the natural fix for 3 — introduces a deadlock. Worth a comment at minimum. |
| 5 | `concurrentRequests` | Unbounded, no eviction, and private with no reset API. Every leaked key is retained for the process lifetime along with its never-closed `Channel`. |

---

## To Reproduce

Every repro below uses only public Coil API against the published artifact — no reflection, no copied internals. Drop any of them into a JVM or Android unit test source set.

<details>
<summary><b>Shared helper</b> (click to expand)</summary>

```kotlin
import coil3.annotation.ExperimentalCoilApi
import coil3.decode.DataSource
import coil3.decode.ImageSource
import coil3.fetch.FetchResult
import coil3.fetch.SourceFetchResult
import coil3.network.DeDupeConcurrentRequestStrategy
import kotlinx.coroutines.*
import okio.Buffer
import okio.FileSystem
import kotlin.random.Random

private fun result(): FetchResult = SourceFetchResult(
    source = ImageSource(Buffer(), FileSystem.SYSTEM),
    mimeType = null,
    dataSource = DataSource.NETWORK,
)
```

</details>

### Repro 1 — cancellation storm, the reported case

Each iteration models one *scroll past, then scroll back* cycle over a shared image URL: a leader plus six waiters, all canceled at jittered moments, then one fresh, never-canceled request for the same key.

```kotlin
@OptIn(ExperimentalCoilApi::class)
@Test
fun cancellationStorm() = runBlocking {
    var stuck = 0
    repeat(400) {
        val strategy = DeDupeConcurrentRequestStrategy()
        val scope = CoroutineScope(Dispatchers.Default + SupervisorJob())
        val started = CompletableDeferred<Unit>()

        val leader = scope.launch {
            strategy.apply("key") {
                started.complete(Unit)
                awaitCancellation()
            }
        }
        started.await()

        val waiters = List(6) { scope.launch { strategy.apply("key") { result() } } }
        delay(40)

        // Recycle: cancel the leader and every waiter at jittered moments.
        val jobs = (waiters + leader).shuffled()
        jobs.forEach { job -> scope.launch { delay(Random.nextLong(3)); job.cancel() } }
        jobs.forEach { it.join() }
        delay(30)

        // Scroll back: can this key ever load again?
        if (withTimeoutOrNull(1_500L) { strategy.apply("key") { result() } } == null) stuck++
        scope.cancel()
    }
    println("permanently stuck keys: $stuck / 400")
}
```

Actual output on `3.6.0`:

```text
permanently stuck keys: 14 / 400
```

The exact count is timing-dependent[^timing], but it is never zero, and every hit is permanent for that key.

A single-threaded variant of the same thing, deterministic and small enough to keep as a regression test — this fails today:

```kotlin
@Test
fun dedupe_canceled_waiter_does_not_strand_the_key() = runTest {
    val strategy = DeDupeConcurrentRequestStrategy()
    val leaderStarted = CompletableDeferred<Unit>()
    val releaseLeader = CompletableDeferred<Unit>()

    val leader = launch {
        strategy.apply("key") {
            leaderStarted.complete(Unit)
            releaseLeader.await()
            newFetchResult(1)
        }
    }
    leaderStarted.await()

    val waiter = launch { strategy.apply("key") { newFetchResult(2) } }
    runCurrent()
    waiter.cancelAndJoin()          // the case that is never exercised today

    releaseLeader.complete(Unit)
    leader.join()

    var attempted = false
    withTimeout(5_000) { strategy.apply("key") { attempted = true; newFetchResult(3) } }
    assertTrue(attempted)
}
```

### Repro 2 — what it looks like in an app

No special setup beyond enabling the strategy; the duplicated URL is the only ingredient.

```kotlin
ImageLoader.Builder(context)
    .components {
        add(
            OkHttpNetworkFetcherFactory(
                concurrentRequestStrategy = { DeDupeConcurrentRequestStrategy() },
            )
        )
    }
    .build()
```

```kotlin
// The same URL in many rows: placeholder avatars, channel logos, category art…
LazyColumn {
    items(200) {
        AsyncImage(
            model = "https://example.com/placeholder.png",
            contentDescription = null,
        )
    }
}
```

Fling hard a few times. Some or all of those images stop loading and never recover, with no error surfaced through `onError` or the `DebugLogger`.

### Repro 3 (secondary) — a non-`Exception` failure strands waiters

> [!NOTE]
> This is a **separate, lower-priority defect** (1 in the table above) that happens to sit on the same code path. It is included because it needs no cancellation, no threads and no race, so it isolates the `catch (e: Exception)` half of the problem. Feel free to split it into its own issue.

The leader fails with a `Throwable` that is not an `Exception`. Nothing is canceled; a single waiter is parked.

```kotlin
@OptIn(ExperimentalCoilApi::class)
@Test
fun leaderErrorStrandsWaiter() = runBlocking {
    val strategy = DeDupeConcurrentRequestStrategy()
    val started = CompletableDeferred<Unit>()
    val fail = CompletableDeferred<Unit>()
    val scope = CoroutineScope(Dispatchers.Default + SupervisorJob())

    val leader = scope.async {
        runCatching {
            strategy.apply("key") {
                started.complete(Unit)
                fail.await()
                throw OutOfMemoryError("decode blew up")
            }
        }
    }
    started.await()

    val waiter = scope.async<Any?> {
        withTimeoutOrNull(2_000L) { strategy.apply("key") { result() } }
    }
    delay(200)
    fail.complete(Unit)

    println("leader = ${leader.await().exceptionOrNull()}")
    println("waiter = ${waiter.await() ?: "null  <-- HUNG"}")
    scope.cancel()
}
```

Actual output on `3.6.0`:

```text
leader = java.lang.OutOfMemoryError: decode blew up
waiter = null  <-- HUNG
```

Expected: the waiter is resumed and retries, as `dedupe_failed_request_resumes_single_waiter` establishes for the `Exception` case.

---

## Why the existing test does not catch this

[`dedupe_canceled_request_resumes_waiter`](https://github.com/coil-kt/coil/blob/main/coil-network-core/src/commonTest/kotlin/coil3/network/ConcurrentRequestStrategyTest.kt) reads as if it covers cancellation, but it **never cancels a `Job`**:

```kotlin
val leader = async {
    runCatching {
        strategy.apply("key") {
            started.send(1)
            cancelLeader.await()
            throw CancellationException("leader canceled")   // <-- thrown by hand
        }
    }
}
```

Throwing a `CancellationException` by hand from inside `block()`, in a coroutine whose `Job` is still active, is caught by `catch (e: Exception)` and behaves *identically* to `error("leader failed")` in `dedupe_failed_request_resumes_single_waiter`. No coroutine is ever unwound by cancellation and no continuation is ever canceled while suspended, so the test duplicates the failure-path coverage that already exists and adds none of the coverage its name promises.

Two structural gaps keep the defect out of reach of the whole suite:

- [ ] **No test ever cancels a waiter.** The waiter always runs to completion. Only the leader ever fails — and the leader has `shouldWait == false`, so it never touches `receiveCatching()`. The one coroutine that *can* be suspended in the unprotected call is never canceled.
- [ ] **No test ever inspects cleanup with waiters present.** `dedupe_failed_request_without_waiters_is_cleaned_up` covers only the zero-waiter case, where `observerCount` goes `1 → 0` and the cleanup branch is trivially correct. The leak only manifests when waiters hold the count above zero.

Two further notes on the harness:

- Every test runs under `runTest`'s single-threaded scheduler, so the token-loss race (defect 2) cannot occur there by construction. A regression test for the *hang* needs `Dispatchers.Default` and repetition; a `runTest` test can still catch the deterministic `observerCount` leak.
- `dedupe_canceled_request_resumes_waiter` uses `Channel<Int>(capacity = 2)` while the sibling failure test uses `Channel.UNLIMITED`. A corrected test producing a third attempt would suspend inside `block()` on `started.send(...)`, adding an unintended suspension point in the middle of the coordinated region.

The `runTest` snippet under [Repro 1](#repro-1--cancellation-storm-the-reported-case) is the missing test in its smallest form: it cancels a parked waiter, then asserts that a later request for the same key still runs. It fails today.

---

## Suggested fix

The minimal patch is a scoping change plus a widened `catch`:

```diff
     var shouldWait = true
     val request = synchronized(lock) {
         concurrentRequests.getOrPut(key) {
             shouldWait = false
             Request()
         }
     }.acquire()
 
-    if (shouldWait) {
-        request.channel.receiveCatching()
-    }
-
     try {
+        if (shouldWait) {
+            request.channel.receiveCatching()
+        }
         return block().also {
             request.markSucceeded()
         }
-    } catch (e: Exception) {
+    } catch (e: Throwable) {
         request.channel.trySend(Unit)
         throw e
     } finally {
         request.release {
             synchronized(lock) {
                 concurrentRequests -= key
             }
         }
     }
```

That closes the leak and the `Error` hang. It leaves defects 2–5 in place, since the single-token baton is still losable and `acquire()` is still outside the lock.

<details>
<summary><b>Alternative: make the baton and the observer count unnecessary</b> (click to expand)</summary>

The bookkeeping exists to decide *when* the key may be released. If ownership is represented directly, waiters hold no shared state at all, so cancelling one cannot corrupt anything — there is nothing to leak, no token to lose, and only one lock:

```kotlin
class DeDupeConcurrentRequestStrategy : ConcurrentRequestStrategy {

    private val lock = SynchronizedObject()

    /** Key -> leader completion, carrying `true` when the leader's `block` succeeded. */
    private val inFlight = mutableMapOf<String, CompletableDeferred<Boolean>>()

    override suspend fun apply(
        key: String,
        block: suspend () -> FetchResult,
    ): FetchResult {
        while (true) {
            var isLeader = false
            val request = synchronized(lock) {
                inFlight.getOrPut(key) {
                    isLeader = true
                    CompletableDeferred()
                }
            }

            if (!isLeader) {
                // Cancellation here leaves no shared state behind, so the key stays usable.
                if (request.await()) return block()
                continue
            }

            var succeeded = false
            try {
                return block().also { succeeded = true }
            } finally {
                // Release ownership before waking waiters so they never observe a stale entry.
                synchronized(lock) { inFlight -= key }
                request.complete(succeeded)
            }
        }
    }
}
```

The observable contract is unchanged: one leader per key; on success every waiter is released and re-runs `block()` for the disk-cache hit; on failure or cancellation one waiter takes over and retries while the rest keep waiting. Because the leader completes the deferred in a `finally`, no `catch` is needed and `Error`s behave correctly for free.

Against Repro 1 this reports `permanently stuck keys: 0 / 400`, and against Repro 3 `waiter = SourceFetchResult(…)`.

</details>

### Unrelated but adjacent

`NetworkFetcher` wraps the whole of `doFetch()` in the strategy, including the `readFromDiskCache()` fast path. A caller whose image is already fully on disk still parks behind an unrelated in-flight *network* request for the same key, and eats the full round-trip. Only the network portion needs coordinating. Happy to open that separately if it is worth its own issue.

---

## Version

| | |
| :--- | :--- |
| **Coil version** | `3.6.0` |
| **Also affected** | `3.4.0`, `3.5.0`, current `main` — `ConcurrentRequestStrategy.kt` is byte-identical across all four |
| **Module** | `coil-network-core` (`commonMain`) |
| **Platform** | Not platform-specific. The file is common code with no `expect`/`actual`; repros above were run on the JVM. |
| **API level** | Not API-level or device specific — no Android framework code is involved. |
| **Reproducibility** | Repro 1: ~3.5% of iterations, permanent once hit; its `runTest` variant fails 100% of the time. Repro 3: 100%, deterministic, single-threaded. |

> [!NOTE]
> Logging shows nothing useful here, which is part of what makes this hard to diagnose in the wild: the request never fails, it never completes, and `onError` is never invoked. Setting `concurrentRequestStrategy` back to the default `ConcurrentRequestStrategy.UNCOORDINATED` is an effective workaround.

[^timing]: 14/400 on one machine; the window depends on how often `trySend` targets a continuation that is mid-cancellation, so core count, dispatcher load and the cancel jitter all move the number. It reproduces on every run.

colinrtwhite:
Fixes: https://github.com/coil-kt/coil/issues/3562

### DIFF
--- a/coil-network-core/src/commonMain/kotlin/coil3/network/ConcurrentRequestStrategy.kt
+++ b/coil-network-core/src/commonMain/kotlin/coil3/network/ConcurrentRequestStrategy.kt
@@ -5,7 +5,7 @@ import coil3.fetch.FetchResult
 import kotlin.jvm.JvmField
 import kotlinx.atomicfu.locks.SynchronizedObject
 import kotlinx.atomicfu.locks.synchronized
-import kotlinx.coroutines.channels.Channel
+import kotlinx.coroutines.CompletableDeferred
 
 /**
  * Coordinates concurrent requests for the same key.
@@ -39,63 +39,40 @@ private class UncoordinatedConcurrentRequestStrategy : ConcurrentRequestStrategy
  */
 @ExperimentalCoilApi
 class DeDupeConcurrentRequestStrategy : ConcurrentRequestStrategy {
-    private val concurrentRequests = mutableMapOf<String, Request>()
+    private val concurrentRequests = mutableMapOf<String, CompletableDeferred<Boolean>>()
     private val lock = SynchronizedObject()
 
     override suspend fun apply(
         key: String,
         block: suspend () -> FetchResult,
     ): FetchResult {
-        var shouldWait = true
-        val request = synchronized(lock) {
-            concurrentRequests.getOrPut(key) {
-                shouldWait = false
-                Request()
+        while (true) {
+            var isLeader = false
+            val request = synchronized(lock) {
+                concurrentRequests.getOrPut(key) {
+                    isLeader = true
+                    CompletableDeferred()
+                }
             }
-        }.acquire()
 
-        if (shouldWait) {
-            request.channel.receiveCatching()
-        }
-
-        try {
-            return block().also {
-                request.markSucceeded()
+            if (!isLeader) {
+                val succeeded = request.await()
+                if (succeeded) {
+                    return block()
+                }
+                continue
             }
-        } catch (e: Exception) {
-            request.channel.trySend(Unit)
-            throw e
-        } finally {
-            request.release {
+
+            var succeeded = false
+            try {
+                return block().also {
+                    succeeded = true
+                }
+            } finally {
                 synchronized(lock) {
                     concurrentRequests -= key
                 }
-            }
-        }
-    }
-
-    private class Request {
-        val channel = Channel<Unit>(Channel.UNLIMITED)
-
-        private val lock = SynchronizedObject()
-        private var hasSucceeded = false
-        private var isClosed = false
-        private var observerCount = 0
-
-        fun acquire(): Request = synchronized(lock) {
-            observerCount++
-            this
-        }
-
-        fun markSucceeded() = synchronized(lock) {
-            hasSucceeded = true
-        }
-
-        fun release(cleanup: () -> Unit) = synchronized(lock) {
-            if ((--observerCount <= 0 || hasSucceeded) && !isClosed) {
-                channel.close()
-                cleanup()
-                isClosed = true
+                request.complete(succeeded)
             }
         }
     }
--- a/coil-network-core/src/commonTest/kotlin/coil3/network/ConcurrentRequestStrategyTest.kt
+++ b/coil-network-core/src/commonTest/kotlin/coil3/network/ConcurrentRequestStrategyTest.kt
@@ -9,10 +9,12 @@ import kotlin.test.assertEquals
 import kotlin.test.assertFalse
 import kotlin.test.assertIs
 import kotlin.test.assertTrue
-import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.CompletableDeferred
 import kotlinx.coroutines.async
+import kotlinx.coroutines.awaitCancellation
+import kotlinx.coroutines.cancelAndJoin
 import kotlinx.coroutines.channels.Channel
+import kotlinx.coroutines.launch
 import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
 import kotlinx.coroutines.test.runTest
@@ -133,18 +135,17 @@ class ConcurrentRequestStrategyTest {
     fun dedupe_canceled_request_resumes_waiter() = runTest {
         val strategy = DeDupeConcurrentRequestStrategy()
         val started = Channel<Int>(capacity = 2)
-        val cancelLeader = CompletableDeferred<Unit>()
+        val leaderStarted = CompletableDeferred<Unit>()
         val waiterStarted = CompletableDeferred<Unit>()
 
-        val leader = async {
-            runCatching {
-                strategy.apply("key") {
-                    started.send(1)
-                    cancelLeader.await()
-                    throw CancellationException("leader canceled")
-                }
+        val leader = launch {
+            strategy.apply("key") {
+                leaderStarted.complete(Unit)
+                started.send(1)
+                awaitCancellation()
             }
         }
+        leaderStarted.await()
         assertEquals(1, started.receive())
 
         val waiter = async {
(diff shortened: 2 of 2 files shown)


coil-network-core/src/commonMain/kotlin/coil3/network/ConcurrentRequestStrategy.kt
```diff
@@ -39,63 +39,40 @@ private class UncoordinatedConcurrentRequestStrategy : ConcurrentRequestStrategy
...
+            val request = synchronized(lock) {
+                concurrentRequests.getOrPut(key) {
+                    isLeader = true
+                    CompletableDeferred()
+                }
             }
-        }.acquire()
-
-        if (shouldWait) {
-            request.channel.receiveCatching()
```
colinrtwhite, line 58: This is the main issue. We would throw a `CancellationException` here and then `request.release {}` would never be called and the request's `observerCount` wouldn't be decremented.

coil-network-core/src/commonMain/kotlin/coil3/network/ConcurrentRequestStrategy.kt
```diff
@@ -39,63 +39,40 @@ private class UncoordinatedConcurrentRequestStrategy : ConcurrentRequestStrategy
...
-        }
 
-        try {
-            return block().also {
-                request.markSucceeded()
+            if (!isLeader) {
+                val succeeded = request.await()
+                if (succeeded) {
```
colinrtwhite, line 60: Instead of manually managing the observer count, now we let all observers watch the same `CompletableDeferred`. If the leader fails, then all observers will race to be the new leader (guarded by the lock).

chris-horner: Neat!

coil-network-core/src/commonMain/kotlin/coil3/network/ConcurrentRequestStrategy.kt
```diff
@@ -39,63 +39,40 @@ private class UncoordinatedConcurrentRequestStrategy : ConcurrentRequestStrategy
...
+                    return block()
+                }
+                continue
```
chris-horner, line 63: It took me a while to grok how this would work to unwind the chain of leaders but it's pretty slick.

chris-horner (approved):

PR #3568, opened 2026-09-02
Use InternalSkikoApi awaitSkiko directly in web worker decoding

terrakok:
To avoid wasm linkage error with CMP 1.13

https://github.com/coil-kt/coil/pull/3534#issuecomment-5511825937


colinrtwhite (approved):
Thanks for catching and fixing this!

PR #3575, opened 2026-09-07
Web image decoding performance

terrakok:
Improve Web image decoding performance and cancellation handling:

 - Mark decoded bitmaps as immutable so Skia can share their pixel memory instead of copying it during rendering.
 - Notify the web worker when an image request is cancelled and skip unnecessary canvas rendering, pixel extraction, and result delivery.
 - Use shared worker event listeners with a pending-request map instead of registering separate listeners for every decode request.


colinrtwhite (approved):
Awesome, thank you!

PR #3577, opened 2026-09-08
Fix #2094 - preserve gainmaps in hardware decodes

### LINKED ISSUE #2094: Consider adding workaround for U + gainmap + hardware bitmap bug
Prior to Android V there exists a bug where when:

1) A decode to Config.HARDWARE is requested
2) The image format is A8 / gray8
3) The device supports A8 hardware buffers
4) The device is using OpenGL

Then the decode fails. As A8 images are extremely rare, this hadn't come up before. However, with the addition of gainmap images the usage of A8 "images" became much more common as a common space/quality tradeoff for the gainmap is to use a8 and provide only luminance modulation as a result. In this scenario it becomes quite possible for all 4 of the above conditions to be hit.

However, it is rather easy to both detect the presence of the bug (simply try to copy an A8 bitmap to Config.HARDWARE - if null is returned the bug is present) and to workaround it in a decently efficient means by simply decoding to non-hardware first, converting the A8 gainmap to 8888, and then copying the result to HARDWARE and returning that. This is not dissimilar to how decoding directly to HARDWARE works, so performance impact should be minimal. See the Glide implementation of both here: https://github.com/bumptech/glide/pull/5357

If nothing is done about this then the failure mode is simply that the decoding of the gainmap portion fails, but the primary image is still successfully returned. As in, the behavior simply matches what Android T did, prior to the introduction of gainmaps. So the failure mode is quite forgiving.

AvikMakwana:
﻿## Summary

Fixes #2094.

Android 14 can silently drop an Ultra HDR gainmap when decoding a JPEG directly to `Bitmap.Config.HARDWARE` if the device cannot allocate an `ALPHA_8` hardware bitmap.

This change detects that capability at runtime and, only on affected Android 14 devices, decodes JPEGs through software first. If the decoded image contains an `ALPHA_8` gainmap, its values are converted to an opaque `ARGB_8888` bitmap before the base image and gainmap are copied to hardware.

The workaround is limited to API 34 JPEG hardware decodes on devices where `ALPHA_8 -> HARDWARE` is unsupported. Other formats, API levels, and software bitmap requests continue through their existing decode paths.

## Testing

Reproduced the issue on an Android 14 / API 34 emulator using the AOSP `fountain_night.jpg` Ultra HDR image with the `skiagl` renderer.

Added device tests covering:
- `BitmapFactoryDecoder` hardware gainmap preservation
- `StaticImageDecoder` hardware gainmap preservation
- software decode behavior
- non-gainmap JPEG behavior
- non-JPEG hardware decoding
- workaround capability gating
- gainmap pixel conversion and metadata preservation

Validation:
- Android 14 gainmap device tests: 8/8 passed
- `spotlessCheck`: passed
- `checkKotlinAbi`: passed
- `:coil-core:lint`: passed

`./test.sh` was also run. Its ABI and formatting checks passed, as did the relevant Coil core Android/JVM tests. The full script could not complete on the Windows host because LinuxX64 native linking requires Linux Skiko artifacts and the multiplatform test suite encountered host-specific Roborazzi snapshot formatting.

review-path-agent


coil-core/src/androidMain/kotlin/coil3/decode/GainmapUtils.kt
```diff
@@ -0,0 +1,138 @@
...
+        }
+        if (convertedGainmapContents != null && hardwareGainmapContents !== convertedGainmapContents) {
+            convertedGainmapContents.recycle()
+        }
+
+        return hardwareBaseBitmap
+    }
+
+    @RequiresApi(34)
+    internal fun convertAlpha8ToArgb8888(source: Bitmap): Bitmap {
```
colinrtwhite, line 96: `internal` methods on an `internal` object have no effect.

coil-core/src/androidMain/kotlin/coil3/decode/GainmapUtils.kt
```diff
@@ -0,0 +1,138 @@
...
+            false
+        }
+    }
+
+    /**
+     * Returns true if the decoding path must work around the Android 14 platform bug
+     * where single-channel (ALPHA_8) gainmaps are dropped when decoding directly to HARDWARE.
+     */
+    fun shouldWorkAroundHardwareGainmap(mimeType: String?): Boolean {
+        return SDK_INT == 34 && mimeType == MIME_TYPE_JPEG && !isHardwareGainmapSupported
```
colinrtwhite, line 32: Why is this only applied to `MIME_TYPE_JPEG`? We can't depend on mime type being valid.

coil-core/src/androidMain/kotlin/coil3/decode/GainmapUtils.kt
```diff
@@ -0,0 +1,138 @@
...
+        }
+
+        return hardwareBaseBitmap
+    }
+
+    @RequiresApi(34)
+    internal fun convertAlpha8ToArgb8888(source: Bitmap): Bitmap {
+        val width = source.width
+        val height = source.height
+        val alphaBuffer = ByteBuffer.allocate(width * height)
```
colinrtwhite, line 99: Don't we want to do [this](https://github.com/coil-kt/coil/issues/2094#issuecomment-1993157144) instead? This current solution will allocate 5 bytes per pixel in the gainmap (`ByteBuffer` + `IntArray` below), which is a lot.

coil-core/src/androidMain/kotlin/coil3/decode/GainmapUtils.kt
```diff
@@ -0,0 +1,138 @@
...
+        }
+
+        val argbBitmap = Bitmap.createBitmap(width, height, Bitmap.Config.ARGB_8888)
+        argbBitmap.setPixels(argbArray, 0, width, 0, 0, width, height)
+        return argbBitmap
+    }
+
+    @RequiresApi(34)
+    internal fun copyGainmap(source: Gainmap, newContents: Bitmap): Gainmap {
+        val gainmap = Gainmap(newContents)
```
colinrtwhite, line 116: Is there any way we can avoid manually copying these properties? Currently if new properties are added we'll have to notice and manually update this method.

colinrtwhite (commented):
With respect, did you review this PR prior to submission and do you understand the underlying changes? I notice the `review-path-agent` token has not been removed from this PR description, which is only added when an LLM creates the pull request.

Last time I took a quick look at doing something similar to solve that issue it had a performance cost due to having to use software bitmaps for this edge case instead of hardware bitmaps. This is a sensitive area of the codebase performance-wise and changes need to be made carefully.

AvikMakwana:
Thanks for the review. I revisited the implementation and found an important tradeoff before pushing another iteration.

Removing the MIME gate would force all hardware decodes through a software intermediate on affected Android 14 devices, including normal non-gainmap images. Android’s decode APIs don’t expose gainmap presence before choosing the allocator, so narrowing the workaround would require either encoded-format detection or additional metadata parsing

Before I update the implementation, would you prefer the workaround to apply to all hardware decodes on affected API 34 devices, or keep it limited to detected JPEG/container formats?

colinrtwhite:
> Removing the MIME gate would force all hardware decodes through a software intermediate on affected Android 14 devices, including normal non-gainmap images. Android’s decode APIs don’t expose gainmap presence before choosing the allocator, so narrowing the workaround would require either encoded-format detection or additional metadata parsing

Yes, that's the performance issue I mentioned. We can't restrict this change to just JPG files as many other file formats support gainmaps. I'm still not sure how to properly address this bug without the performance hit on API 34, which is why it isn't fixed yet.

Sorry, given your response and the fact that you ignored my other comments makes me think you're offloading this understanding work to an LLM. This is a sensitive change to a performance-critical part of the codebase and changes to it need to be made to it with care. LLM-assisted changes are OK in this codebase, however ultimately you need to understand and validate your own changes separately from the LLM.

PR #3583, opened 2026-09-15
Replace deprecated Gradle project dependency notation.

colinrtwhite:
Passing a `Project `directly to `api()` was deprecated in a recent Gradle version. This PR removes usages of that API.


coil-bom/build.gradle.kts
```diff
@@ -15,7 +15,7 @@ dependencies {
     constraints {
         for (subproject in rootProject.subprojects) {
             if (subproject.name in publicModules) {
-                api(subproject)
+                api(project(subproject.path))
```
colinrtwhite, line 18: `project` returns a `ProjectDependency`.

chris-horner (approved):

PR #3599, opened 2026-09-21
Make sample app image grids adapt to available width instead of using a fixed number of columns

colinrtwhite:
Update the samples to use a minimum column width instead of a fixed number of columns. Also includes a few insets-related changes to make sure it works properly on the iPhone Duo.

coil-duo-closed-finalcoil-duo-open-before


samples/view/src/main/java/sample/view/ImageGridView.kt
chris-horner: Image

chris-horner (approved):

